PR #1, opened 2019-12-07
Whitelist added to interface.json, accessed by QML and C++

kasenvr:
The Entity Script Whitelist now combines the environment variables with the interface.json retrieved settings, there is now a QML interface under Settings (right below the Security... tab) to view and edit that interface.json whitelist.

Changes made to the whitelist through the QML interface will be reflected and be operational immediately. This allows for live handling of the whitelist without restarting interface. This opens up the door for creating temporary whitelists and accepting new domains for whitelists via prompt so users do not have to manually manage it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 ### [Download (Windows 64-bit, .zip)](https://realities.dev/cdn/hifi-community/v0860-kasen-VS-release+freshstart/Packaged_Release.zip)
 
-#### Changes for **v0.86.0 K1** consist of:
+#### Changes for **v0.86.0** consist of:
+
+#### Added in K1 (12/3/19)
 
 * Audio Buffer choppy audio bugfix by increasing the buffer size.
 * User Activity Logger disabled, option in code to log the reports to console.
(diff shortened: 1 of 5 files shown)


kasenvr:
Also, should the High Fidelity CDN URLs be whitelisted by default, hardcoded into the .cpp?

MarcusLlewellyn:
Suggestion: Maybe a "Strict Mode" checkbox. When checked even known trusted sources must be explicitly whitelisted. Default would allow sources like High Fidelity and potentially services like Tivoli and Kitely. Don't hardcore, rather have a trusted source json object with url properties, and then User object.

kasenvr:
That's a good suggestion. It has also been suggested to forgo that and skip straight ahead to temp lists or whitelisting entire domains to enable experiences to work. It might be better suited to assign programming time in C++ and QML towards that end, skipping ahead and such and getting us those features faster.

So, I'm thinking in the spirit of rapid release, by today or tomorrow's end we get this as K2 out the door with the advanced and easy to use whitelist in people's hands as a good starting point. Then any contributors can focus on handing off new features for K3, K4, and so on.

I sense with the slowness of HiFi to get out 87 and subsequent releases we should be able to get some solid features out in between and quickly. That's a major draw I imagine, right? To have that agility.

PR #3, opened 2019-12-08
Upgrade kcpkg port files from SDL2 from 2.0.8 to 2.0.10. This fixes patch issues.

MarcusLlewellyn:


kasenvr:
Built the CMake successfully. This issue is caused due to an old VCPKG version being retrieved. You will need the "HIFI_VCPKG_BOOTSTRAP" environment variable set to "1" so it retrieves the 2019-9-12 version of VCPKG instead of the 2018 version.

PR #8, opened 2019-12-10
Removes the environment variable requirement for procedural material shaders.

MarcusLlewellyn:
With this change, it is no longer necessary to create the HIFI_ENABLE_MATERIAL_PROCEDURAL_SHADERS environment variable. Users only need to have the Developer menu active, and then check the menu option Developer->Render->Enable Procedural Materials

Tested on:
Windows 10 Pro 1909 build 18363.476
CMake 3.16.0
Microsoft Visual Studio Community 2019 Version 16.4.0


kasenvr:
Closed but not merged? What's the matter?

PR #11, opened 2019-12-11
Replace the interface icons with Kasen logo icon.

MarcusLlewellyn:
We probably don't want to keep using the High Fidelity "voxel" icon for the interface. This PR replaces it with the Kasen logo. Works on Windows, but can't be sure about the Mac. If we feel we want a different icon altogether, I suggest using this in the rinterim until a new one is chosen.
Kasen-logo
Tested on:
Windows 10 Pro 1909 build 18363.476
CMake 3.16.0
Microsoft Visual Studio Community 2019 Version 16.4.0


kasenvr:
O-oh... I was literally just doing this. Just pushed the branch. LOL thanks! :) I wanted to use this icon, yes. It seems perfect for an icon as well.

Also I was thinking themed edition names would be nice. I like the idea that they named the "Firestorm" client as such and that Linux editions have flavors and/or nicknames. So when High Fidelity does a protocol upgrade e.g. v0.87.0 we do something like v0.87.0 "Athena" K1. This way we can have that brand recognition and mutual understanding amongst peers "Oh, you're on Athena?"

PR #12, opened 2019-12-11
Adds community script directory, adds chat as default app.

MarcusLlewellyn:
This adds a community directory under the scripts directory. George Deacon's chat app has been placed within, and is loaded as a default script.


kasenvr:
I suggest we get in touch with George to get a version system going for his app so we can better keep track of upgrading the app. :)

PR #16, opened 2019-12-14
QML Whitelisting

FluffyJenkins:
Allows external whitelisting of QML files.

Requires the whole url (up to the .qml) in the entity script whitelist

e.g. 

image

sample urls:

https://realities.dev/cdn/qml/FloofChat.qml
http://home.darlingvr.club/hifi-community/FloofChat.qml

Test code:

```
var chatBar = new OverlayWindow({
    source: "https://realities.dev/cdn/qml/FloofChat.qml",
    width: 360,
    height: 180
});
Script.setTimeout(function(){
    chatBar.close();
},2000)
```

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3182,13 +3182,37 @@ static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://content.highfidelity
 
 void Application::initializeUi() {
 
+    
     // Allow remote QML content from trusted sources ONLY
     {
         auto defaultUrlValidator = OffscreenQmlSurface::getUrlValidator();
-        auto newValidator = [=](const QUrl& url)->bool {
-            if (AUTHORIZED_EXTERNAL_QML_SOURCE.isParentOf(url)) {
-                return true;
+        auto newValidator = [=](const QUrl& url) -> bool {
+            QString whitelistPrefix = "[WHITELIST ENTITY SCRIPTS]";
+            QList<QString> safeURLS = { "" };
+            safeURLS += qEnvironmentVariable("EXTRA_WHITELIST").trimmed().split(QRegExp("\\s*,\\s*"), QString::SkipEmptyParts);
+
+            // PULL SAFEURLS FROM INTERFACE.JSON Settings
+
+            QVariant raw = Setting::Handle<QVariant>("private/settingsSafeURLS").get();
+            QStringList settingsSafeURLS = raw.toString().trimmed().split(QRegExp("\\s*[,\r\n]+\\s*"), QString::SkipEmptyParts);
+            safeURLS += settingsSafeURLS;
+
+            // END PULL SAFEURLS FROM INTERFACE.JSON Settings
+
+            bool isInWhitelist = false;  // assume unsafe
+            for (const auto& str : safeURLS) {
+                qDebug() << "url.toString().startsWith(str) = " << url.toString().startsWith(str);
+                qDebug() << "str.endsWith('.qml ') = " << str.endsWith(".qml");
+                if (!str.isEmpty() && str.endsWith(".qml") && url.toString().endsWith(".qml") &&
+                    url.toString().startsWith(str)) {
+                    qDebug() << "found matching url!" << url.host();
+                    isInWhitelist = true;
+                    return true;
+                    break;  // bail early since we found a match
+                }
             }
+
+            qDebug() << "no matching url :c" << url.host();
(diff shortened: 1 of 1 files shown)


kasenvr:
Darling has tested and confirmed this is working as intended. Has anyone else had a chance to try it?

interface/src/Application.cpp
```diff
@@ -3182,13 +3182,37 @@ static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://content.highfidelity
...
+
+            QVariant raw = Setting::Handle<QVariant>("private/settingsSafeURLS").get();
+            QStringList settingsSafeURLS = raw.toString().trimmed().split(QRegExp("\\s*[,\r\n]+\\s*"), QString::SkipEmptyParts);
+            safeURLS += settingsSafeURLS;
+
+            // END PULL SAFEURLS FROM INTERFACE.JSON Settings
+
+            bool isInWhitelist = false;  // assume unsafe
+            for (const auto& str : safeURLS) {
+                qDebug() << "url.toString().startsWith(str) = " << url.toString().startsWith(str);
```
thoys: are these debug lines needed?

interface/src/Application.cpp
```diff
@@ -3182,13 +3182,37 @@ static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://content.highfidelity
...
+                qDebug() << "str.endsWith('.qml ') = " << str.endsWith(".qml");
```
thoys: if you keep the debug lines here it would be preferred to use qCDebug instead and assign the right category, otherwise it will be hard to find or filter in the log.

interface/src/Application.cpp
```diff
@@ -3182,13 +3182,37 @@ static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://content.highfidelity
 
 void Application::initializeUi() {
 
+
```
thoys: new line can be omitted

interface/src/Application.cpp
kasenvr: Like so:
`qCDebug(scriptengine) << `
This is how I did it for the Entity Script Whitelist but I removed the debug statements, though I must say it might still be necessary to at least post if it's been approved since it's easier to do that code-wise, as if you don't see an approval then you should assume it failed (When debugging as an end-user.)

FluffyJenkins: dude I have no idea what qCDebug is

thoys: A variant on qDebug which also takes the category. See https://doc.qt.io/qt-5/qloggingcategory.html#qCDebug-1     .  There are many categories already defined see if you find one that is appropriate

MarcusLlewellyn:
Building on Windows and Linux now.

MarcusLlewellyn:
Built, run, and feature tested on the following platforms.

CMake 3.16.0


gcc/g++ 7.4.0
CMake 3.10.2

interface/src/Application.cpp
```diff
@@ -3185,10 +3185,30 @@ void Application::initializeUi() {
...
+            QStringList settingsSafeURLS = raw.toString().trimmed().split(QRegExp("\\s*[,\r\n]+\\s*"), QString::SkipEmptyParts);
+            safeURLS += settingsSafeURLS;
+
+            // END PULL SAFEURLS FROM INTERFACE.JSON Settings
+
+            bool isInWhitelist = false;  // assume unsafe
+            for (const auto& str : safeURLS) {
+                if (!str.isEmpty() && str.endsWith(".qml") && url.toString().endsWith(".qml") &&
+                    url.toString().startsWith(str)) {
+                    qCDebug(interfaceapp) << "Found matching url!" << url.host();
```
kasenvr: Is it possible to put this in the "qml" section of the debug log and put as a prefix for the message something such as: [QML Whitelist] so it immediately stands out? :)

interface/src/Application.cpp
```diff
@@ -3185,10 +3185,30 @@ void Application::initializeUi() {
...
+                    isInWhitelist = true;
+                    return true;
+                }
             }
+
+            qCDebug(interfaceapp) << "No matching url" << url.host();
```
kasenvr: https://doc.qt.io/qt-5/qurl.html#host If the URL does not have a host or valid host, it will not show, both statements should show more information (the full URL including the file itself) to the user so they know exactly what QML file was trying to load and if it succeeded or failed, what file it was.

PR #19, opened 2019-12-16
Change asset URLs to Athena S3 bucket.

daleglass:
This uses a variable that can be overridden with an environment variable.
This makes it easy for the community to create alternate asset hosts.


CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ else()
   cmake_minimum_required(VERSION 3.2)
 endif()
 
+set(EXTERNAL_BUILD_ASSETS "$ENV{EXTERNAL_BUILD_ASSETS}")
+if (NOT "${EXTERNAL_BUILD_ASSETS}")
+  set(EXTERNAL_BUILD_ASSETS "'https://athena-public.s3.amazonaws.com")
```
MarcusLlewellyn: Looks like there is a typo here in the URL?

MarcusLlewellyn:
So, this doesn't work. This is because vcpkg creates a shell with a pristine build environment for each package, wiping out any environment variables.

According to the vcpkg GitHub, the recommended way to do this is to read our value from a file, and then to set the environment variable within a triplet. Our code uses a number of portfiles[1] affected by this and it looks like some changes need to be added to at least a few of them. Here's a template:
```cmake
file(READ "${CMAKE_CURRENT_LIST_DIR}/../FOO_SDK_DIR.txt" _foo_sdk_dir)
set(ENV{FOO_SDK_DIR} "${_foo_sdk_dir}")
```
This reads the value from a file, and then sets the environment variable within the shell vcpkg creates. I don't know CMake well enough to decipher exactly where the first line is looking for the file, but hopefully this is a clue.

[1] portfile.cmake for:
hifi-clients-deps
hifi-deps
hifi-host-tools
hifi-scribe

daleglass:
> So, this doesn't work. This is because vcpkg creates a shell with a pristine build environment for each package, wiping out any environment variables.

The heck? It seemed to work when I tried this yesterday. Maybe some vcpkg caching issue...

Okay, I'm looking at all of that.

daleglass:
I see, useful link on this: https://github.com/Microsoft/vcpkg/issues/3712

Fixing.

CMakeLists.txt
thoys:
```suggestion
  set(EXTERNAL_BUILD_ASSETS "https://athena-public.s3.amazonaws.com")
```

daleglass: Fixed

daleglass:
Okay, second attempt. Everything seems to work fine on Linux.

kasenvr:
CMake completed successfully with 

CMake 3.15.5
OS Name	Microsoft Windows 10 Pro
Version	10.0.18362 Build 18362
Visual Studio 2019

MarcusLlewellyn:
This has been successfully built and run on the following platforms.

CMake 3.16.0


gcc/g++ 7.4.0
CMake 3.10.2

PR #21, opened 2019-12-17
Fix the fix for SDL2

MarcusLlewellyn:
vcpkg doesn't like unexpected blank lines. Live and learn.


kasenvr:
Fix the fix for the fix of the fix.

PR #23, opened 2019-12-18
Update Primary CMakeLists project name to Athena

MarcusLlewellyn:
Having a "hifi.sln" file feels wrong. Let's fix that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,7 @@ if(NOT EXISTS "${CMAKE_BINARY_DIR}/vcpkg.cmake")
 endif()
 include("${CMAKE_BINARY_DIR}/vcpkg.cmake")
 
-project(hifi)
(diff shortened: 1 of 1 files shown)


kasenvr:
Is that the only place it shows up in, I wonder?

MarcusLlewellyn:
There are like 11000 uses of the string "hifi" all over the codebase. But this one is one of the most front-facing, and we gotta start somewhere. ;)

kasenvr:
Oh I 100% agree, I just meant if there are any other references towards that that could break it.

MarcusLlewellyn:
It really doesn't affect anything other than the name of the project.

thoys:
the short name could be "athena"  instead of "hifi"

thoys:
ah you already did it, great. did someone test this?

PR #29, opened 2019-12-19
Redo graphics presets

daleglass:
* Add "battery saver" mode, which is the only one that has "economical"
  rendering. This mode will never be selected automatically and must be
  chosen by the user.
* Change all other modes to realtime.

Reasoning:

Whenever interface crashes, it sets the graphics settings to low. This not
only has low detail, but also limited framerate. This may create the impression
that there's something seriously wrong with the rendering engine.

Now the interface will always have the best framerate, except when the user
wishes otherwise.

### DIFF
--- a/interface/src/PerformanceManager.cpp
+++ b/interface/src/PerformanceManager.cpp
@@ -117,7 +117,7 @@ void PerformanceManager::applyPerformancePreset(PerformanceManager::PerformanceP
             DependencyManager::get<LODManager>()->setWorldDetailQuality(WORLD_DETAIL_LOW);
 
         break;
-        case PerformancePreset::ECO:
+        case PerformancePreset::BATT:
             RenderScriptingInterface::getInstance()->setRenderMethod(RenderScriptingInterface::RenderMethod::FORWARD);
             RenderScriptingInterface::getInstance()->setShadowsEnabled(false);
             qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::ECO);
--- a/interface/src/PerformanceManager.cpp
+++ b/interface/src/PerformanceManager.cpp
@@ -103,11 +103,21 @@ void PerformanceManager::applyPerformancePreset(PerformanceManager::PerformanceP
             RenderScriptingInterface::getInstance()->setViewportResolutionScale(recommandedPpiScale);
 
             RenderScriptingInterface::getInstance()->setShadowsEnabled(false);
-            qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::INTERACTIVE);
+            qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::REALTIME);
             DependencyManager::get<LODManager>()->setWorldDetailQuality(WORLD_DETAIL_MEDIUM);
 
         break;
         case PerformancePreset::LOW:
+            RenderScriptingInterface::getInstance()->setRenderMethod(RenderScriptingInterface::RenderMethod::FORWARD);
+            RenderScriptingInterface::getInstance()->setShadowsEnabled(false);
+            qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::REALTIME);
+
+            RenderScriptingInterface::getInstance()->setViewportResolutionScale(recommandedPpiScale);
+
+            DependencyManager::get<LODManager>()->setWorldDetailQuality(WORLD_DETAIL_LOW);
+
+        break;
+        case PerformancePreset::ECO:
             RenderScriptingInterface::getInstance()->setRenderMethod(RenderScriptingInterface::RenderMethod::FORWARD);
             RenderScriptingInterface::getInstance()->setShadowsEnabled(false);
             qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::ECO);
--- a/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
+++ b/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
@@ -53,6 +53,20 @@ Item {
                 spacing: 0
 
                 HifiControlsUit.RadioButton {
+                    id: performanceEco
+                    colorScheme: hifi.colorSchemes.dark
+                    height: 18
+                    fontSize: 16
+                    leftPadding: 0
+                    text: "Battery saver"
+                    checked: Performance.getPerformancePreset() === PerformanceEnums.ECO
+                    onClicked: {
+                        Performance.setPerformancePreset(PerformanceEnums.ECO);
+                        root.refreshAllDropdowns();
+                    }
+                }
+
+                HifiControlsUit.RadioButton {
                     id: performanceLow
                     colorScheme: hifi.colorSchemes.dark
                     height: 18
--- a/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
+++ b/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
@@ -59,9 +59,9 @@ Item {
                     fontSize: 16
                     leftPadding: 0
                     text: "Battery saver"
-                    checked: Performance.getPerformancePreset() === PerformanceEnums.ECO
+                    checked: Performance.getPerformancePreset() === PerformanceEnums.BATT
                     onClicked: {
-                        Performance.setPerformancePreset(PerformanceEnums.ECO);
+                        Performance.setPerformancePreset(PerformanceEnums.BATT);
                         root.refreshAllDropdowns();
                     }
                 }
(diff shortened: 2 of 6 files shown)


kasenvr:
Fixes https://github.com/kasenvr/project-athena/issues/18 ?

interface/src/PerformanceManager.cpp
```diff
@@ -103,11 +103,21 @@ void PerformanceManager::applyPerformancePreset(PerformanceManager::PerformanceP
             RenderScriptingInterface::getInstance()->setViewportResolutionScale(recommandedPpiScale);
 
             RenderScriptingInterface::getInstance()->setShadowsEnabled(false);
-            qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::INTERACTIVE);
+            qApp->getRefreshRateManager().setRefreshRateProfile(RefreshRateManager::RefreshRateProfile::REALTIME);
```
thoys: you changed INTERACTIVE to REALTIME here,  what would the side-effect be?

daleglass: INTERACTIVE caps the framerate at 30 FPS.

I think capping the framerate is an odd thing to do. Normally one chooses "low detail" settings because one wants things to work faster. Capping framerate seems to me something like that should only be done when the user explicitly wants it.

thoys: gotcha, but the lower fps might be a reason why it runs better on low-end machines now.

thoys: can someone with a low-end machine test the difference?

ctrlaltdavid:
This should be tagged with a "Scripting API Change" label.
The JSDoc that's in HiFi's master branch as of a few days ago will need to be updated when those changes are included in Project Athena.

interface/src/PerformanceManager.cpp
MarcusLlewellyn: Will be able to test in a bit on an intel GPU after I compile.

daleglass: Hmm, interesting. How does it improve usability?

In pretty much all games I'm familiar with the reason to drop quality is to improve framerate.

thoys: I think that hifi is setup to dynamically change the LOD if the framerate is not at the preferred level. So if this drops to a level that works on low-end-machines the LOD doesn't endlessly fall. But I could be mistaken @HifiExperiments ?

HifiExperiments: yeah exactly.  basically, when a low end machine can’t hit the target framerate, all sorts of things go wrong.  you start missing frames and then work piles up that we couldn’t do in our frame budget and the LOD system will freak out and you won’t be able to see anything.  I’d recommend against this change for now, but I agree there’s things that need to be improved here

kasenvr: So if my computer is of low performance and I lower the settings then raise the refresh rate, I will get unintended behavior (to the user)? Because people lower settings to achieve higher framerates. 

They lower framerate targets for screen tearing, quirks in the graphics engine or drivers as they try to hit max FPS possible, or just to help maintain low temperatures and usage on the GPU.

kasenvr: > 
> 
> yeah exactly. basically, when a low end machine can’t hit the target framerate, all sorts of things go wrong. you start missing frames and then work piles up that we couldn’t do in our frame budget and the LOD system will freak out and you won’t be able to see anything. I’d recommend against this change for now, but I agree there’s things that need to be improved here

So if my computer is of low performance and I lower the settings then raise the refresh rate, I will get unintended behavior (to the user)? Because people lower settings to achieve higher framerates.

They lower framerate targets for screen tearing, quirks in the graphics engine or drivers as they try to hit max FPS possible, or just to help maintain low temperatures and usage on the GPU.

HifiExperiments: yeah sorta...*theoretically* you should get a better *experience* when you lower the setting, even if the framerate is lower.

obviously the goal is to have everything work at 60/90 fps though, this was more of a stopgap solution to get Interface working on the type of low end machines that a lot of companies use.  the idea was that they’d be fine with lower frame rates as long as everything else was usable

daleglass: Hmm. Interesting. Okay, let's see if I understand it:

Interface tries to achieve a particular framerate, depending on the performance setting. If it fails to achieve that particular framerate, LOD kicks in and tries to rein things in.

So if I have a machine that can do 35 FPS and set the target to 60, hifi even in desktop mode is going to kick the LOD into high gear, kill a whole bunch of stuff, and the result will be bad.

On the other hand if I set the target to 20 FPS, the LOD will not kick in, and graphically the result will be better.

Am I understanding correctly here?

If so, is a targeted framerate even a good thing to start with on a PC? On VR it's of course of great importance, but why wouldn't simply rendering as fast as it can be done be a better approach? That way, the hardware is used to the fullest ability and you're not leaving a part unused when the target is below the hardware's ability.

HifiExperiments: that’s a good summary, yeah.  and not having a targeted framerate is basically how it used to be, this was added because it vastly improves the experience on low end machines.

it’s basically a question of power.  power = work / time.  to reduce power, you either do less work, or spread it over more time.

daleglass: Could you explain in more detail how the experience is improved?

Specifically the bit about power: are you talking about battery powered devices here?

HifiExperiments: we found that it was better to have a low framerate but lower CPU usage, because you could see more things and your machine wouldn’t start blasting its fans

it sounds more like your machine is getting incorrectly profiled into the low setting?  or that the problem is just after you’ve crashed?  those sound like separate bugs

daleglass: Ok, fans I get. But how do you see more things by passing on using available resources?

Yes, there are two things here:

1. Presets don't behave the way I expect them to behave in a game. I expect this:

https://i.imgur.com/3vVbrrO.png

That is, low quality is high framerate, because each frame takes less work so you can do more of them per second. High quality is low framerate. The game never throttles itself for any reason, except for the lack of computing resources. That, to me, is the way every single game I've seen works and is expected to work. "Low quality" throttling things down is completely backwards from my expectations.

This is historical -- as far back as DOOM and even before it, you could reduce quality to increase speed.

That's what this PR is about -- me trying to bringing it into line with my expectations. 

2. On Linux, perhaps due to GPUIdent.cpp not having Linux code, my machine is auto-configured to "low". This is a Xeon 1245v3 and a GTX 1080. This creates the impression that the renderer is really, really slow on what still is perfectly good hardware. 

I have a separate bug filed for this and been poking at it for a bit.

HifiExperiments: The problem is the in games, the resources and effects are known.  Here, with infinite user generated content, we can’t always guarantee that just turning down some quality settings will make for a good experience.  of course, that was back when it was intended for business use, so maybe the new audience will feel differently.

Either way, yeah, if we’re not profiling Linux devices, that’s definitely a good first step to take, since it’s less risky and won’t affect low end non-Linux devices

daleglass: I've been thinking about this for some time. I think our current team is more one of VR enthusiasts, and as such we probably want to impress people with our quality and performance when this is possible, and turn any power/noise/battery saving settings into something that a minority of users might want to choose by hand, when their circumstances require it.

Here's my revised idea for how I think the settings screen could work in light of this discussion:

* Battery saver: Low quality graphics, throttled down to 20 FPS. Intended to maximize runtime for people using laptops on battery. 
* Low: Low quality graphics, running at up to monitor HZ.
* Medium: Medium quality graphics, running at up to monitor HZ.
* High: High quality graphics, running at up to monitor HZ.

[x] Drop level of detail automatically below 20 FPS.

So the idea here is that trying to maintain a framerate would be only done when it dips below comfortable levels. If you set it to high and you get 40 FPS while rendering everything, that's where it stays. It doesn't try to bring it up to 60, because in desktop mode it's not really that important. Only try to raise the framerate when it gets really bad.

How does that sound?

ctrlaltdavid: Perhaps better labels for the current functionality would be “Low Power”, “Medium Power”, and “High Power”. (As in, “I’ve got a low-powered PC; give me reasonable defaults for it”.)

Also, changes need to be carefully considered because it’s not a straightforward area:
- Understand what the current settings do, and why. (For example, there are differences between desktop and HMD modes.)
- See what improvements or fixes have been made in the HiFi repo since 0.86.
- Identify the bugs. (For example, settings aren’t saved in some situations, …)
- Understand the code. (For example, it looks like both the LODManager API and Performance API are involved and possibly have conflicts.)
- Consider the different users. (For example, desktop vs. HMD x gamers vs. non-gamers).

HifiExperiments: I would suggest:
1. fix the issue where custom settings aren’t remembered
2. profile linux machines correctly

and only once those are done:
3. gather various laptops that fall into the different tiers and test different settings to figure out what we want.  I agree that maybe another mode or a gamer/non-gamer switch for target FPS could be useful, maybe.  but ideally we can set the settings automatically for you to whatever we define as the best experience

daleglass:
Ok, this turned out to be more involved than initially thought, but I'll definitely revisit this eventually.

daleglass:
Note: should address #792

daleglass:
So for reference here's what we have in the dialog right now:

graphics_settings

What should we have? I think in principle it makes sense to offer presets that conserve battery and maximize performance.

daleglass:
After looking into this one, the new solution will be completely different, so I'm closing it.

PR #31, opened 2019-12-19
Audio Failure Due to missing VR fix

FluffyJenkins:
fixes: [https://github.com/kasenvr/project-athena/issues/27](https://github.com/kasenvr/project-athena/issues/27)

Testing:

1. Close SteamVR.
2. Start Interface.
3. Wait until Interface fully loads.

Expected results:
SteamVR gets opened,
Interface doesn't take forever to load due to SteamVR
Audio still works.

### DIFF
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -108,9 +108,11 @@ QString getVrSettingString(const char* section, const char* setting) {
     return result;
 }
 
+int headsetNotConnected = -1;
+
 vr::IVRSystem* acquireOpenVrSystem() {
     bool hmdPresent = vr::VR_IsHmdPresent();
-    if (hmdPresent) {
+    if (hmdPresent && (headsetNotConnected <= 0)) {
         Lock lock(mutex);
         if (!activeHmd) {
             #if DEV_BUILD
@@ -122,6 +124,14 @@ vr::IVRSystem* acquireOpenVrSystem() {
             #if DEV_BUILD
                 qCDebug(displayplugins) << "OpenVR display: HMD is " << activeHmd << " error is " << eError;
             #endif
+
+            if (eError == 108) { // vr::HmdError_Init_HmdNotFound
+                headsetNotConnected = 1;
+                activeHmd = nullptr;
+                #if DEV_BUILD
+                    qCDebug(displayplugins) << "OpenVR: No HMD connected, setting nullptr!";
+                #endif
+            }
         }
         if (activeHmd) {
             #if DEV_BUILD
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -108,11 +108,11 @@ QString getVrSettingString(const char* section, const char* setting) {
     return result;
 }
 
-bool isHMDConnected = false;
+bool isHMDInErrorState = false;
 
 vr::IVRSystem* acquireOpenVrSystem() {
     bool hmdPresent = vr::VR_IsHmdPresent();
-    if (hmdPresent && (!isHMDConnected)) {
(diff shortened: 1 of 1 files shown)


plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -108,9 +108,11 @@ QString getVrSettingString(const char* section, const char* setting) {
     return result;
 }
 
+int headsetNotConnected = -1;
```
thoys: Please use a `bool`     and use the boolean way of writing a variable name see [CODING_STANDARD.md#1.2.10]( https://github.com/kasenvr/project-athena/blob/kasen/core/CODING_STANDARD.md#1210-the-prefix-is-should-be-used-for-boolean-variables-and-methods)  in this case it would be something like `bool isHMDConnected = true`

FluffyJenkins: oops I forgot to add something that actually used it.... I guess it doesn't even need that now as it actually works

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -126,7 +126,7 @@ vr::IVRSystem* acquireOpenVrSystem() {
             #endif
 
             if (eError == 108) { // vr::HmdError_Init_HmdNotFound
-                headsetNotConnected = 1;
+                isHMDConnected = true;
```
thoys: Now it doesn't make sense,   since it wont be connected if the error happens,  hehehehe, flip the bools or rename the bool to `isHMDInErrorState` or so (sorry it's getting late :) )

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -122,6 +124,14 @@ vr::IVRSystem* acquireOpenVrSystem() {
             #if DEV_BUILD
                 qCDebug(displayplugins) << "OpenVR display: HMD is " << activeHmd << " error is " << eError;
             #endif
+
+            if (eError == 108) { // vr::HmdError_Init_HmdNotFound
```
thoys: 108 is a magic number,  please just use `vr::VRInitError_Init_HmdNotFound` instead and remove the comment

FluffyJenkins: I couldn't it kept erroring :c

thoys: note that the constant you were using is different than the one I showed you `vr::HmdError_Init_HmdNotFound`  != `vr::VRInitError_Init_HmdNotFound`     perhaps the docs you were looking at were from another API version?

FluffyJenkins: That is always possible, Eitherway, will do it when VCPKG finishes taking an eternity

plugins/openvr/src/OpenVrHelpers.cpp
```diff
@@ -108,9 +108,11 @@ QString getVrSettingString(const char* section, const char* setting) {
     return result;
 }
 
+bool isHMDInErrorState = false;
+
 vr::IVRSystem* acquireOpenVrSystem() {
     bool hmdPresent = vr::VR_IsHmdPresent();
-    if (hmdPresent) {
+    if (hmdPresent && (!isHMDInErrorState)) {
```
thoys: excessive parentheses
```suggestion
    if (hmdPresent && !isHMDInErrorState) {
```

MarcusLlewellyn:
Tested, ran, and Q&A on Windows and Linux.

Ran this 10 times on Windows with HMD disconnected in an effort to replicate missing audio devices. The audio devices appeared correctly all ten times.

Also tested on Windows to make sure VR still worked as expected when the HMD is plugged in. It does.

Ran successfully on Linux. Did not test VR on Linux.

CMake 3.16.0
SteamVR 1.9.13

gcc/g++ 7.4.0
CMake 3.10.2

PR #32, opened 2019-12-20
Workaround for Secondary camera crash

FluffyJenkins:
This fixes a possible crash with the secondary camera.

It's a hard one to repro cause it doesn't always happen.

But if you're turning the camera on and off a lot it can crash.


kasenvr:
@MarcusLlewellyn haven't you had scripts that use the secondary camera? Could you try and force a crash by rapid firing them or something of the sort? Also just in general confirm there are no adverse effects?

MarcusLlewellyn:
> @MarcusLlewellyn haven't you had scripts that use the secondary camera?

 No, only a camera controller script for the primary camera. The secondary camera API isn't something I'm very familiar with.

kasenvr:
Have we at least built and confirmed there are no immediately known adverse effects?

PR #33, opened 2019-12-20
Ability to change Metaverse API from setting

thoys:
This PR adds a setting for the Metaverse API URL: 

"private/selectedMetaverseURL"

`Settings.getValue("private/selectedMetaverseURL"); // to retrieve the current set value`


libraries/ui/src/ui/types/RequestFilters.cpp
```diff
@@ -25,7 +26,7 @@
 namespace {
 
     bool isAuthableHighFidelityURL(const QUrl& url) {
-        auto metaverseServerURL = NetworkingConstants::METAVERSE_SERVER_URL();
+    auto metaverseServerURL = MetaverseAPI::getCurrentMetaverseServerURL();
```
thoys:
```suggestion
        auto metaverseServerURL = MetaverseAPI::getCurrentMetaverseServerURL();
```

libraries/networking/src/NetworkingConstants.h
```diff
@@ -25,9 +25,8 @@ namespace NetworkingConstants {
     // if you manually generate a personal access token for the domains scope
     // at https://staging.highfidelity.com/user/tokens/new?for_domain_server=true
```
ctrlaltdavid: Code comments above need addressing / updating?

thoys: I don't think at this point, it might be helpful to preserve these comments for a while for Project Apollo. And if we do remove/update them I'd suggest a separate PR for that.

PR #34, opened 2019-12-21
Update a small portion of the build instructions

vegaslon:
Just a small update to build instructions to account for new branding and repository name shift


kasenvr:
Thank you for the changes! These kind of housekeeping things are especially helpful. :)

PR #38, opened 2019-12-25
Feature/quality of life k2

kasenvr:
More changes later, various housekeeping.


interface/src/Menu.cpp
```diff
@@ -772,6 +772,7 @@ Menu::Menu() {
     // Help > About High Fidelity
     action = addActionToQMenuAndActionHash(helpMenu, "About High Fidelity");
     connect(action, &QAction::triggered, [] {
+        DependencyManager::get<OffscreenUi>()->clearCache();
```
thoys: I assume this was for testing? please remove it since it could result in problems.

kasenvr: All done.

PR #41, opened 2019-12-27
Pantheon Launcher - Initial Version

kasenvr:
The launcher can be packaged into an installer for Windows using NSIS. For other platforms like Mac or Linux, there are ways to configure that that I have not yet explored.

image
`


kasenvr:
Due to the launcher being completely separate from the interface + some edits made to make it work with the current installs, it is possible to have a release candidate of K2 without shipping the launcher yet.

kasenvr:
The repository for this application can be found here https://github.com/kasenvr/pantheon-launcher now.

PR #43, opened 2019-12-29
Re-enable JSDoc building

ctrlaltdavid:
API documentation is put in \build\interface\Release\jsdoc directory when Interface is built.


kasenvr:
I built Interface in Release mode with VS2019 and did not find the JSDocs where specified. Is it that I need to do it in Debug mode?

ctrlaltdavid:
I'll update the logo in a separate PR.

PR #49, opened 2019-12-30
Optional SteamVR

FluffyJenkins:
Made Interface not start SteamVR, To use VR you need to start SteamVR first.

Testing:

Start interface without SteamVR running.
Should see no VR icon.
Close interface
Start interface with SteamVR and headset running.
Should see VR icon and VR should work!

### DIFF
--- a/plugins/openvr/src/OpenVrHelpers.cpp
+++ b/plugins/openvr/src/OpenVrHelpers.cpp
@@ -117,7 +117,7 @@ vr::IVRSystem* acquireOpenVrSystem() {
                 qCDebug(displayplugins) << "OpenVR: No vr::IVRSystem instance active, building";
             #endif
             vr::EVRInitError eError = vr::VRInitError_None;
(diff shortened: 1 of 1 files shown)


kasenvr:
Functionality clarification: As I understand it... to connect to SteamVR after Interface is running is not possible. You will have to close Interface, start SteamVR, then open Interface again for it to bind correctly.

MarcusLlewellyn:
Part of me really likes this, But I'm kinda wondering about user expectations. Most VR applications simply start SteamVR (or other runtime) automagically. If this is expected behavior, might we not confuse potential VR users? We could wind up with lots of "Why isn't VR working?" queries.

In addition, our launcher is expected to have an option to disable VR from the start, keeping SteamVR from starting if it isn't already.

jug007:
I guess we should just have an option like "Enable SteamVR on startup" which is ticked by default.

PR #53, opened 2020-01-03
Add support for Ubuntu 19.10, modernize vcpkg

daleglass:
This PR does several, related things:

* Makes it possible to use packages other than .tar.gz (for instance tar.xz and tar.bz2)
* Removes the usage of the deprecated platform.linux_distribution() functions.
* Adds experimental support for Ubuntu 19.10, by using a package I built and that is already uploaded to athena-public.
* Is a bit friendlier in the error reporting

### DIFF
--- a/BUILD_LINUX.md
+++ b/BUILD_LINUX.md
@@ -63,9 +63,9 @@ sudo apt-get install libasound2 libxmu-dev libxi-dev freeglut3-dev libasound2-de
 ```bash
 sudo apt-get -y install libpulse0 libnss3 libnspr4 libfontconfig1 libxcursor1 libxcomposite1 libxtst6 libxslt1.1
 ```
-1.  Install Python 3:
+1.  Install Python 3 and required packages
 ```bash
-sudo apt-get install python3.6
+sudo apt-get install python3 python3-distro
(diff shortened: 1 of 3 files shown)


sitongial:
I have tested Dale's Athena installer on kubuntu 19.10 and it works.

daleglass:
**Testing**:

1. Install Ubuntu 19.10
2. Go through the normal build process
3. Verify that it downloads a Qt package and builds successfully

Possible issues:
* vcpkg fails with a complaint about the 'distro' package. This is fine, and it will need to be installed with the package manager, or pip. The change replaces a deprecated part of the Python API, so this will be needed anyway.
* vcpkg fails with some other error: This is a problem to be fixed.
* The build starts, but there's either an error with using the package for 19.10, or it fails to run after the build. In this case either the 19.10 branch of the code should be removed until the package is fixed.

**Optional**:

1. Install Ubuntu 19.04
2. Go through the normal build process
3. The process should fail with "Sorry, Ubuntu 19.04 is old and won't be officially supported. Please consider upgrading". This is fine because 19.04 isn't a LTS release and support expired at the start of this year.

PR #54, opened 2020-01-03
fix linux build

sethalves:


daleglass:
This needs an explanation. I'm building on Linux and this hasn't been a problem so far. In fact I just ran "make nitpick" against 0bdfdfff, and it builds and starts correctly.

sethalves:
if you `export HIFI_MEMORY_DEBUGGING` when you build, everything will be instrumented with asan.  Without this patch, the build fails.

sethalves:
Here is the cmake line I use (which fails on master): `cmake  .. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDOWNLOAD_SERVERLESS_CONTENT=ON -DOpenGL_GL_PREFERENCE=GLVND -DBUILD_SHARED_LIBS=ON`

sethalves:
the first patch is because the Qt version you get with asan can't cope with that DEBUG_EVENT_QUEUE crap, and the 2nd patch is because if you build with shared libraries, nitpick fails to link.

PR #55, opened 2020-01-03
Fix scaled avatar walk speed

ctrlaltdavid:
The walk speed of your avatar is now proportional to your avatar's scale: small scales no longer run instead of walk; large scales no longer walk s l o w l y.

Tested:
- Desktop keyboard: walking (fixed), flying.
- Desktop gamepad: walking (fixed), flying.
- HMD walking, flying.

Desktop keyboard and gamepad walking were both broken; both are now fixed.

Issue: https://github.com/kasenvr/project-athena/issues/17


Darlingnotin:
Tested with desktop keyboard and Xbox controller. And with both in hmd. I didn't encounter any bugs after scaling. And the walking speed seems to be working.

PR #64, opened 2020-01-05
inspect.js improvements

ctrlaltdavid:
- Shorter initial look-at duration.
- Leave camera where it is after releasing the Alt key.
- Press Esc or move avatar to exit inspect mode.
- Don't go "away" when press Esc to exit inspect mode.
- Orbit about point in space if nothing intersected.
- Ignore invisible entities.
- Fix jitter when start orbiting entity with Create app focused on entity.
- Use Picks API.


PedroBauer:
The four points work on my side without any issue.

kasenvr:
This happened to me with another version of inspect.js, but it is the case that when I use a camera mode to exist inspect (num pad 1, 2, or 3) instead of "esc" then the inspect mode will refuse to let me reengage it unless I move first.

ctrlaltdavid:
@kasenvr Can now re-engage inspect mode after using camera mode to exit.

MarcusLlewellyn:
This seems cromulent, so I'm gonna go ahead and merge this.

PR #65, opened 2020-01-06
Whitelist Quality of Life Updates

kasenvr:
This PR does the following: 

- Disables the whitelist on the server scripts.
- File:///, cache:, and atp: are hardcoded as defaults
- You can whitelist an entire domain e.g. "hifi://thepalace"
- A setting has been added to enable or disable the whitelist, however an interface toggle has not yet been added.

To test: 

- Build a server from Athena's master, try to run entity server scripts and check the log to see if the whitelist catches them. If there is no whitelist output, it should indicate they are getting through.
- Try to load a script from the file system and from the ATP, they should both get through without hassle.
- Try to whitelist a domain with many scripts that you have not whitelisted on it, they should be accepted.
- Modify the setting for "private/whitelistEnabled" and set it to false, try to load various scripts. Then reenable it and see if they get blocked. I have not done checks on if cached scripts get through or not. This may still be an issue.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -2361,10 +2361,9 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         }
     }
     else {
-		// ENTITY SCRIPT WHITELIST STARTS HERE
+        // ENTITY SCRIPT WHITELIST STARTS HERE
         auto nodeList = DependencyManager::get<NodeList>();
-        bool isInWhitelist = false;  // assume unsafe
-        bool autoPassList = false;
+        bool passList = false;  // assume unsafe
         QString whitelistPrefix = "[WHITELIST ENTITY SCRIPTS]";
         QList<QString> safeURLS = { "file:///", "atp:", "cache:" };
         safeURLS += qEnvironmentVariable("EXTRA_WHITELIST").trimmed().split(QRegExp("\\s*,\\s*"), QString::SkipEmptyParts);
@@ -2372,14 +2371,13 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         // IF WHITELIST IS DISABLED IN SETTINGS
         bool whitelistEnabled = Setting::Handle<bool>("private/whitelistEnabled", true).get();
         if (!whitelistEnabled) {
-            autoPassList = true;
+            passList = true;
         }
-        // PULL SAFEURLS FROM INTERFACE.JSON Settings
         
+        // PULL SAFEURLS FROM INTERFACE.JSON Settings
         QVariant raw = Setting::Handle<QVariant>("private/settingsSafeURLS").get();
         QStringList settingsSafeURLS = raw.toString().trimmed().split(QRegExp("\\s*[,\r\n]+\\s*"), QString::SkipEmptyParts);
         safeURLS += settingsSafeURLS;
-        
         // END PULL SAFEURLS FROM INTERFACE.JSON Settings
         
         // GET CURRENT DOMAIN WHITELIST BYPASS, IN CASE AN ENTIRE DOMAIN IS WHITELISTED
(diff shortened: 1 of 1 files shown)


Darlingnotin:
Tested changes. Changes allowed for server entity Scripts to run on the server as they should. Further testing to make sure that it did not adversely affect the interface waitlist didn't seem to have any adverse effects from this change.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2377,15 +2377,21 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
+				qCDebug(scriptengine) << whitelistPrefix << "Script URL: " << scriptOrURL << "TESTING AGAINST" << str << "RESULTS IN"
+						 << scriptOrURL.startsWith(str);
+				if (!str.isEmpty() && scriptOrURL.startsWith(str)) {
+					isInWhitelist = true;
+					qCDebug(scriptengine) << whitelistPrefix << "Script approved.";
+					break;  // bail early since we found a match
+				}
+			}
+		}
+
```
HifiExperiments: tabs instead of spaces

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2377,15 +2377,21 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
-            qCDebug(scriptengine) << whitelistPrefix << "Script URL: " << scriptOrURL << "TESTING AGAINST" << str << "RESULTS IN"
-                     << scriptOrURL.startsWith(str);
-            if (!str.isEmpty() && scriptOrURL.startsWith(str)) {
-                isInWhitelist = true;
-                qCDebug(scriptengine) << whitelistPrefix << "Script approved.";
-                break;  // bail early since we found a match
-            }
-        }
+		
+		if(ScriptEngine::getContext() == "entity_server") {
```
thoys:
```suggestion
		if (ScriptEngine::getContext() == "entity_server") {
```

vegaslon:
The server already has a client entity script/server entity script white list setup in the domain settings so this change is a good idea.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,11 +2361,19 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         }
     }
     else {
-      // ENTITY SCRIPT WHITELIST STARTS HERE
+		// ENTITY SCRIPT WHITELIST STARTS HERE
```
HifiExperiments: tabs

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2376,17 +2382,37 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
+                if (!str.isEmpty() && scriptOrURL.startsWith(str)) {
+                    isInWhitelist = true;
+                    qCDebug(scriptengine) << whitelistPrefix << "Script approved.";
+                    break; // bail early since we found a match
+                }
             }
         }
-        if (!isInWhitelist) {
+
+        if (!isInWhitelist || autoPassList == false) {
```
HifiExperiments: do you need both isInWhitelist and autoPassList?  seems like they could be combined

also, should this be !isInWhitelist && !autoPassList

kasenvr: Makes sense, I've made adjustments.

Darlingnotin:
Tested this. Couldn't find any glaring problems after Testing.
Can't wait into this one's added in. I will be using this one a lot.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,30 +2361,57 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
+                passList = true;
             }
         }
-        if (!isInWhitelist) {
+        // END CURRENT DOMAIN WHITELIST BYPASS
+
+        // START CHECKING AGAINST THE WHITELIST
+        if (ScriptEngine::getContext() == "entity_server" || passList == true) { // If running on the server or waved through, do not engage whitelist.
+            passList = true;
+        } else {
```
HifiExperiments: instead of checking ‘passList == true’ above, this could be:

‘} else if (!passList) {‘

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,30 +2361,57 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
         safeURLS += settingsSafeURLS;
-        
         // END PULL SAFEURLS FROM INTERFACE.JSON Settings
         
-        bool isInWhitelist = false;  // assume unsafe
+        // GET CURRENT DOMAIN WHITELIST BYPASS, IN CASE AN ENTIRE DOMAIN IS WHITELISTED
+        QString currentDomain = DependencyManager::get<AddressManager>()->getDomainURL().host();
+        
+        QString domainSafeIP = nodeList->getDomainHandler().getHostname();
+        QString domainSafeURL = "hifi://" + currentDomain;
```
thoys: I believe there is already a constant with the string "hifi://"  somewhere in the code.. It would be easier to replace in the future if we keep those definitions in the same place.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,30 +2361,57 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         }
     }
     else {
```
thoys: Someone messed up the else here can you fix it back to `} else {` while you're touching this file anyway?

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,30 +2361,57 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
...
-      // ENTITY SCRIPT WHITELIST STARTS HERE
+        // ENTITY SCRIPT WHITELIST STARTS HERE
+        auto nodeList = DependencyManager::get<NodeList>();
+        bool passList = false;  // assume unsafe
         QString whitelistPrefix = "[WHITELIST ENTITY SCRIPTS]";
-        QList<QString> safeURLS = { "" };
+        QList<QString> safeURLS = { "file:///", "atp:", "cache:" };
```
thoys: Could be renamed to `safeURLPrefixes` or something similar to make it more clear that it's only about how a URL starts

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -2363,30 +2361,57 @@ void ScriptEngine::entityScriptContentAvailable(const EntityItemID& entityID, co
         }
     }
     else {
-      // ENTITY SCRIPT WHITELIST STARTS HERE
+        // ENTITY SCRIPT WHITELIST STARTS HERE
+        auto nodeList = DependencyManager::get<NodeList>();
+        bool passList = false;  // assume unsafe
```
thoys: since this whole code section below is about if we pass the whitelist or not, it would make less complexity if you add another function (maybe in its own class) for checking a URL.  That way you can simply return true or false out of the function instead of making all these boolean checks with `passList`

kasenvr: This is a good idea, helps with making the codebase cleaner but also makes the whitelist more accessible. Will probably do this in a later PR. :)

Darlingnotin:
Seemed to work fine.
After testing the only thing that it didn't block was atp Scripps. And it didn't seem to have any adverse side effects.
I'm not sure if it was supposed to block atp Scripts.

Darlingnotin:
I was informed that it is not supposed to block atp scripps so it worked perfectly. Good to go.

PR #66, opened 2020-01-07
Added ArrayBuffer support to websockets

FluffyJenkins:
Testplan:

Run the below code in script console with PR build,
result should look like this

If it looks like this
then you don't have arraybuffer support!

Code to run:
```
var ws = new WebSocket('wss://echo.websocket.org');
ws.binaryType = 'arraybuffer';
ws.onmessage = function(evt) {
    if (evt.data instanceof ArrayBuffer) {
       console.info('received binary', evt.data.byteLength, 'bytes');
       console.info('as floats', [].slice.call(new Float32Array(evt.data)));
    } else {
       console.info('received text', evt.data.length, 'chars', JSON.stringify(evt.data));
    }       
};
ws.onopen= function() { 
    console.info('connected', ws.url);
    ws.send('._.');
    ws.send(new Float32Array([0xFACE,Math.PI]).buffer)
    ws.send('._.');
};
```

### DIFF
--- a/libraries/script-engine/src/WebSocketClass.cpp
+++ b/libraries/script-engine/src/WebSocketClass.cpp
@@ -16,6 +16,8 @@
 
 #include "ScriptEngine.h"
 
+#include "ScriptEngineLogging.h"
+
 WebSocketClass::WebSocketClass(QScriptEngine* engine, QString url) :
     _webSocket(new QWebSocket()),
     _engine(engine)
@@ -109,10 +111,10 @@ void WebSocketClass::handleOnBinaryMessage(const QByteArray& message) {
         QScriptValue arg = _engine->newObject();
         QScriptValue data = _engine->newVariant(QVariant::fromValue(message));
         QScriptValue ctor = _engine->globalObject().property("ArrayBuffer");
-        ArrayBufferClass* array = qscriptvalue_cast<ArrayBufferClass*>(ctor.data());
+        auto array = qscriptvalue_cast<ArrayBufferClass*>(ctor.data());
         QScriptValue arrayBuffer;
         if (!array) {
-            qWarning() << "WebSocketClass::handleOnBinaryMessage !ArrayBuffer";
+            qCWarning(scriptengine) << "WebSocketClass::handleOnBinaryMessage !ArrayBuffer";
         } else {
             arrayBuffer = _engine->newObject(array, data);
         }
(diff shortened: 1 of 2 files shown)


Darlingnotin:
Follow the test plan and results of test reflected the one that worked correctly.
Test plan went as it should. 

Capture

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -34,10 +34,11 @@ WebSocketClass::WebSocketClass(QScriptEngine* engine, QWebSocket* qWebSocket) :
 void WebSocketClass::initialize() {
     connect(_webSocket, &QWebSocket::disconnected, this, &WebSocketClass::handleOnClose);
     connect(_webSocket, &QWebSocket::textMessageReceived, this, &WebSocketClass::handleOnMessage);
+    connect(_webSocket, &QWebSocket::binaryMessageReceived, this, &WebSocketClass::handleOnBinaryMessage);
     connect(_webSocket, &QWebSocket::connected, this, &WebSocketClass::handleOnOpen);
     connect(_webSocket, static_cast<void(QWebSocket::*)(QAbstractSocket::SocketError)>(&QWebSocket::error), this,
         &WebSocketClass::handleOnError);
-    _binaryType = QStringLiteral("blob");
+    _binaryType = QStringLiteral("arraybuffer");
```
ctrlaltdavid: This doesn't actually do anything because `_binaryType` isn't used. It would be good to sometime support `"blob"` vs. `"arraybuffer"` binary result types but need not be done now.

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -97,6 +103,25 @@ void WebSocketClass::handleOnMessage(const QString& message) {
     }
 }
 
+void WebSocketClass::handleOnBinaryMessage(const QByteArray& message) {
+    if (_onMessageEvent.isFunction()) {
+        QScriptValueList args;
+        QScriptValue arg = _engine->newObject();
+        QScriptValue data = _engine->newVariant(QVariant::fromValue(message));
+        QScriptValue ctor = _engine->globalObject().property("ArrayBuffer");
+        ArrayBufferClass* array = qscriptvalue_cast<ArrayBufferClass*>(ctor.data());
```
thoys: could use `auto` here to prevent duplicate `ArrayBufferClass*`

libraries/script-engine/src/WebSocketClass.cpp
```diff
@@ -97,6 +103,25 @@ void WebSocketClass::handleOnMessage(const QString& message) {
...
+        QScriptValue arrayBuffer;
+        if (!array) {
+            qWarning() << "WebSocketClass::handleOnBinaryMessage !ArrayBuffer";
```
thoys: Would you still want to send an event when there is no data present? or could we return here?

ctrlaltdavid: I'd say still send the event: Interface has received an event from the socket so we should pass that info on to the script, even if there is not data.

thoys: 👍

thoys: Can you make this warning more helpful and set a proper category with qCWarning ?

PR #68, opened 2020-01-07
Fix thousands of 'is too small to hold all values' warnings on Linux

daleglass:
When building with GCC, it generates 3020 warnings like this:

libraries/gpu/src/gpu/State.h:208:18: warning: ‘gpu::State::BlendFunction::destColor’
is too small to hold all values of ‘enum gpu::State::BlendArg’

This is because ‘enum gpu::State::BlendArg’ is declared as an enum
of uint16 size, and gpu::State::BlendFunction::destColor is a bit
field.

GCC correctly deduces that a 16 bit value won't always fit in a 4
bit field, and emits a warning.

The problem is that the amount is such that it floods the output and hides
other warnings.

Changing the 'enum foo : uint16' declarations to 'enum foo' stops gcc
from emitting the warning.

Making this change required the removal of a set of assertion checks. At least
empirically they seem unnecessary -- the interface compiles and runs fine.

### DIFF
--- a/libraries/gpu/src/gpu/State.h
+++ b/libraries/gpu/src/gpu/State.h
@@ -136,25 +136,26 @@ public:
         uint8 getWriteMask() const { return writeMask; }
 
         bool operator==(const DepthTest& right) const {
-            return writeMask == right.writeMask && enabled == right.enabled && function == right.function;
+            return
+                    writeMask == right.writeMask &&
+                    enabled   == right.enabled &&
+                    function  == right.function;
         }
 
         bool operator!=(const DepthTest& right) const {
-            return writeMask != right.writeMask || enabled != right.enabled || function != right.function;
+            return !(right == *this);
         }
 
         operator QString() const {
             return QString("{ writeMask = %1, enabled = %2, function = %3 }").arg(writeMask).arg(enabled).arg(function);
-
         }
-
     };
 
     struct StencilTest {
-        ComparisonFunction function : 4;
-        StencilOp failOp : 4;
-        StencilOp depthFailOp : 4;
-        StencilOp passOp : 4;
+        ComparisonFunction function;
+        StencilOp failOp;
+        StencilOp depthFailOp;
+        StencilOp passOp;
         int8 reference{ 0 };
         uint8 readMask{ 0xff };
 
@@ -176,47 +177,56 @@ public:
         int8 getReference() const { return reference; }
         uint8 getReadMask() const { return readMask; }
 
-        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
-        StencilTest(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
-        bool operator==(const StencilTest& right) const { return getRaw() == right.getRaw(); }
-        bool operator!=(const StencilTest& right) const { return getRaw() != right.getRaw(); }
+        bool operator==(const StencilTest& right) const {
+            return
+                    function    == right.function &&
+                    failOp      == right.failOp &&
+                    depthFailOp == right.depthFailOp &&
+                    passOp      == right.passOp &&
+                    reference   == right.reference &&
+                    readMask    == right.readMask;
+
+        }
+
+        bool operator!=(const StencilTest &right) const { return !(right==*this); }
     };
-    static_assert(sizeof(StencilTest) == sizeof(uint32_t), "StencilTest size check");
 
     StencilTest stencilTestFront;
 
     struct StencilActivation {
         uint8 frontWriteMask = 0xFF;
         uint8 backWriteMask = 0xFF;
-        bool enabled : 1;
-        uint8 _spare1 : 7;
-        uint8 _spare2{ 0 };
+        bool enabled;
 
     public:
         StencilActivation(bool enabled = false, uint8 frontWriteMask = 0xFF, uint8 backWriteMask = 0xFF) :
-            frontWriteMask(frontWriteMask), backWriteMask(backWriteMask), enabled(enabled), _spare1{ 0 } {}
+            frontWriteMask(frontWriteMask), backWriteMask(backWriteMask), enabled(enabled) {}
 
         bool isEnabled() const { return enabled; }
         uint8 getWriteMaskFront() const { return frontWriteMask; }
         uint8 getWriteMaskBack() const { return backWriteMask; }
 
-        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
-        StencilActivation(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
-        bool operator==(const StencilActivation& right) const { return getRaw() == right.getRaw(); }
-        bool operator!=(const StencilActivation& right) const { return getRaw() != right.getRaw(); }
-    };
+        bool operator==(const StencilActivation& right) const {
+            return
+                frontWriteMask == right.frontWriteMask &&
+                backWriteMask  == right.backWriteMask &&
+                enabled        == right.enabled;
+        }
 
-    static_assert(sizeof(StencilActivation) == sizeof(uint32_t), "StencilActivation size check");
+        bool operator!=(const StencilActivation& right) const {
+            return !(right == *this);
+        }
+    };
 
     struct BlendFunction {
         // Using uint8 here will make the structure as a whole not align to 32 bits
-        uint16 enabled : 8;
-        BlendArg sourceColor : 4;
-        BlendArg sourceAlpha : 4;
-        BlendArg destColor : 4;
-        BlendArg destAlpha : 4;
-        BlendOp opColor : 4;
-        BlendOp opAlpha : 4;
+        uint16 enabled;
+        BlendArg sourceColor;
+        BlendArg sourceAlpha;
+        BlendArg destColor;
+        BlendArg destAlpha;
+        BlendOp opColor;
+        BlendOp opAlpha;
 
     public:
         BlendFunction(bool enabled,
@@ -227,7 +237,7 @@ public:
                       BlendOp operationAlpha,
                       BlendArg destinationAlpha) :
             enabled(enabled),
-            sourceColor(sourceColor), sourceAlpha(sourceAlpha), 
+            sourceColor(sourceColor), sourceAlpha(sourceAlpha),
             destColor(destinationColor), destAlpha(destinationAlpha),
             opColor(operationColor), opAlpha(operationAlpha) {}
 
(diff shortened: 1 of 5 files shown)


daleglass:
Yeah, don't merge this yet without a lot of testing. It's not exploding for me, but I'm not quite sure of everything this touches.

I'm looking into where this gets used.

daleglass:
Yes, doh. This is pretty much certainly wrong as it is, but what I still have not clear if it's fixable.

So for instance in DepthTest we have:


```
        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
        DepthTest(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
        bool operator==(const DepthTest& right) const { return getRaw() == right.getRaw(); }
        bool operator!=(const DepthTest& right) const { return getRaw() != right.getRaw(); }
```

From those, ```DepthTest(int32 raw)``` seems unnecessary, nothing breaks by removing it.

```operator==``` and ```operator!=``` are easily fixed.

The trouble is with getRaw(), which does get used in several places. ```libraries/gpu-gl-common/src/gpu/gl/GLState.cpp:122``` looks like a possibly troublesome place.

Looking at OpenGL docs, there doesn't seem to be any requirement in the API to pack stuff into 32 bits. And indeed, the ```GLBackend::do_setStateDepthTest(State::DepthTest test)``` function that uses this class to call GL functions just accesses the various member member functions of the DepthTest class.

So, so far it seems that all this packing the class into 32 bits business isn't strictly necessary, but I'm still not clear on why it's being done -- optimization? implementation convenience?

**Comment by Nex-Pro** [comment]
Could we direct this to Sam?

**Comment by daleglass** [comment]
I think that's Sam commenting up there, and this was in part a result of a discussion with him on Discord.

In any case, this is definitely a "don't merge" sort of thing for now, more of a for now half-baked idea to seek feedback and testing on various platforms.

Based on my research this should be a doable change, but it probably needs more work still.

**Comment by Nex-Pro** [comment]
How would we be confident that any testing was comprehensive enough to avoid problems down the line ?

**Comment by HifiExperiments** [comment]
the depth test stuff is only used in a few places.  basically, if things are depth testing properly, we’re probably fine, but it’s good to be extra sure about places that are obviously problematic

for example, in the case of getRaw, is the current code even doing what we expect?  there’s a number of places in the code were we do similar things with reinterpret_cast and I’ve always been suspicious of them.  isn’t that really checking that the two objects are the same object (the pointers are equal, deep equality) instead of that they represent the same DepthTest (a shallow comparison)?  don’t we really want shallow equality?

**Comment by daleglass** [comment]
The getRaw() seems to be doing the right thing, sorta:

    int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }

"this" is a pointer, so reinterpret_cast<const int32*>(this) is "pretend this pointer to a class is a pointer to an int32", and the outer * dereferences that. So what's getting compared to is the 32 bits starting at the address of the instance, not pointers.

Where I think this is very fishy is that I don't think you're supposed to play such games with classes, and I don't think C++ guarantees that the first thing to be found at the pointer to a class is going to be the first member -- it might well be something else, like the vtable if there is one.

So far this just looks like an excess of optimization to me. I'm working on getting rid of all of it, and then I'll see if anything breaks or there's a visible performance impact. My inclination is that such optimizations should be avoided if possible, lest it run into some compiler oddity.

**Comment by daleglass** [comment]
Okay, I have dealt with DepthTest in that last commit. I removed the getRaw() and the reinterpret_cast stuff, and made it into something more normal and boring.

I wanted to make the data private, but that breaks some macros and seems outside of the scope of fixing warnings.

I removed State::getKey() because it's not called from anywhere anyway, and its purpose is completely unclear. Might have had an use before?

So far things compile, build, and there's a lot less warnings now. I would say not ready for a merge yet, but it doesn't seem to be breaking things, and all this data packing business doesn't seem to have any relationship with networking or OpenGL, and to exist just for the sake of optimization.

**Comment by daleglass** [comment]
Okay, that last commit should finish the changes. So in the course of doing this I gained more understanding of what was going on, so I'm going to resume everything in one comment.

When building with gcc, it generates 3020 warnings like this:

    libraries/gpu/src/gpu/State.h:208:18: warning: ‘gpu::State::BlendFunction::destColor’ is too small to hold all values of ‘enum gpu::State::BlendArg’

This is because `enum gpu::State::BlendArg` is declared as an enum of uint16 size, and `gpu::State::BlendFunction::destColor` is a bit field of 4 bits. GCC correctly deduces that a 16 bit value won't always fit in a 4 bit field, and emits a warning. The problem is that in C++ an enum is ultimately a numeric value. Though you might have declared an enum with only 4 names with values from 0 to 3, nothing stops you from assigning the value 42 to that variable.

Also, assuming we go just by the values declared in the enum, we have count values like NUM_STENCIL_OPS, that can exceed the maximum range, though that's not an actual value that shouldn't ever be given to the variable. But gcc can't know that.

So, this is very much a reasonable warning, even though there's really no bug because nothing actually goes out of range. Gcc can't really tell that won't ever happen, though.


Now, why is the code doing this at all? Why the bit fields?

`State.h` has an optimization: it has a number of classes, take for instance `StencilTest` here: 
https://github.com/kasenvr/project-athena/blob/98a5ec84a9ac163c18593503aa5282b0a1708d2d/libraries/gpu/src/gpu/State.h#L145

It packs all its data into 32 bits, and then uses reinterpret_cast to turn copying and comparison into operations on a 32 bit integer. 

This scheme has some upsides:

* It uses memory very efficiently
* It turns copies and comparisons into operations on 32 bit ints
* The associated asserts make it hard to screw things up by modifying the classes. The assets will fail if the required adjustments aren't made.

There are downsides, though:

* The main one is the warning. The compiler is right in being suspicious. 3020 errors is a huge bother that can easily hide something actually important.
* It's possibly slower in practice, due to having to unpack this structure to make use of the data in it.
* It relies on the class not having anything extra. This for instance wouldn't work with a class with a vtable.
* It only works so long the class can be packed into 32/64 bits, since there needs to be a data type capable of holding the entire class.
* It's a bit overly magical. I'm not sure if the C++ standard guarantees that the first member of a class has to be at offset 0. This might well not work somewhere.
* It requires a fair amount of thought to understand what's going on. I think simplicity is a good thing to have, and makes the codebase more accessible.
* Adding more values is fiddly as it requires adjusting the scheme.

**Comment by daleglass** [comment]
Testing this change:

* Compile on both Linux and Windows
* Verify that it removes warnings (on Linux at least) and that it doesn't introduce new ones.
* Verify the interface runs, and that rendering works correctly.

This change touches on fundamental parts of rendering, like depth sorting. If it breaks something, it should be easily visible.

File: libraries/gpu/src/gpu/State.h [hunk 1]
```diff
@@ -134,19 +135,27 @@ class State {
...
-    };
+        bool operator==(const DepthTest& right) const {
+            return
+                    writeMask == right.writeMask &&
+                    enabled   == right.enabled &&
+                    function  == right.function;
+        }
+
+        bool operator!=(const DepthTest& right) const {
+            return !(right == *this);
```
thoys: if you call it right, please put it right :)

libraries/gpu/src/gpu/State.h
```diff
@@ -168,47 +177,56 @@ class State {
         int8 getReference() const { return reference; }
         uint8 getReadMask() const { return readMask; }
 
-        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
-        StencilTest(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
-        bool operator==(const StencilTest& right) const { return getRaw() == right.getRaw(); }
-        bool operator!=(const StencilTest& right) const { return getRaw() != right.getRaw(); }
+        bool operator==(const StencilTest& right) const {
+            return
+                    function    == right.function &&
```
thoys: Spacing looks nice and all, but it's not in the standard of one space

libraries/gpu/src/gpu/State.h
```diff
@@ -134,19 +135,27 @@ class State {
...
 
-        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
-        DepthTest(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
-        bool operator==(const DepthTest& right) const { return getRaw() == right.getRaw(); }
-        bool operator!=(const DepthTest& right) const { return getRaw() != right.getRaw(); }
-    };
+        bool operator==(const DepthTest& right) const {
+            return
+                    writeMask == right.writeMask &&
+                    enabled   == right.enabled &&
```
thoys: alignment looks nice, but unfortunately not following the code style of the rest of the project

libraries/gpu/src/gpu/State.h
```diff
@@ -168,47 +177,56 @@ class State {
...
 
-        int32 getRaw() const { return *(reinterpret_cast<const int32*>(this)); }
-        StencilActivation(int32 raw) { *(reinterpret_cast<int32*>(this)) = raw; }
-        bool operator==(const StencilActivation& right) const { return getRaw() == right.getRaw(); }
-        bool operator!=(const StencilActivation& right) const { return getRaw() != right.getRaw(); }
-    };
+        bool operator==(const StencilActivation& right) const {
+            return
+                frontWriteMask == right.frontWriteMask &&
+                backWriteMask  == right.backWriteMask &&
```
thoys: 1 space

libraries/gpu/src/gpu/State.h
```diff
@@ -236,32 +254,51 @@ class State {
...
+                    sourceAlpha  == right.sourceAlpha &&
+                    destColor    == right.destColor &&
+                    destAlpha    == right.destAlpha &&
+                    opColor      == right.opColor &&
+                    opAlpha      == right.opAlpha;
+
+        }
+
+        bool operator!=(const BlendFunction& right) const {
+            return !(right == *this);
```
thoys: param name is right, but you compare it on the left side

libraries/gpu/src/gpu/State.h
```diff
@@ -236,32 +254,51 @@ class State {
...
+        bool depthClampEnable;
+        bool scissorEnable;
+        bool multisampleEnable;
+        bool antialisedLineEnable;
+        bool alphaToCoverageEnable;
+
+
+        bool operator==(const Flags& right) const {
+            return
+                    frontFaceClockwise    == right.frontFaceClockwise &&
```
thoys: alignment is nice, but one space is the style

daleglass:
Ok, style fixed.

PR #78, opened 2020-01-12
Zone occlusion

HifiExperiments:
All entities have a new property `renderWithZones`, a vector of Uuids.
An entity will only render if:
1. it does not specify this property
2. it does specify the property and your avatar is currently inside one of the zones indicated **(currently, the zone also has to be marked as visible.  is that desirable?)**

**This PR does not expose the new property to Create.  There's not currently any other list property like this that's exposed and I'm not sure how best to do it.**

This PR also does not handle physics...although the entities stop rendering, they still affect you physically.  I don't see any way around this...now, two people can be seeing different things, but they need to agree on the physics of everything.

Test plan:
- Go somewhere with no zones and run [this script](https://gist.githubusercontent.com/HifiExperiments/24cbe3779d20ed5cc30966a9aa16fe16/raw/10731f9d69b4555c5571a1b62aadde79ed2e2cfc/ZoneOcclusionTest.js).  You'll see:
image
- Press "o".  You'll see:
image
- Press "o".  You'll see:
image
- Press "o".  You'll see:
image
- Press "i" and go through all the zones by pressing "o" multiple times again.  The entities will change to draw in front, but otherwise you'll see the same things.


kasenvr:
Thinking about attaching a prototype HTML+CSS+JS UI to the create menu to enable this setting from the interface whilst also paving a way towards having a non-QML dependent UI system.

ctrlaltdavid:
One possibility for adding the property to Create: In the Create app, present the list as a comma-delimited string, in a text area.

HifiExperiments:
@ctrlaltdavid I think that leaves a lot of room for user error.  I feel like we should be able to help people, like only let them add valid zone entities, and maybe providing a button to automatically set an entity to only render with the zones it is currently in.

kasenvr:
> 
> 
> @ctrlaltdavid I think that leaves a lot of room for user error. I feel like we should be able to help people, like only let them add valid zone entities, and maybe providing a button to automatically set an entity to only render with the zones it is currently in.

I think if we take that into account, code we write as assist tools for that should be forward thinking so that they can be reused to improve the whole create app. e.g. something that takes from the list of entities and displays only the zones, but in a drop down, or maybe a window dialog for you to select each (when you click it pushes that zone ID into the box, or maybe the name + zone ID, name being ignored by the system when passed though so you can confirm against bugs and user error that you were performing what you thought you were.)

ctrlaltdavid:
@HifiExperiments It does leave a lot of room for user error, however it could be an initial way to enable the new functionality to be used without having to write a script to edit entities that you want to set the zone thing for.

Aitolda:
Is it possible this could be done without a protocol change?

HifiExperiments:
@Aitolda no, anytime entity properties are added/removed/changed, the protocol must be changed

@ctrlaltdavid @kasenvr I think we should merge this as is and tackle the UI separately, I don’t know if I’m the right person to do it and I think it needs more discussion

AleziaKurdis:
I have started to see how to put in place a zone selector in the "Create" app. So far I have been able to expose the list of the Zones in the entityProperties.js
This PR can definitely released without the change in Create App.

AleziaKurdis:
I developed a new type of field: **Multiple Zones Selector** 
_(For the demo I plugged temporarily on another property. )_

Here in **Edit mode**:
MultipleZonesSelector-Edit

When **locked** (Read-only):
MultipleZonesSelector-Locked

This is supporting **multiple selections of entities**:
MultipleZonesSelector-MULTIPLE-EDIT

**To complete, I would need:**
1- A **PR Build** with this new property to be able to add the new property with my selector
(Sorry, I didn't have the patience to make the product build. I'm more a scripter for now.)
2- The text for the "**Label**".
3- The text for the "**Tooltip**".

Also:
**I was wondering if this new property is ok to be available even on "Zone" entities. Is it supporting this?**

MarcusLlewellyn:
Unofficial PR build for Windows is available here: [ProjectAthena-Alpha-PR78-5fdd563.exe](https://cdn.vircadia.com/dist/pr-builds/ProjectAthena-Alpha-PR78-5fdd563.exe)

kasenvr:
With regards to the issues on this PR, we should generally test it and try to break it. If we can't reasonably break it, we should simply merge it and continue using it in our RC releases for K3, hopefully then any major issues will be found that way because then we'll have more motivated testers using it normally instead of testing an isolated PR build.

Aitolda:
Test plan works as expected.  Will try combining this PR with Alezia's create alterations tomorrow.

PR #99, opened 2020-01-16
Removed "Marketplace Item Tester" from developer menu.(interface)

Nex-Pro:
To test, check that "Test Marketplace Item" is not present on the developer menu

### DIFF
--- a/scripts/system/commerce/wallet.js
+++ b/scripts/system/commerce/wallet.js
@@ -632,7 +632,7 @@ function isReturnedDataEmptyHistory(data) {
     var historyArray = data.data.history;
     return historyArray.length === 0;
 }
-
+/*
 var DEVELOPER_MENU = "Developer";
 var MARKETPLACE_ITEM_TESTER_LABEL = "Marketplace Item Tester";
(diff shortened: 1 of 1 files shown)


Nex-Pro:
Currently code is simply commented out. We should probably remove the code and the referenced QML files for the app.

thoys:
I see this code comments a line in the wallet code.  Since the menu item gets created in the wallet, wouldn't it make more sense to just exclude the wallet from the default scripts?

Nex-Pro:
@thoys hence my comment

Nex-Pro:
but this to remove the tester

Nex-Pro:
we don't want to remove the wallet AFAIK

MarcusLlewellyn:
From a style standpoint, the ugly thing here is the giant block of commented code. I'd remove the block comments at lines 635 and 665. As long as the functions called from within that block are commented out  (lines 704 and 728), maybe with an additional `// UNDONE: <REASON>` task comment for those lines, I'd call it okay.

Although we do not want to remove the wallet at this point, it will inevitably go away in the future, so sweating the details probably isn't worth our time.

thoys:
This menu item will only display for people that run the wallet script. It is a thing that only shows up in the developer menu. Why does it bother you?

Nex-Pro:
Seems this test wasn't needed.

PR #101, opened 2020-01-17
Proposed rpm build spec

odysseus654:
Proposed rpm spec along with systemd service files.

Code is stored in /usr/share/athena
Data is stored in /var/lib/athena

This configuration (optionally) supports multiple instances per server through the use of systemd template files.  New instances can be created through /etc/athena/new-instance.  On installation an instance named "default" is created automatically (if it does not currently exist).  When using the instance mechanism environment variables can be specified in /etc/athena/{instance}


kasenvr:
Odysseus says works for Amazon Linux. Supposedly not that much work to get it packaged into .deb

odysseus654:
I'm guessing the biggest issue generating for a different distribution is checking for any idiosyncrasies where it puts files in different places.  I'm not expecting that much really though.

I've renamed the folder in the PR from "rpm" to "pkg-scripts" as I'm starting to suspect I/we might want to throw specs in there for other distributions as well (while reusing the systemd files)

kasenvr:
Readme needs some refinement and it needs to be converted to an MD file, but so far things work as expected to generate a .deb

PR #104, opened 2020-01-18
Make default copy constructors explicit

daleglass:
Recent versions of gcc generate a warning without this. This follows the deprecation of this in the C++11 standard. The result is a huge amount of warnings in the Linux build.

This change should be pretty much a no-op, because everything was declared as "= default", meaning the compiler will keep doing exactly what it was doing before just without a warning.

The reasoning for this change is the rule of three: https://en.cppreference.com/w/cpp/language/rule_of_three

> If a class requires a user-defined destructor, a user-defined copy constructor, or a user-defined copy assignment operator, it almost certainly requires all three.

So, I reviewed those classes, and so far my conclusion that a default copy constructions and operator= should work correctly, but I definitely would welcome second opinions here.

This shouldn't make anything worse or different in any way. The worst possible outcome should be to give the okay to an existing bug.

### DIFF
--- a/interface/src/raypick/StylusPick.h
+++ b/interface/src/raypick/StylusPick.h
@@ -31,6 +31,8 @@ public:
         surfaceNormal = stylusPickResult.surfaceNormal;
     }
 
+    StylusPickResult& operator=(const StylusPickResult &right) = default;
+
     IntersectionType type { NONE };
     bool intersects { false };
     QUuid objectID;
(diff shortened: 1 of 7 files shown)


daleglass:
Another commit with a couple I had missed.

Really, on a second thought, this looks like something that can be fixed by removing code instead, because most of these are happening because somebody wrote a copy constructor or operator= that does the exact same thing a default one would.

Also, what is up with this? https://github.com/kasenvr/project-athena/blob/kasen/core/libraries/image/src/image/Image.h#L71

        void operator=(const Image& other) {
            if (&other != this) {
                _packedData = other._packedData;
                _floatData = other._floatData;
                _dims = other._dims;
                _format = other._format;
            }
        }

There's no operator!= in the class, so all this check seems to do is to waste CPU time.

HifiExperiments:
that != is checking pointer equality as a quick way of avoiding a bunch of copies

generally though I think this is a fine solution, as would be deleting the unnecessary methods.  I have no preference, most of these classes aren’t touched very often

daleglass:
> that != is checking pointer equality as a quick way of avoiding a bunch of copies

Well, that's a bit embarrassing. Somehow I managed to read something else than what is written. After looking at it again it makes perfect sense.

PR #115, opened 2020-01-23
API Docs now default to a brand new dark theme.

kasenvr:
Easy on the eyes, for those late night metaverse scripting sessions. :)

image


ctrlaltdavid:
Looking at new dark mode, new light mode, and old theme...

1) Move the "toggle" button to under "Looking for Project Athena Documentation?"?

2) Dark mode "Search API docs ..." color makes it almost invisible.

3) Page headings/titles (e.g., "Project Athena API Reference") is very bright compared with context text. (It's #fffff and being a large and bolded font the brightness is accentuated.)

4) Light mode "Search API docs ..." is almost invisible. Input background should be white.

Looking at content...

5) "Deprecated" text in light rows of, say, methods list is hard to read / appears washed out.

6) "Deprecated" text in method documentation is hard to read / dark compared with other text and links.

New Dark Mode

New Light Mode

Old Theme

Controller API deprecation example

Deprecated text

kasenvr:
All corrections have been made, the style has been updated. :)

PR #139, opened 2020-01-30
Add an example of QT_CMAKE_PREFIX_PATH for 5.13.2

mikebdp2:
Despite doing a sudo make install of the latest stable QT version 5.13.2, the only location of the appropriate directory I was able to find, is $path_to_qt5_cloned_git/qt5/qtbase/lib/cmake


daleglass:
Hello, are you still interested in this one?

mikebdp2:
@daleglass Yes, I'm still interested, just unexperienced with Github and don't know how to separate a pull request. Feel free to extract a pthread commit and merge it separately, that will be much faster.

daleglass:
Closing this one.

I verified that the CMake change doesn't appear to be necessary, as it looks like recent versions pick up the required libraries without having to add them specifically.

For the QT5 build example I'll pass on it for now, since it may soon become obsolete. With the availability of Qt 5.15 in distributions we now should be able to build against the system Qt, and so some parts of the build process might need a bit of tweaking.

PR #145, opened 2020-02-01
feature: Allow users to set avatar thumbnails that show in avatar app.

vegaslon:
To use place a jpg next to avatar fst with same name as fst.  jpeg should show up as thumbnail inside avatarapp when fst is worn.  Such as if your avatar fst is "bob.fst" the thumbnail next to it would be "bob.jpg".

### DIFF
--- a/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
+++ b/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
@@ -30,21 +30,21 @@ ListModel {
     }
 	
     function trimFileExtension(url) {
-        var trimmedUrl = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
-        trimmedUrl = trimmedUrl.substring(0, (trimmedUrl.indexOf("?") == -1) ? trimmedUrl.length : trimmedUrl.indexOf("?"));
+        var trimmedUrl = url.substring(0, (url.indexOf("#") === -1) ? url.length : url.indexOf("#"));
+        trimmedUrl = trimmedUrl.substring(0, (trimmedUrl.indexOf("?") === -1) ? trimmedUrl.length : trimmedUrl.indexOf("?"));
         trimmedUrl = trimmedUrl.substring(0, trimmedUrl.lastIndexOf("."));
 
         return trimmedUrl;
     }
 	
-    function imageExists(image_url){
+    function imageExists(imageUrl) {
 
         var http = new XMLHttpRequest();
 
-        http.open('HEAD', image_url, false);
+        http.open('HEAD', imageUrl, false);
         http.send();
 
-        return http.status != 404;
+        return http.status !== 404;
 
     }
 
@@ -54,7 +54,7 @@ ListModel {
             return makeMarketThumbnailUrl(marketId);
         }
         
-        var avatarThumbnailFileUrl = trimFileExtension(avatarUrl)+".jpg";
+        var avatarThumbnailFileUrl = trimFileExtension(avatarUrl) + ".jpg";
         var thumbnailExist = imageExists(avatarThumbnailFileUrl);
         
         if (!thumbnailExist) {
--- a/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
+++ b/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
@@ -22,7 +22,7 @@ ListModel {
         return marketItemUrl;
     }
 	
-	function makeMarketThumbnailUrl(marketId) {
+    function makeMarketThumbnailUrl(marketId) {
         var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
             .split('%marketId%').join(marketId);
             
@@ -30,14 +30,14 @@ ListModel {
     }
 	
     function trimFileExtension(url) {
-        url = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
-        url = url.substring(0, (url.indexOf("?") == -1) ? url.length : url.indexOf("?"));
-        url = url.substring(0, url.lastIndexOf("."));
+        var trimmedUrl = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
+        trimmedUrl = trimmedUrl.substring(0, (trimmedUrl.indexOf("?") == -1) ? trimmedUrl.length : trimmedUrl.indexOf("?"));
+        trimmedUrl = trimmedUrl.substring(0, trimmedUrl.lastIndexOf("."));
 
-        return url;
+        return trimmedUrl;
     }
 	
-	function imageExists(image_url){
+    function imageExists(image_url){
 
         var http = new XMLHttpRequest();
 
@@ -55,9 +55,9 @@ ListModel {
         }
         
         var avatarThumbnailFileUrl = trimFileExtension(avatarUrl)+".jpg";
-        var exists= imageExists(avatarThumbnailFileUrl);
+        var thumbnailExist = imageExists(avatarThumbnailFileUrl);
         
-        if (!exists) {
+        if (!thumbnailExist) {
             return '';
         }
         
--- a/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
+++ b/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
@@ -21,17 +21,47 @@ ListModel {
 
         return marketItemUrl;
     }
+	
+	function makeMarketThumbnailUrl(marketId) {
+        var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
+            .split('%marketId%').join(marketId);
+            
+        return avatarThumbnailUrl;
+    }
+	
+    function trimFileExtension(url) {
+        url = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
+        url = url.substring(0, (url.indexOf("?") == -1) ? url.length : url.indexOf("?"));
+        url = url.substring(0, url.lastIndexOf("."));
+
+        return url;
+    }
+	
+	function imageExists(image_url){
+
+        var http = new XMLHttpRequest();
+
+        http.open('HEAD', image_url, false);
+        http.send();
+
+        return http.status != 404;
+
+    }
 
     function makeThumbnailUrl(avatarUrl) {
         var marketId = extractMarketId(avatarUrl);
-        if (marketId === '') {
+        if (marketId !== '') {
+            return makeMarketThumbnailUrl(marketId);
+        }
+        
+        var avatarThumbnailFileUrl = trimFileExtension(avatarUrl)+".jpg";
+        var exists= imageExists(avatarThumbnailFileUrl);
+        
+        if (!exists) {
             return '';
         }
-
-        var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
-            .split('%marketId%').join(marketId);
-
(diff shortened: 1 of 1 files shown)


interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -20,18 +20,48 @@ ListModel {
...
     }
 
     function makeThumbnailUrl(avatarUrl) {
         var marketId = extractMarketId(avatarUrl);
-        if (marketId === '') {
+        if (marketId !== '') {
+            return makeMarketThumbnailUrl(marketId);
+        }
+        
+        var avatarThumbnailFileUrl = trimFileExtension(avatarUrl)+".jpg";
```
HifiExperiments: it would be nice if this came from the fst (maybe if it’s not in there, it falls back to this)

this also might affect baking...I think with this solution, the thumbnails won’t be transferred over when you bake, so you’ll lose the thumbnail

vegaslon: Do we have a method to bake avatars in interface? if you are using the oven it still has files. it technically does not care what your avatar extension is it just cares that a jpg has the same name. this method will work even if you just wear avatars as a gltf or a fbx.

SilverfishVR: Related: https://github.com/kasenvr/project-athena/issues/84

HifiExperiments: sure but when you bake using the oven, it copies your files over to a new directory.  will this thumbnail image also get copied to the new location?

kasenvr:
Needs testing, but otherwise I think it's fine to merge for now since it's not a feature-loss but rather a feature-add. We can open an issue to outline that the baker does not know about it, so we can add that support later. :)

MarcusLlewellyn:
This works as advertised on Windows and Linux.
interface_VbmlRH3RNw
```
CMake 3.16.0


gcc/g++ 7.4.0
CMake 3.10.2
```

MarcusLlewellyn:
Unofficial PR build for windows here: [ProjectAthena-Alpha-PR145-e74fccb.exe](https://drive.google.com/file/d/1BocdDwBVQGCgpFRCvLA8HrP5p9nqS4is/view?usp=sharing)

Nex-Pro:
function trimFileExtension(url) {
        url = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
        url = url.substring(0, (url.indexOf("?") == -1) ? url.length : url.indexOf("?"));
        url = url.substring(0, url.lastIndexOf("."));

        return url;
    }
Isn't modifying the argument in a function considered to be an anti-pattern ?

perhaps use trimmedUrl

Nex-Pro:
in this block **thesame** is always true.
therefore the code in the if statement that checks **thesame** will always execute doesn't that make the  check pointless?

for (var i = 0; i < allAvatars.count; ++i) {
            var thesame = true;
            var bookmarkedAvatar = allAvatars.get(i);

            if (bookmarkedAvatar.avatarUrl !== avatar.avatarUrl) {
                continue;
            }

            if (bookmarkedAvatar.avatarScale !== avatar.avatarScale) {
                continue;
            }

            if (!modelsAreEqual(bookmarkedAvatar.wearables, avatar.wearables, compareWearables)) {
                continue;
            }

            if (thesame) {
                index = i;
                break;
            }
        }

Nex-Pro:
Magic 500's ?

```function compareNumericObjects(o1, o2) {
        if (o1 === undefined && o2 !== undefined) {
            return false;
        }
        if (o1 !== undefined && o2 === undefined) {
            return false;
        }

        for (var prop in o1) {
            if (o1.hasOwnProperty(prop) && o2.hasOwnProperty(prop)) {
                var v1 = o1[prop];
                var v2 = o2[prop];


                if (v1 !== v2 && Math.round(v1 * 500) != Math.round(v2 * 500)) {
                    return false;
                }
            }
        }```

        return true;
    }

**Comment by Nex-Pro** [comment]
Copyright header is missing

**Comment by Nex-Pro** [comment]
I should have made it clear that I meant  create a new variable.
var trimmedUrl = url;

then work on that.

File: interface/resources/qml/hifi/avatarapp/AvatarsModel.qml [hunk 2]
```diff
@@ -20,18 +20,48 @@ ListModel {
             .split('%marketId%').join(extractMarketId(avatarUrl));
 
         return marketItemUrl;
+    }
+	
+	function makeMarketThumbnailUrl(marketId) {
```
ctrlaltdavid: Indenting

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -20,18 +20,48 @@ ListModel {
...
+	
+    function trimFileExtension(url) {
+        url = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
+        url = url.substring(0, (url.indexOf("?") == -1) ? url.length : url.indexOf("?"));
+        url = url.substring(0, url.lastIndexOf("."));
+
+        return url;
+    }
+	
+	function imageExists(image_url){
```
ctrlaltdavid: Indenting

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -20,18 +20,48 @@ ListModel {
             .split('%marketId%').join(extractMarketId(avatarUrl));
 
         return marketItemUrl;
+    }
+	
+	function makeMarketThumbnailUrl(marketId) {
+        var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
+            .split('%marketId%').join(marketId);
```
daleglass: We shouldn't be using the name "hifi", and also is that an official hifi repo? We can expect those to go away eventually.

Also, is .split.join idiomatic JS?

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -21,17 +21,47 @@ ListModel {
...
+        trimmedUrl = trimmedUrl.substring(0, trimmedUrl.lastIndexOf("."));
+
+        return trimmedUrl;
+    }
+	
+    function imageExists(image_url){
+
+        var http = new XMLHttpRequest();
+
+        http.open('HEAD', image_url, false);
```
thoys: `HEAD` is not a valid HTTP method, I suspect you meant to put `GET`

daleglass: No, HEAD is perfectly legitimate. It's a GET without content, only headers. See for instance:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

thoys: Seems to save bandwidth, thanks for clarifying good to know this!

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -21,17 +21,47 @@ ListModel {
...
+	
+    function trimFileExtension(url) {
+        var trimmedUrl = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
+        trimmedUrl = trimmedUrl.substring(0, (trimmedUrl.indexOf("?") == -1) ? trimmedUrl.length : trimmedUrl.indexOf("?"));
+        trimmedUrl = trimmedUrl.substring(0, trimmedUrl.lastIndexOf("."));
+
+        return trimmedUrl;
+    }
+	
+    function imageExists(image_url){
```
thoys: image_url has inconsistent naming,  please use camelCase  e.g. `imageUrl`   ,   also space is missing between ) and {

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
thoys: space missing around the + operator `trimFileExtension(avatarUrl) + ".jpg";`

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -21,17 +21,47 @@ ListModel {
...
+
+        var http = new XMLHttpRequest();
+
+        http.open('HEAD', image_url, false);
+        http.send();
+
+        return http.status != 404;
```
thoys: please use `strict comparison`   change `!=` to `!==`

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
thoys: I think that this PR only works with the marketplace avatars from hifi at the moment. It probably won't work for long like this, but I think it's acceptable for now.

thoys: split and join is part of the JS language yes.

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -21,17 +21,47 @@ ListModel {
...
+    function makeMarketThumbnailUrl(marketId) {
+        var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
+            .split('%marketId%').join(marketId);
+            
+        return avatarThumbnailUrl;
+    }
+	
+    function trimFileExtension(url) {
+        var trimmedUrl = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
+        trimmedUrl = trimmedUrl.substring(0, (trimmedUrl.indexOf("?") == -1) ? trimmedUrl.length : trimmedUrl.indexOf("?"));
```
thoys: use strict comparisons change `==` to `===` which strictly checks it for the type as well.

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
```diff
@@ -21,17 +21,47 @@ ListModel {
...
+	
+    function makeMarketThumbnailUrl(marketId) {
+        var avatarThumbnailUrl = "https://hifi-metaverse.s3-us-west-1.amazonaws.com/marketplace/previews/%marketId%/large/hifi-mp-%marketId%.jpg"
+            .split('%marketId%').join(marketId);
+            
+        return avatarThumbnailUrl;
+    }
+	
+    function trimFileExtension(url) {
+        var trimmedUrl = url.substring(0, (url.indexOf("#") == -1) ? url.length : url.indexOf("#"));
```
thoys: use strict comparisons change `==` to `===` which strictly checks it for the type as well.

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
daleglass: I mean, is this a recognizable coding pattern in JS?

Like you could this in any other language, but you almost always would do it differently: use interpolation, a replace() call, sprintf, that kind of thing.

So the question is whether split and replace used for this particular purpose is a common feature of JS programs, or if most JS programmers would take a different approach here.

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
vegaslon: Yes before thumbnails only came from the marketplace. Just making sure do not break that functionality if you still have marketplace avatars bookmarked.

interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
vegaslon: There has been constant back and forth about the efficiency of replace vs split/join in JS over the past 10 years. In modern browsers they say that replace has been made more efficient so it might very well be the case with QT, when this file was originally made, that split/join was more efficient or still is. I was under the impression that that line of code had passed hifi's code review and that I should not fix what is not broken.

kasenvr:
Tested it out, works correctly. Honestly it's pretty darn awesome from what I see. Thanks for the PR! :)

Issue #151, opened 2020-02-03
Interface crashes on Linux on shutdown


daleglass:
Here's what I've been able to figure out so far. I turned on malloc debugging.
```
free(): invalid pointer

Thread 1 "interface" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	 return ret;
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007fffee1c88d9 in __GI_abort () at abort.c:79
#2  0x00007fffee2234af in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7fffee332f4b "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007fffee22aa9c in malloc_printerr (str=str@entry=0x7fffee3310e0 "free(): invalid pointer") at malloc.c:5339
#4  0x00007fffee22e982 in free_check (mem=0x5e42320, caller=<optimized out>) at hooks.c:255
#5  0x0000000003876984 in CPUIdent::CPUIdent_Internal::~CPUIdent_Internal() (this=0x5677f40 <CPUIdent::CPU_Rep>, __in_chrg=<optimized out>) at /home/dale/git/athena/project-athena/libraries/shared/src/CPUIdent.h:103
#6  0x00007fffee1e1e87 in __run_exit_handlers (status=0, listp=0x7fffee365578 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#7  0x00007fffee1e2040 in __GI_exit (status=<optimized out>) at exit.c:139
#8  0x00007fffee1ca1aa in __libc_start_main (main=0x2410958 <main(int, char const**)>, argc=1, argv=0x7fffffffdf38, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdf28) at ../csu/libc-start.c:342
#9  0x00000000020aa0be in _start ()
```

The problem happens in `CPUIdent::CPUIdent_Internal::~CPUIdent_Internal`, an implicit destructor. Disassembly shows it's destroying strings:
```

(gdb) disassemble
Dump of assembler code for function CPUIdent::CPUIdent_Internal::~CPUIdent_Internal():
[...]

PR #158, opened 2020-02-05
Force dynamic build on Linux

daleglass:
Fixes crashes on shutdown on Linux, see  #151 

Static libraries result in misbehavior on Linux. Here's my understanding of what happens:

1. You do a `static Foo foo;` in shared.
2. Logically, this creates a call to the `Foo` constructor.
3. Shared gets statically linked into both libfoo and libbar. Of course each of those now has the constructor call. They're static, so each has to include that code.
4. Both libfoo and libbar get statically linked into libplugin, which now gets two copies of the same code, and stuff gets  called twice.

This results in crashes on shutdown as destructors get called more than once, and this results in a double-free. This is sort of harmless, but there are possible grave issues:

1. Constructors also are called multiple times. This will overwrite the global data on events like a plugin load, leading to a loss of state. This can mean weird, hard to trace down bugs.
2. A bad shutdown can result in data loss from things not being given the chance to terminate correctly.
3. It will lead to a report of a crash to the user, which leaves a bad impression
4. It anything that relies on correct termination (like scripts or some debugging tools) malfunction.

Possible solutions were considered:

1. Change the code to avoid the crash. This was attempted, but seems to be a fair amount of work for little benefit. After fixing the initial crash a new, similar one appeared.
2. Ignore the problem and call `exit_`. This means a bad shutdown with possible data loss. Shutdown handlers exist for a reason.
3. Link dynamically. This works and is very easy to do.

### DIFF
--- a/cmake/init.cmake
+++ b/cmake/init.cmake
@@ -62,3 +62,11 @@ endif ()
 if (APPLE)
   set(CMAKE_XCODE_ATTRIBUTE_OTHER_CODE_SIGN_FLAGS "--deep")
 endif()
+
+if (UNIX)
+  # Static libs result in duplicated constructor and destructor calls on Linux
+  # and crashes on exit, and perhaps loss of global state on plugin loads.
+  #
+  # This will need to be looked at closely before Linux can have a static build.
+  set(BUILD_SHARED_LIBS ON)
+endif ()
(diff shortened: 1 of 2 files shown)


daleglass:
Testing:

1. Build
2. Verify that the different components are generated as .so (eg, there is a `libshared.so`, and not a `libshared.a`)
3. Verify that interface starts and appears to work correctly.
4. Verify that it no longer crashes on shutdown on Linux.

The changes only apply to Linux, and should have no effect on other systems.

Issue #122, opened 2020-01-25
Bug: scripts fail to load


SilverfishVR:
This one was invented for 0.86.0 I think (or was it 85? I reported the bug on oct. 4th). Bug also present in K2 (200122_Unofficial_Opus)
With a frequency of about 10-20%, when I start interface it fails to load most or all of my scripts, including default scripts, it has been verified by several others, including some poor souls trying hifi for the first time.
I did send log files at the time And I can do that again on request but I don’t really see anything in there that looks related.
This came about at the same time as the unwarranted stolen avatar messages and I’m pretty sure they are related, I don't really use marketplace avatars anymore but I would assume that this is also still happening.
There is a non-zero chance this may go away when we change metaverse API but at-least now it's on-record.

To reproduce: well, It happens a lot to me but some never see it? I usually load in my local sandbox but I don't think it makes a difference, maybe metaverse server ping related?

Basinsky:
I can confirm I have the same issue. If you dont have the developer mode on you even cannot push the load defaults button. Loading defaults generally fixes everything. But if you cannot find that, interface is unusable. I also think it has something to do with the unwarranted avatars.

ctrlaltdavid:
PR: https://github.com/kasenvr/project-athena/pull/162

PR #162, opened 2020-02-08
Fix scripts not running at Interface start after reloading all scripts

ctrlaltdavid:
Issue: https://github.com/kasenvr/project-athena/issues/122

### DIFF
--- a/libraries/script-engine/src/ScriptEngines.cpp
+++ b/libraries/script-engine/src/ScriptEngines.cpp
@@ -412,7 +412,7 @@ void ScriptEngines::stopAllScripts(bool restart) {
             continue;
         }
 
-        bool isOverrideScript = it.key().toString().compare(this->_defaultScriptsOverride.toString());
+        bool isOverrideScript = it.key().toString().compare(this->_defaultScriptsOverride.toString()) == 0;
         // queue user scripts if restarting
         if (restart && (scriptEngine->isUserLoaded() || isOverrideScript)) {
             _isReloading = true;


ctrlaltdavid:
#### Test Plan

1. Repro the problem.
- A previous version of Interface.
- Running Scripts dialog, press the "Reload All" button.
- Restart Interface.
There will be no scripts running.

2. Check that problem is fixed.
- Repeat the tests in 1 with the PR version of Interface. 
The scripts should stay running each Interface run.

MarcusLlewellyn:
Unofficial PR build for windows here: [ProjectAthena-Alpha-PR162-5a91de0.exe](https://drive.google.com/file/d/1dWopYhZOaFjQtCb5kdMp2hPX3GCFdFKM/view?usp=sharing)

libraries/script-engine/src/ScriptEngines.cpp
```diff
@@ -412,7 +412,7 @@ void ScriptEngines::stopAllScripts(bool restart) {
             continue;
         }
 
-        bool isOverrideScript = it.key().toString().compare(this->_defaultScriptsOverride.toString());
+        bool isOverrideScript = it.key().toString().compare(this->_defaultScriptsOverride.toString()) == 0;
```
HifiExperiments: god this is confusing

SilverfishVR:
Switching back an forth between K2 RC1 and the PR build over the last few days, It definitely seems fixed. Thank You, this was really starting to annoy me.

Q: Who closes the issue (#122) , the author or the fixer?

kasenvr:
Either way is fine, I would wait to close an issue only once the fix is merged though.

PR #170, opened 2020-02-17
Fix chat server relay

kasenvr:
Created a chat relay on the metaverse server, grid chat now works again.


ctrlaltdavid:
Code review and tested OK.

PR #179, opened 2020-02-25
Makes simplified nametag a default script

kasenvr:
Also made it so that it disables while using the inspect.js script.

Test plan:

When using the inspect.js script (holding the alt key), click on an avatar multiple times as you zoom in, out, and around that person. Ensure the nametag does not show up. Now stop using the inspect script, click on them again and observe that the nametags do show, timeout, or disappear when clicked again.

### DIFF
--- a/scripts/simplifiedUI/ui/simplifiedNametag/simplifiedNametag.js
+++ b/scripts/simplifiedUI/ui/simplifiedNametag/simplifiedNametag.js
@@ -8,10 +8,11 @@
 //
 //    Click on someone to get a nametag for them
 //    
-var PickRayController = Script.require('./resources/modules/pickRayController.js');
+var PickRayController = Script.require('./resources/modules/pickRayController.js?' + Date.now());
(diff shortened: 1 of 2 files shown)


scripts/simplifiedUI/ui/simplifiedNametag/simplifiedNametag.js
```diff
@@ -88,5 +106,6 @@ function handleAvatarNametagMode(newAvatarNameTagMode) {
 
 var nameTag = startup();
 
-module.exports = nameTag;
-
+// if (module) {
+//     module.exports = nameTag;
+// }
```
ctrlaltdavid: Better would be...
```
try {
    module.exports = nameTag;
} catch (e) {
    // module doesn't exist when script run outside of simplified UI.
}
```

odysseus654:
Seems to work okay.  Using the alt-click script, no reaction when pivoting around an avatar, can click avatar afterwards and see the nametag pop up for a few seconds.

PR #184, opened 2020-02-26
Updates the var name and URL for the goto JSON for use in chat

kasenvr:


kasenvr:
Ah, that's because it very much is pre-production. It looks like so far we are using the decen. go-to as an interim app until we deploy the live metaverse.

MarcusLlewellyn:
Unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR184-4455fae.exe](https://drive.google.com/file/d/1TephOuib_WDZ6jGskgfR2BhjllCR99OG/view?usp=sharing)

kasenvr:
Testing the PR, it works for me. The Hub is failing because its IP does not appear to be valid, this may be a problem with the GoTo server itself or with "The Hub"s server, but this is not an issue with the chat app's implementation as it also fails to navigate there on the "EXPLORE" app itself.

PR #215, opened 2020-03-17
Feature/shape alpha

AleziaKurdis:
This modification adds the missing "**alpha**" property to the "**Shape**" entity editor _(in "Create" app)_


MarcusLlewellyn:
Unofficial PR build for Windows is available here: [ProjectAthena-Alpha-PR215-1b3744a.exe](https://drive.google.com/file/d/1roq2c3SwepekYzSvuXcdqjEaMt1euzbz/view?usp=sharing)

Aitolda:
> Unofficial PR build for Windows is available here: [ProjectAthena-Alpha-PR215-1b3744a.exe](https://drive.google.com/file/d/1roq2c3SwepekYzSvuXcdqjEaMt1euzbz/view?usp=sharing)

Permissions needed to download.

AleziaKurdis:
Just entered this issue:
**#221 textEffectThickness property: outlineFill is not working.** 
Something is wrong on C++ side.

Nex-Pro:
@AleziaKurdis 

**createAppTooltips.json** line 39: 
```"tooltip": "The magnitude of the text effect."```

is **magnitude** a good word choice ? would **thickness** be better ?

Nex-Pro:
Same file:

"imageAlpha": {
        "tooltip": "The alpha of the image."
    },

maybe

**The alpha (opacity) of the image**

Most people understand opacity. Alpha is a technical term

AleziaKurdis:
> @AleziaKurdis
> 
> **createAppTooltips.json** line 39:
> `"tooltip": "The magnitude of the text effect."`
> 
> is **magnitude** a good word choice ? would **thickness** be better ?

Wasn't my preferred word either. But I stuck with what it is used in the documentation.
I think they avoided using "thickness" because this same value is also used for the offset of the "Shadow" effect.  I deduced that they used "magnitude" in the meaning of "amount of effect to applied".
That said, I used "Effect Thickness" as label in the editor. I think this was probably a wrong move.

I'm open to suggestions.
What should we used for the label, the tooltip, and the JSdoc?

AleziaKurdis:
> Same file:
> 
> "imageAlpha": {
> "tooltip": "The alpha of the image."
> },
> 
> maybe
> 
> **The alpha (opacity) of the image**
> 
> Most people understand opacity. Alpha is a technical term

All the tooltips talked currently about "alpha". (there are a bunch)
Do you want I do the adjustment for all those tootips ?
If so, should I use:
"**alpha (opacity)**"
or just "**opacity**" (assuming that it's the tooltip of an "Alpha" labeled property)
Should we adjust the JSDoc with "opacity" in the description?

MarcusLlewellyn:
Updated unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR215-1f72e99.exe](https://drive.google.com/file/d/1bzWlzSd78uBYepTC6DAFi682S409lBqa/view?usp=sharing)

Nex-Pro:
I think it is something we should have a conversation about with the others.

If a user knows what alpha is they do not need a tooltip. The tooltip is there to explain what it is not to repeat what the label that already exists says.

Nex-Pro:
@MarcusLlewellyn Chip in on the above please.

Nex-Pro:
I'd probably want the alpha tooltip to read: 

**Set the opacity of the entity between 0.0 and 1.0 with zero being fully transparent and one being completely opaque.**  

assuming that is the right way round.

AleziaKurdis:
I agree with this, the tooltip must be descriptive and educate in popularized words.
There are many that would need to be revisited.

Even if I can certainly detect those that would need to be reworked, I'm probably not the best person to rewrite them. (English is not my base language)
Is anyone interested to do that?

For the Labels, here's the rules I followed:
- Be as closed as possible of the technical name (property name)
- But in human wording
- Keeping in mind that there is a limit in length to fit in the small UI

MarcusLlewellyn:
> I'd probably want the alpha tooltip to read:
> 
> **Set the opacity of the entity between 0.0 fully transparent and 1.0 completely opaque.**
> 
> assuming that is the right way round.

This is good, although as Alezia points out perhaps a little verbose. An alternate might be:

**The transparency of the entity from 0.0 to 1.0.**

Most English speakers will immediately understand "transparency" I think. "Opacity" is not _quite_ as recognizable. "Alpha" is definitely a technical word, making the tool-tip necessary in the first place. That the default is 1.0 is a pretty big hint that it represents complete opacity, and simply playing with the setting will make it very apparent.

The only exception here I can think of is when a procedural material entity is applied, in which case any value less than 1.0 allows the shader to set the opacity. But that's better explained in the shader docs, and is an advanced topic.

AleziaKurdis:
The tooltip (which is actually more a description callout) can take long text without problem.
I would prefer that we go move verbose but without ambiguity.
"The transparency of the entity from 0.0 to 1.0." can be read like this: if transparency = 0 (absence of transparency) then it would be opaque, which is the opposite of what it is.
Give the info about the signification of the min/max is definitely the info that the users need. More than the min/max it-self that is already defined by the UI.

Here's what I suggest to do:
In a first time, I will extract all the tooltips that are:
- empty of signification. 
- containing ambiguities.
- doesn't contain the key informations
In a second time, we will figure out how to address them in a productive way according the quantity we will have.

MarcusLlewellyn:
> The tooltip (which is actually more a description callout) can take long text without problem.
> I would prefer that we go move verbose but without ambiguity.

If tool-tip length is not an issue, then I agree that more descriptive is better, and that NexPro's suggested phrasing is a good one.

AleziaKurdis:
I create this excel sheet:
[https://github.com/AleziaKurdis/athena_editor_tooltips/blob/master/TOOLTIPS_REVISION.xlsx](https://github.com/AleziaKurdis/athena_editor_tooltips/blob/master/TOOLTIPS_REVISION.xlsx)

I have so far reviewed about 50% of the tooltips. ( I will complete soon)

What I plan to do **for this current PR**, is to fix the tooltips of the **11 properties** that I marked to the status "**FIXED**" in the excel sheet. (Mainly those related to Alpha)
**Please confirm they are ok.**

**For all the other,** I will open a separate issue to have those tooltips rewritten correctly.

kasenvr:
Awesome, yes let's keep this in the scope of the PR! Let's make this PR good, but fixing other aspects (such as other tooltips) should be separate. So that's not something that you should tackle in this PR @AleziaKurdis and if you want to, you can tackle the others later later but you don't have to! :)

MarcusLlewellyn:
Updated unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR215-90ef2ee.exe](https://drive.google.com/file/d/1EEKiiQRg_-3_sAGoElD9Njh23AzWfMeq/view?usp=sharing)

PR #222, opened 2020-03-19
Fix updated TBB library hash in vcpkg portfile

MarcusLlewellyn:
For some reason, the directory that vcpkg puts the TBB library into has changed. As a result, a different SHA256 hash is needed to keep vcpkg from dying with an error.

Testing: Not much to test here. It should start and run reliably.


MarcusLlewellyn:
Unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR222-acbc8a8.exe](https://drive.google.com/file/d/1Z7k7T4ubFiw9jJso4Lo2apyBc-XEdWHU/view?usp=sharing)

ctrlaltdavid:
Also downloaded and ran the unofficial PR build: appears to work OK.

ctrlaltdavid:
@daleglass built and ran on Linux, OK.

PR #230, opened 2020-03-22
Very early and incomplete support for translating the menus

daleglass:
This is mostly for the benefit of languagelovr and anyone else interested in working on a translation. 

It's incomplete and won't work due to bugs in the interface, but it provides an example of how to make something translatable, shouldn't hurt, and works for testing the translation toolset. It compiles, and seems to do no harm, so it can be merged, it just won't do much good yet.

Instructions for how to use this:

First, run the tool that extracts strings from the code:

    /usr/lib64/qt5/bin/lupdate  interface/src/Menu.h -ts interface.ts

Then, translate it with Qt Linguist:

    linguist-qt5 interface.ts

I think you need to check the checkbox next to each translation for that translation to be deemed as completed and usable.

Use File/Release to generate an `interface.qm` file. This then can be placed in `i18n/interface_en.qm`, in the build tree.The `i18n` folder needs to be created. It's "interface_en", because that name is hardcoded into interface, and there's no code to actually select the right language. This would need to be fixed.

At that point, this *should* work. However, it doesn't. Placing breakpoints in:

    interface/src/main.cpp:443
    interface/src/Menu.cpp:86

one can find that `Menu.cpp` is hit first, before the translation file is loaded in `main.cpp`. This seems to be some sort of bug in how the interface is coded.

This PR also doesn't cover the entire menu system, part is in the code. Also, QML references menus by their English text, which is likely to provide some additional complications.

And, `lupdate` and `lrelease` probably need to be added to cmake.


daleglass:
Relevant Qt docs:

https://doc.qt.io/qt-5/linguist-manager.html
https://doc.qt.io/qt-5/i18n-source-translation.html
https://doc.qt.io/qt-5/qtlinguist-hellotr-example.html

daleglass:
Okay, I got the very basics of translation working. Changes in the last commit:

First of all, now translations are actually being applied correctly, and it will try to use a country-specific version based on the system locale. On Linux, the locale can be forced with a command like this (see the discussion of language codes below):

    $ set LC_ALL=es_ES.UTF-8

Interface will tell you at the top of the log what translation it's trying to use, and if it worked:

    [03/22 19:18:49] [INFO] [hifi.interface] Loaded translation for locale  QLocale(Spanish, Latin, Spain)

or not:

    [03/22 19:08:17] [WARNING] [hifi.interface] Failed to load translation for locale  QLocale(Spanish, Latin, Spain)

Unfortunately since Qt does its magic internally, it doesn't tell you the actual filename it's using, so here is what it tries for reference, in order, from top to bottom:

    i18n/interface_es_ES.qm
    i18n/interface_es_es.qm
    i18n/interface_es.qm
    i18n/interface.qm

This is currently relative to the build folder. That will probably change in the future.

Language codes have two parts: language, and country. So, "es_ES" is Spanish, as spoken in Spain. "es_AR" would be Spanish, as spoken in Argentina, "es_MX" is Spanish, in Mexico. And so on. The all lowercase variations are to deal with possible issues on case sensitive systems like Linux.

English is "en_US" for the USA, and "en_GB" for the UK.

If a country-specific variation isn't found it'll try for a general one, so just "es" or "en".

Handy list: http://quivi.sourceforge.net/languagecodes.html


**Warning:** Even though this should work correctly, actually translating the interface is likely to break things in some cases, due to the heavy use of strings that are now localized to refer to menus, rather than having some sort of unchanging identifier. This will need to be addressed in the future. Some things should be localizable harmlessly, but some will need fixing.

MarcusLlewellyn:
Unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR230-01fd781.exe](https://drive.google.com/file/d/1TY9Xl6-F_QYQAomgleBBW6ya24AT-G27/view?usp=sharing)

daleglass:
Turns out to break scripting, and needs much more work. Will revisit this later.

PR #253, opened 2020-04-02
Rebrand server console to Vircadia with Kasen icon.

MarcusLlewellyn:


MarcusLlewellyn:
Here is an unnoficial Windows build: [ProjectAthena-Alpha-K2-console_test.exe](https://drive.google.com/file/d/17Wdm5Td7IIbKM_sDG2ElEMGoAOQBoTSc/view?usp=sharing)
This is __not__ a PR build. This is a release build. That's required to test the server console properly.

Also suggest uninstalling any previous builds. It is safe to keep you settings in place.

kasenvr:
I missed one PR build of v83 of HiFi but it doesn't seem to have affected anything when I didn't uninstall it.

I get my Project Athena shortcut in the start menu, and I get Sandbox, the Sandbox still shows the wrong icon though for me (ONLY IN THE START MENU)... when I view the actual folder for the shortcuts, it looks fine, it also looks fine in the actual tray (icon wise.)

MarcusLlewellyn:
> I missed one PR build of v83 of HiFi but it doesn't seem to have affected anything when I didn't uninstall it.
> 
> I get my Project Athena shortcut in the start menu, and I get Sandbox, the Sandbox still shows the wrong icon though for me (ONLY IN THE START MENU)... when I view the actual folder for the shortcuts, it looks fine, it also looks fine in the actual tray (icon wise.)

MarcusLlewellyn:
> I missed one PR build of v83 of HiFi but it doesn't seem to have affected anything when I didn't uninstall it.
> 
> I get my Project Athena shortcut in the start menu, and I get Sandbox, the Sandbox still shows the wrong icon though for me (ONLY IN THE START MENU)... when I view the actual folder for the shortcuts, it looks fine, it also looks fine in the actual tray (icon wise.)

It seems likely that Windows may be caching the icon here.

kasenvr:
After rebooting post-uninstall, it appears there's only one straggler icon which I'm sure must be related to the icon cache so otherwise, it looks fine to me.

Aitolda:
Mostly working, but the same issue as Kalila with regarding windows search and the old sandbox. However I still have the original High Fidelity sandbox so it's not surprising that it finds it, but that it ONLY shows the original High Fidelity sandbox, and not both (when using windows search).  Otherwise all other usage of the icon is as expected, including server-console.

PR #257, opened 2020-04-03
Separate user activity logging from crash logging

odysseus654:
separate user activity logging from crash logging, allowing people to say "no" to tracking while still sending crashes

Adds new setting CrashLoggerDisabled along with a new corresponding user preference.

### DIFF
--- a/interface/src/main.cpp
+++ b/interface/src/main.cpp
@@ -218,11 +218,11 @@ int main(int argc, const char* argv[]) {
     }
     qDebug() << "UserActivityLogger is enabled:" << ual.isEnabled();
 
-    if (ual.isCrashEnabled()) {
+    qDebug() << "Crash handler logger is enabled:" << ual.isCrashMonitorEnabled();
+    if (ual.isCrashMonitorEnabled()) {
         auto crashHandlerStarted = startCrashHandler(argv[0]);
         qDebug() << "Crash handler started:" << crashHandlerStarted;
     }
-
 
     const QString& applicationName = getInterfaceSharedMemoryName();
     bool instanceMightBeRunning = true;
--- a/interface/src/main.cpp
+++ b/interface/src/main.cpp
@@ -218,7 +218,7 @@ int main(int argc, const char* argv[]) {
     }
     qDebug() << "UserActivityLogger is enabled:" << ual.isEnabled();
 
-    if (ual.isEnabled()) {
+    if (ual.isCrashEnabled()) {
         auto crashHandlerStarted = startCrashHandler(argv[0]);
         qDebug() << "Crash handler started:" << crashHandlerStarted;
     }
--- a/interface/src/ui/PreferencesDialog.cpp
+++ b/interface/src/ui/PreferencesDialog.cpp
@@ -254,7 +254,15 @@ void setupPreferences() {
         auto setter = [](bool value) { Menu::getInstance()->setIsOptionChecked(MenuOption::DisableActivityLogger, !value); };
         preferences->addPreference(new CheckPreference("Privacy", "Send data - High Fidelity uses information provided by your "
                                 "client to improve the product through the logging of errors, tracking of usage patterns, "
-                                "installation and system details, and crash events. By allowing High Fidelity to collect "
+                                "installation and system details. By allowing High Fidelity to collect this information "
+                                "you are helping to improve the product. ", getter, setter));
+    }
+
+    {
+        auto getter = []()->bool { return !Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
+        auto setter = [](bool value) { Menu::getInstance()->setIsOptionChecked(MenuOption::DisableCrashLogger, !value); };
+        preferences->addPreference(new CheckPreference("Privacy", "Send crashes - Vircadia uses information provided by your "
+                                "client to improve the product through crash events. By allowing Vircadia to collect "
                                 "this information you are helping to improve the product. ", getter, setter));
     }
 
--- a/interface/src/ui/PreferencesDialog.cpp
+++ b/interface/src/ui/PreferencesDialog.cpp
@@ -259,7 +259,7 @@ void setupPreferences() {
     }
 
     {
-        auto getter = []()->bool { return !Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
+        auto getter = []()->bool { return Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
         auto setter = [](bool value) { Menu::getInstance()->setIsOptionChecked(MenuOption::DisableCrashLogger, !value); };
         preferences->addPreference(new CheckPreference("Privacy", "Send crashes - Vircadia uses information provided by your "
                                 "client to improve the product through crash reports. By allowing Vircadia to collect "
--- a/interface/src/ui/PreferencesDialog.cpp
+++ b/interface/src/ui/PreferencesDialog.cpp
@@ -259,7 +259,7 @@ void setupPreferences() {
     }
 
     {
-        auto getter = []()->bool { return Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
+        auto getter = []()->bool { return !Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
         auto setter = [](bool value) { Menu::getInstance()->setIsOptionChecked(MenuOption::DisableCrashLogger, !value); };
         preferences->addPreference(new CheckPreference("Privacy", "Send crashes - Vircadia uses information provided by your "
                                 "client to improve the product through crash reports. By allowing Vircadia to collect "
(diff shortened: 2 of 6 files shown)


MarcusLlewellyn:
This code will no longer compile until it is updated for the latest vcpkg dependencies.

odysseus654:
...do I need to rebase it?

MarcusLlewellyn:
Unofficial PR build for Windows available here: [ProjectAthena-Alpha-PR257-7cea4fc.exe](https://cdn.vircadia.com/dist/pr-builds/ProjectAthena-Alpha-PR257-7cea4fc.exe)

odysseus654:
I'm expecting the configuration message to need to be changed at the minimum

odysseus654:
Also I would like this applied to the K2 branch as well

kasenvr:
> 
> 
> Also I would like this applied to the K2 branch as well

It will get applied to K2.1's branch.

kasenvr:
Never been so happy to crash myself, it works as expected except one thing. I had to check the "Disable Crash Logger" box, then uncheck it, restart Interface to get it to work. I suspect this is because a setting did not exist, and the default setting for "True" did not kick in.

libraries/networking/src/UserActivityLogger.h
```diff
@@ -55,6 +59,7 @@ private slots:
 private:
     UserActivityLogger();
     Setting::Handle<bool> _disabled { "UserActivityLoggerDisabled", true };
```
odysseus654, line 61: Should we change this line as well while we're at it?

```suggestion
    Setting::Handle<bool> _disabled { "UserActivityLoggerDisabled", false };
```

odysseus654:
Thank you ctrlaltdavid, suggested changes applied.

libraries/networking/src/UserActivityLogger.h
kasenvr: That would effectively enable the activity logger by default, I'd say hold off on that for now.

odysseus654: So we'll leave it as "apparently enabled" but not really working unless you uncheck and check it?

...I think I'm okay with that actually.

kasenvr: Our wording gets confusing really fast since the variable is for a "disabled" state, so when we say enabled or disabled... I get confused about which we're referring to lol.

odysseus654: It would not be accurate to not say that you are not unconfused.

interface/src/ui/PreferencesDialog.cpp
```diff
@@ -254,7 +254,15 @@ void setupPreferences() {
         auto setter = [](bool value) { Menu::getInstance()->setIsOptionChecked(MenuOption::DisableActivityLogger, !value); };
         preferences->addPreference(new CheckPreference("Privacy", "Send data - High Fidelity uses information provided by your "
                                 "client to improve the product through the logging of errors, tracking of usage patterns, "
-                                "installation and system details, and crash events. By allowing High Fidelity to collect "
+                                "installation and system details. By allowing High Fidelity to collect this information "
+                                "you are helping to improve the product. ", getter, setter));
+    }
+
+    {
+        auto getter = []()->bool { return !Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
```
odysseus654, line 262:
```suggestion
        auto getter = []()->bool { return Menu::getInstance()->isOptionChecked(MenuOption::DisableCrashLogger); };
```
Looks like one too many "nots" on this as well

MarcusLlewellyn:

MarcusLlewellyn:
Updated unofficial  PR build for Windows available here: [ProjectAthena-Alpha-PR257-aa5413f.exe](https://cdn.vircadia.com/dist/pr-builds/ProjectAthena-Alpha-PR257-aa5413f.exe)

odysseus654:
if i'm correct, the only remaining issue is ensuring that the checkbox in the Preferences dialog box is set properly for whether the crash dump is enabled or not?

FYI the crash dump setting is examined **only** on program launch, one can tell if it is running by looking in Process Explorer and seeing the crash dump handler as a child process; if crash dumping is enabled it will be there, if it's disabled it won't be.

SilverfishVR:
### testing : ProjectAthena-Alpha-PR257-aa5413f.exe

On first run, "send data" is enabled and "send crash" is disabled, crashpad_handler.exe is running.

- "send data" checkbox state is persistent and saves to interface.json.

- "send crash" checkbox is not, in fact, the logic seems reversed, if I manually set:
`"Network/Disable Crash Logger": true,` in interface.json the checkbox appears checked

- simply en/disabling crash logging in interface.json works, but I have to start, and then restart interface for it to take effect (the value in the json does not change)

- **TL:DR** the logic is messed up

kasenvr:
> 
> ### testing : ProjectAthena-Alpha-PR257-aa5413f.exe
> 
> On first run, "send data" is enabled and "send crash" is disabled, crashpad_handler.exe is running.
> 
>     * "send data" checkbox state is persistent and saves to interface.json.
> 
>     * "send crash" checkbox is not, in fact, the logic seems reversed, if I manually set:
>       `"Network/Disable Crash Logger": true,` in interface.json the checkbox appears checked
> 
>     * simply en/disabling crash logging in interface.json works, but I have to start, and then restart interface for it to take effect (the value in the json does not change)
> 
>     * **TL:DR** the logic is messed up

Should it not be the case that "Disable Crash Logging" *should* be unchecked *with* crashpad also running? We want it to be running by default!

Settings in Interface.JSON update weirdly, at some point we probably want to take a look at that phantom system but it appears to generally work for setting and retrieving values through the API during runtime.

odysseus654:
Considering the last change I made to the code reverses the visible state of that checkbox, I'm guessing I'm going to need to look much more closely at this code.

Both checkboxes should be the opposite of what is stored in the file.
The "send data" checkbox should default to unchecked.
The "send crashes" checkbox should default to checked.

I can take another look tonight to see what might be messed up.

This patch isn't big enough for things to be this complicated...

odysseus654:
Okay this is starting to get a bit frustrating, the patch I pushed from the previous code review has been reverted -- it looks like the previous version was more consistent.

I likely need to spend more focus on this.  Is the previous build still available, or do we need to bother Marcus in remaking it?

libraries/networking/src/UserActivityLogger.h
```diff
@@ -35,7 +35,11 @@ public slots:
     bool isEnabled() { return !_disabled.get(); }
     bool isDisabledSettingSet() const { return _disabled.isSet(); }
 
+    bool isCrashEnabled() { return !_crashDisabled.get(); }
+    bool isCrashDisabledSettingSet() const { return _crashDisabled.isSet(); }
+
     void disable(bool disable);
+    void crashDisable(bool disable);
```
thoys, line 42: perhaps rename `crashDisable` to `disableCrashReporting` . 

- `crashDisable` sounds like you're crashing a disable which doesn't make sense

interface/src/main.cpp
```diff
@@ -218,7 +218,7 @@ int main(int argc, const char* argv[]) {
     }
     qDebug() << "UserActivityLogger is enabled:" << ual.isEnabled();
 
-    if (ual.isEnabled()) {
+    if (ual.isCrashEnabled()) {
```
thoys, line 221: perhaps rename to `isCrashReportingEnabled()`    , crashes can always happen even if not reported

libraries/networking/src/UserActivityLogger.h
```diff
@@ -55,6 +59,7 @@ private slots:
 private:
     UserActivityLogger();
     Setting::Handle<bool> _disabled { "UserActivityLoggerDisabled", true };
+    Setting::Handle<bool> _crashDisabled { "CrashLoggerDisabled", false };
```
thoys, line 62: can this be renamed to _crashReporterEnabled ?  ,  and option CrashReporterEnabled

thoys: or Reporting

odysseus654:
Okay after a couple tries I managed to get my own PR build running (although I apparently wasn't able to activate Crashpad support for the build).  These are the results of my own tests (based on the current merge tip, which is equivalent to what was first released on the first PR build that Marcus made:

When starting up with an empty folder the crash monitor flag is: **Enabled**
...and the user activity logger is: **Enabled** _(according to output console)_
...and the child process is: **Missing** _(with launch code executed)_
When going into preferences the checkbox is: **Checked**
When I uncheck the checkbox the setting is set to: **true**
When restarting the interface the crash monitor flag is: **Disabled**
...and the child process is: **Missing** _(with launch code skipped)_
When going into the preferences the checkbox is: **Unchecked**
When I uncheck the checkbox the setting is set to: **false**
When restarting the interface the crash monitor flag is: **Enabled**
...and the child process is: **Missing** _(with launch code executed)_

Based on this I really don't understand the objections that were found with the first build that caused me to knee-jerk a fix that wasn't appropriate (maybe there were other fixes involved).

I will review Thoy's recent change requests (along with some additional debug output tweaks from me) before formally requesting another PR build so this can get tested and cleared.

ctrlaltdavid:
A couple of possibilities if there are still problems here:
- Rename the Developer menu items from "disable" to "enable", to match the settings dialog.
- Remove the Developer menu items given that their function is surfaced in the settings dialog.

Also, note that menu items automatically have their own saving and restoring of state, and this might perhaps be interfering with the explicit setting and restoring of state for the dialog items.

kasenvr:
Alright, David and I agree it should be merged as-is within the scope of the PR. We should get rid of these settings from the developer menu and rely on their application in the interface later. Also the High Fidelity references should be removed in a further housekeeping PR.

Also at some point maybe it might be nice to have the toggle work without a restart.

PR #273, opened 2020-04-04
package scripts: renaming athena -> vircadia

odysseus654:
change all references for "athena" to "vircadia" in the package installation script:

package name changed
service names changed from athena -> vircadia
username the service runs as changed
location of /etc/opt/athena -> /etc/opt/vircadia
location of /opt/athena -> /opt/vircadia
location of /var/lib/athena -> /var/lib/vircadia


daleglass:
Discussion during dev meeting: try to provide a migration script

odysseus654:
The question in my mind is kinda what "migration" means in this context, especially as the new package name indicates this is a different installation that can be installed in parallel.  What's initially come to my mind to do on first time installation's is:
- Copy (but do not move) files from any Athena installs
- Try to figure out which services were enabled for Athena, enable equivalent ones for Vircadia, and disable all Athena services
- The same for running services - scan for running Athena services, shut them down and start equivalent ones as Vircadia.

The fact that we have the installation package for Athena here does make things easier in terms of knowing where things are, rather than say the HiFi installations which I haven't yet found (or were found to be Docker installs)

daleglass:
I think the main concern was an easy migration for the less technically literate users -- the idea is that unless you're doing something strange, the following should work:

1. Install old packages
2. Create a domain
3. Install new vircadia packages
4. Domain continues to work

On the other hand, perhaps we should look into how many people would be affected. If we're talking about just a few maybe it's easier to just handle it by hand for anyone who has any trouble.

odysseus654:
Considering how many random comments have been in the Discord channel for "thanks for the install packages", I suspect we're a couple months late for just doing "let's break it and see if we can fix it afterwards".

I may be able to do a "Conflicts:" line to force the package rename on this too which may make it an effective upgrade.

This is going to be at least a few hours research, I can see about pushing forward on this this weekend.

odysseus654:
I've spent a couple hours tonight trying to upgrade the scripts.  Note that these have NOT been even smoketested yet so the likelihood that they even run without syntax errors is questionable at best.

Significant intents here:
- on installation when no server data exists, look for an athena installation.  If it finds one, note the enabled&active services, shut down all active services, copy the /etc/opt/athena and /var/lib/athena folders over, start up and enable the equivalent services
- on purge, remove all files from /etc/opt/vircadia and /var/lib/vircadia (purge all server data)

I haven't made movements towards auto-deleting the athena files yet, I'm uncertain that we shouldn't permit parallel (but inactive) installations, especially where a protocol bump is involved.  (note that there is no "purge" script for the athena .deb files so they won't be deleting their copy of the files)

kasenvr:
Yeah, makes sense that a deletion should not take place unless prompted... Is there a way to prompt / ask before proceeding with that? It might make life a lot easier since far less automation is needed:

"Would you like to migrate your Project Athena installation? If yes, then we will copy over the relevant files, turn off the services and start new services. You will have to delete any old files that you do not wish to keep." for example is a good start. Can be improved on at a later time so at least for now you we can run with what we have. :)

odysseus654:
This is the version that's been included in the most recent RC2 release.  I've "checkpointed" and created a new branch for further work (which I will switch this to if it hasn't been merged in by that point.

Further work that needs to be done (by me):
- Get all the scripts previously tested for Ubuntu working on AWS Linux
- Get the "uninstall" action to stop the service

odysseus654:
okay, think I've exhausted everything to test on both AWS Linux and Ubuntu, considering this complete

PR #279, opened 2020-04-07
Fix VS2019 (Version 16.5.1) builds

thoys:
Building in the latest version of Visual studio 2019 fails with a `'operator =' is ambigious` error. This small patch will fix the error.

PR #281, opened 2020-04-07
Improve performance of MyAvatar::simulate when you have lots of descendants

HifiExperiments:
Run [this script](https://gist.githubusercontent.com/HifiExperiments/8f79e199b1f132f06e512a18f91c3b53/raw/3ea3bfcee0fdd5849eba8f53394eafd510f6724f/ModelParenting.js).  You'll see a bunch of mannequins in front of you.  Press 'o'.  This will make the bottom model a child of your avatar.  On master, your framerate will basically go to 0.  Now, your framerate will still drop, but not nearly as much, and interface should still be usable.

### DIFF
--- a/interface/src/avatar/MyAvatar.cpp
+++ b/interface/src/avatar/MyAvatar.cpp
@@ -972,7 +972,7 @@ void MyAvatar::simulate(float deltaTime, bool inView) {
         recorder->recordFrame(FRAME_TYPE, toFrame(*this));
     }
 
-    locationChanged(true, false);
+    locationChanged(true, true);
     // if a entity-child of this avatar has moved outside of its queryAACube, update the cube and tell the entity server.
     auto entityTreeRenderer = qApp->getEntities();
     EntityTreePointer entityTree = entityTreeRenderer ? entityTreeRenderer->getTree() : nullptr;
@@ -981,16 +981,7 @@ void MyAvatar::simulate(float deltaTime, bool inView) {
         entityTree->withWriteLock([&] {
             zoneInteractionProperties = entityTreeRenderer->getZoneInteractionProperties();
             EntityEditPacketSender* packetSender = qApp->getEntityEditPacketSender();
-            forEachDescendant([&](SpatiallyNestablePointer object) {
-                locationChanged(true, false);
-                // we need to update attached queryAACubes in our own local tree so point-select always works
-                // however we don't want to flood the update pipeline with AvatarEntity updates, so we assume
-                // others have all info required to properly update queryAACube of AvatarEntities on their end
-                EntityItemPointer entity = std::dynamic_pointer_cast<EntityItem>(object);
-                bool iShouldTellServer = !(entity && entity->isAvatarEntity());
-                const bool force = false;
-                entityTree->updateEntityQueryAACube(object, packetSender, force, iShouldTellServer);
-            });
+            entityTree->updateEntityQueryAACube(shared_from_this(), packetSender, false, true);
         });
         bool isPhysicsEnabled = qApp->isPhysicsEnabled();
         bool zoneAllowsFlying = zoneInteractionProperties.first;
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -3150,21 +3150,30 @@ glm::vec3 EntityTree::getUnscaledDimensionsForID(const QUuid& id) {
     return glm::vec3(1.0f);
 }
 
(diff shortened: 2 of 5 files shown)


interface/src/avatar/MyAvatar.cpp
```diff
@@ -981,16 +981,7 @@ void MyAvatar::simulate(float deltaTime, bool inView) {
...
             EntityEditPacketSender* packetSender = qApp->getEntityEditPacketSender();
-            forEachDescendant([&](SpatiallyNestablePointer object) {
-                locationChanged(true, false);
-                // we need to update attached queryAACubes in our own local tree so point-select always works
-                // however we don't want to flood the update pipeline with AvatarEntity updates, so we assume
-                // others have all info required to properly update queryAACube of AvatarEntities on their end
-                EntityItemPointer entity = std::dynamic_pointer_cast<EntityItem>(object);
-                bool iShouldTellServer = !(entity && entity->isAvatarEntity());
-                const bool force = false;
-                entityTree->updateEntityQueryAACube(object, packetSender, force, iShouldTellServer);
```
HifiExperiments, line 992: Before, we were calling updateEntityQueryAACube once on every descendant, but updateEntityQueryAACube already updates all of the descendants of whatever it is called on.

libraries/entities/src/EntityTree.cpp
```diff
@@ -3150,21 +3150,30 @@ glm::vec3 EntityTree::getUnscaledDimensionsForID(const QUuid& id) {
...
+    glm::vec3 min(FLT_MAX);
+    glm::vec3 max(-FLT_MAX);
+
     // if the queryBox has changed, tell the entity-server
     EntityItemPointer entity = std::dynamic_pointer_cast<EntityItem>(object);
     if (entity) {
         bool queryAACubeChanged = false;
         if (!entity->hasChildren()) {
-            // updateQueryAACube will also update all ancestors' AACubes, so we only need to call this for leaf nodes
-            queryAACubeChanged = entity->updateQueryAACube();
```
HifiExperiments, line 3161: Before, we relied on each leaf node to call this to update all of its ancestors, but if an entity has multiple children, each child would update its entire ancestor tree multiple times.  Instead, each child now computes its queryAACube and passes it up to its parent, which aggregates them all just once.

libraries/shared/src/SpatiallyNestable.cpp
```diff
@@ -1171,9 +1171,39 @@ bool SpatiallyNestable::updateQueryAACube() {
...
+        auto parent = getParentPointer(success);
+        if (success && parent) {
+            parent->updateQueryAACube();
+        }
+    }
+
+    return true;
+}
+
+bool SpatiallyNestable::updateQueryAACubeWithDescendantAACube(const AACube& descendantAACube, bool updateParent) {
```
HifiExperiments, line 1184: this and the below method are identical to the existing methods, but they don't need to loop over all their descendants

MarcusLlewellyn:

MarcusLlewellyn:
Unofficial PR build for Windows is available here: [ProjectAthena-Alpha-PR281-35da672.exe](https://cdn.vircadia.com/dist/pr-builds/ProjectAthena-Alpha-PR281-35da672.exe)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR281/ProjectAthena-Alpha-PR281-97e3985_b6db0e13.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR281/ProjectAthena-Alpha-PR281-97e3985.tar_b6db0e13.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR281/ProjectAthena-Alpha-PR281-97e3985.tar_b6db0e13.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR281/ProjectAthena-Alpha-PR281-97e3985_b6db0e13.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR281/ProjectAthena-Alpha-PR281-97e3985_b6db0e13.exe

languagelovr:
Running the ModelParenting.js script caused the framerate to drop from 60 down to 1 on other builds while on this one (281) it went down to only around 30. I'd say this is a major improvement. Thanks HifiExperiments

kasenvr:
image
Works over the network, game rate drops on the client that is running the test script and its parented entities but not on the client observing.

kasenvr:
image

Editing of and and moving child entities around does in fact appear to be working.

PR #284, opened 2020-04-09
Use model's pivot

Phil-Palmer:
This PR allows model entities to match the pivot from their model, and allows a pivot to be outside the entity's extents.  Both are required for an efficient and familiar workflow for modellers, who set their pivots at the modelling stage to facilitate placement of the objects.  It also provides a correction for the shift that occurs when a model is edited and its extents change.

### How it works:
Nothing changes by default.  In the Create app, model entities have a new button: **Pivot > Use Model's Pivot**.  This sets the registration point based on the existing `EntityItemProperties::_naturalPosition` and `EntityItemProperties::_naturalDimensions`, so that it matches the pivot given to the object by the modeller.  It's analogous to the 'Reset Dimensions' button, but for positioning instead of size.

### Test plan:
1. Go to a domain whose server is running this PR's code.
2. Place a model entity using this model of a sink: http://djdrumphil.ca/hifi/pull-requests/use-models-pivot/WC_Sink.gltf
3. Observe that the default value used for the pivot is still (0.5, 0.5, 0.5).
4. Adjust the pivot XYZ values and observe that they can now be set outside the 0..1 range.
5. Click 'Use Model's Pivot'.  Observe that this sets the pivot to (0.5, -0.9111, 1), which is illustrated in the screenshot from 3ds Max below.
6. Set the entity's pivot Y to -100, so that the sink floats high in the sky.
7. CULLING: Fly close to the sink and observe that it still renders as it should even though its entity position is out of the frustum.
8. CULLING: With no other objects in view, set a breakpoint in `ModelMeshPartPayload::render` and count the number of different `this` pointers that are cycled through.  Observe  that with the sink in the frustum, it's one more than with the sink out the frustum.  This shows that the entity is correctly culled even when its position is far outside of its extents.
9. PHYSICS: Set the entity's Scale to 1000 and click Rescale.  Change its pivot Y to -10.  Try the different Collision Shape options on the entity and observe that in all cases, you can walk on its top surface even though the entity's position is far outside of its extents.
10. PICKING: In the Create app, click on and around the object.  Observe that the picking still works precisely even though the entity's position is far outside of its extents.

### Examples:
Here are some typical pivots in model files, that the new button allows us to preserve:

image

### For the future:
Ultimately we'd like an option for model entities to automatically **always** apply 'Reset Dimensions' and 'Use Model's Pivot', without having to use the buttons to correct each instance of an object whenever the geometry is edited.

### DIFF
--- a/scripts/system/create/edit.js
+++ b/scripts/system/create/edit.js
@@ -2489,6 +2489,35 @@ var PropertiesTool = function (opts) {
                     pushCommandForSelections();
                     selectionManager._update(false, this);
                 }
+            } else if (data.action === "resetToNaturalPivot") {
+                if (selectionManager.hasSelection()) {
+                    selectionManager.saveProperties();
+                    for (i = 0; i < selectionManager.selections.length; i++) {
+                        properties = selectionManager.savedProperties[selectionManager.selections[i]];
+                        var naturalDimensions = properties.naturalDimensions;
+
+                        // If any of the natural dimensions are not 0, set the entity's pivot according to its natural dimensions and natural position.
+                        // In this way, a model entity will match the pivot of its model.
+                        if (properties.type === "Model" && naturalDimensions.x === 0 && naturalDimensions.y === 0 &&
+                            naturalDimensions.z === 0) {
+                            Window.notifyEditError("Cannot apply the model's pivot to the entity: Model URL" +
+                                " is invalid or the model has not yet been loaded.");
+                        } else {
+                            var naturalMinimumExtent = Vec3.subtract(properties.naturalPosition, Vec3.multiply(naturalDimensions, 0.5));
+                            var negativeNaturalMinimumExtent = Vec3.multiply(naturalMinimumExtent, -1);
+
+                            var invNaturalDimensions = { x: 1.0 / naturalDimensions.x, y: 1.0 / naturalDimensions.y, z: 1.0 / naturalDimensions.z };
+
+                            var naturalPivot = Vec3.multiplyVbyV(negativeNaturalMinimumExtent, invNaturalDimensions);
+
+                            Entities.editEntity(selectionManager.selections[i], {
+                                registrationPoint: naturalPivot
+                            });
+                        }
+                    }
+                    pushCommandForSelections();
+                    selectionManager._update(false, this);
+                }
             } else if (data.action === "previewCamera") {
                 if (selectionManager.hasSelection()) {
                     Camera.mode = "entity";
--- a/libraries/entities/src/EntityItem.cpp
+++ b/libraries/entities/src/EntityItem.cpp
@@ -1795,8 +1795,7 @@ float EntityItem::getVolumeEstimate() const {
 void EntityItem::setRegistrationPoint(const glm::vec3& value) {
     if (value != _registrationPoint) {
         withWriteLock([&] {
-            _registrationPoint = glm::clamp(value, glm::vec3(ENTITY_ITEM_MIN_REGISTRATION_POINT), 
-                                                   glm::vec3(ENTITY_ITEM_MAX_REGISTRATION_POINT));
+            _registrationPoint = value;
         });
         dimensionsChanged(); // Registration Point affects the bounding box
         markDirtyFlags(Simulation::DIRTY_SHAPE);
--- a/libraries/entities/src/EntityItemProperties.cpp
+++ b/libraries/entities/src/EntityItemProperties.cpp
@@ -684,8 +684,8 @@ EntityPropertyFlags EntityItemProperties::getChangedProperties() const {
  * @property {Vec3} position=0,0,0 - The position of the entity in world coordinates.
  * @property {Quat} rotation=0,0,0,1 - The orientation of the entity in world coordinates.
  * @property {Vec3} registrationPoint=0.5,0.5,0.5 - The point in the entity that is set to the entity's position and is rotated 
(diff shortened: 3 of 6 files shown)


kasenvr:
WRT "For the future", the easy way to do that is for something I think is long overdue: Create App defaults.

So for example it always spawns entities as grabbable, this is because for VR users, it's a given that you want the world to be interactive. However, it's very troublesome when you are carefully building something and forgot about that, or even if you do remember... you still have to meticulously undo the "grabbable" aspect of it.

So, I think that things like that and this could be a setting that you can toggle as a default for all entities you work with going forward. This way, the user can decide which aspect is more important to them and their use case.

Nex-Pro:
@kasenvr I thought we had a menu item to turn off grabbable for new entities.

Nex-Pro:
Menu **Edit** =>(entities section)**Create entities as grabbable**

Nex-Pro:
@kasenvr clear that and your freshly created items will be non-grabbable.

MarcusLlewellyn:

Phil-Palmer:
> This PR requires a merge from upstream to compile. In particular, it needs to have PR #279 in it.

Thanks @MarcusLlewellyn, I've merged again from upstream now.

MarcusLlewellyn:
Nope, not a noob thing. Blame microsoft. :)

HifiExperiments:
I don’t think this is the best solution for this.  First, we probably shouldn’t piggyback on registration point.  Second, the toggle should be a proper model entity property with the implementation in c++ instead of the create app.

MarcusLlewellyn:
Unofficial PR build for Windows is available here: [ProjectAthena-Alpha-PR284-71358c2.exe](https://cdn.vircadia.com/dist/pr-builds/ProjectAthena-Alpha-PR284-71358c2.exe)

Phil-Palmer:
> I don’t think this is the best solution for this. First, we probably shouldn’t piggyback on registration point. Second, the toggle should be a proper model entity property with the implementation in c++ instead of the create app.

An actual property for this would be ideal @HifiExperiments :)
What we have here isn't a toggle, just a button that copies the model's pivot into the entity (in the same way that 'reset dimensions' copies the model's dimensions into the entity).

scripts/system/create/edit.js
```diff
@@ -2489,6 +2489,35 @@ var PropertiesTool = function (opts) {
...
+                        } else {
+                            var naturalMinimumExtent = Vec3.subtract(properties.naturalPosition, Vec3.multiply(naturalDimensions, 0.5));
+                            var negativeNaturalMinimumExtent = Vec3.multiply(naturalMinimumExtent, -1);
+
+                            var invNaturalDimensions = { x: 1.0 / naturalDimensions.x, y: 1.0 / naturalDimensions.y, z: 1.0 / naturalDimensions.z };
+
+                            var naturalPivot = Vec3.multiplyVbyV(negativeNaturalMinimumExtent, invNaturalDimensions);
+
+                            Entities.editEntity(selectionManager.selections[i], {
+                                registrationPoint: naturalPivot
```
HifiExperiments, line 2514: Instead of setting the registration point and removing the range etc, could we just set the position here?

Phil-Palmer: Hi, do you mean set the entity's position?  That's not the desired outcome; it's definitely the pivot we want to control.

Normally, when we set an entity's pivot, it's stored in the registration point.  Could you explain why it should be any different when we want the entity's pivot to match its model's pivot?
We could bypass the registration point and do something else for that case, but what would be the advantage?  It would seem to me like adding another pivot system alongside the one we have.

Thanks for the feedback :)

HifiExperiments: I’m concerned that this is overloading the use of registration point and that doing it this way will have unintended consequences for culling, physics, picking, etc, especially now that registration point isn’t clamped.  I also think this should be a top level entity property, so that people can properly toggle it.  This is an awesome feature and one that I expect lots of people will use, so it seems important to implement it in a future-proof way.  For example, this implementation only works using the built in Create app, but other scripters still have no way of doing it automatically.

Phil-Palmer: I see, thanks, those are good points.  So far I've been considering the registration point as just the pivot (local origin) by another name, but what's it actually for?  It's presented as a pivot in the Create app, and works like one, except that it's clamped within the geometry bounds and I didn't know why.  I thought it was most likely just clamped because no-one had tested it outside that range yet.  The tests I've done for culling, collision and picking have shown no problems with faraway pivots, but there could well be things I don't know about.

HifiExperiments: It has a similar effect, but I think it’s not a good idea to overload it like this.  The registration point is relative to the dimensions of the entity.  If someone models a scene and exports each model as a separate entity (with their own pivots), setting the pivots like this will of course work and position them properly in Interface...but the actual values of the registrations points will be meaningless.  Ideally the pivots would be presented in world coordinates, or not at all, and you’d just toggle the “Respect Original Pivot” button and not worry about the values at all.

Phil-Palmer:
I'm going to go back and try solving this the way @HifiExperiments suggested.  I'll open another PR when that's done.
Thanks for the help!

daleglass:
There seems to be an agreement that a new PR will be created, should we close this one then?

kasenvr:
We should revisit this issue, this may not be the best way to handle it but it should be handled at some point. :)

Phil-Palmer:
Update: @HifiExperiments is currently solving this in the way they suggested above :)
https://discord.com/channels/564926326025224212/656552319722323978/784903895679762482

digisomni:
@HifiExperiments is going to address in a later PR. :)

PR #289, opened 2020-04-11
Move python-distro import to where it is used

thoys:
Move python-distro import to where it is used, to prevent having to install this on osx/windows


thoys:
has already been handled in https://github.com/kasenvr/project-athena/commit/a1df2208de7981800c7f138c2991025dea7f2690

PR #292, opened 2020-04-12
Make custom Qt builds work on unsupported Ubuntu distributions

daleglass:
Okay, this is a bit embarrassing.

In the current code, Ubuntu 20.04 (and other versions without prebuilt packages) won't work right. This is because due to the lack of a "return" statement, the code will emit the error, then try to download the URL "NOT DEFINED", even if everything is set up to use an externally provided Qt.

The overall way this works is far from ideal, but proper fixes will need to wait until the master merge is complete, because those have a fair amount of changes to the build system. This will allow things to work with minimal changes meanwhile.


ctrlaltdavid:
Superseded by changes in `master` branch which does an equivalent thing: Qt code moved to hifi_qt.py which does `raise Exception(…)` instead of `return` in the two places.

PR #296, opened 2020-04-16
Add community-apps repo to default QML whitelist.

### LINKED ISSUE #295: More App does is not QML whitelisted
So apps that are added to the "More" app that use QML won't work because then the whitelist would have to be enabled with the kasenvr.github.io URL added to work. We should just add it as a default whitelisted domain.

kasenvr:
This cannot be tested until there is a QML based app in the community-apps repo, we should get the Spectator cam merged first in order to test this.

Closes #295

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3167,7 +3167,7 @@ void Application::showLoginScreen() {
 #endif
 }
 
-static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://kasenvr.github.io/community-apps/applications" };
+static const QUrl AUTHORIZED_EXTERNAL_QML_SOURCE { "https://cdn.vircadia.com/community-apps/applications" };
 
 void Application::initializeUi() {
 
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -3186,18 +3186,14 @@ void Application::initializeUi() {
             safeURLS += settingsSafeURLS;
 
             // END PULL SAFEURLS FROM INTERFACE.JSON Settings
-
-            bool isInWhitelist = false;  // assume unsafe
             
             if (AUTHORIZED_EXTERNAL_QML_SOURCE.isParentOf(url)) {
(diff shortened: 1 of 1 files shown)


kasenvr:
https://cdn.vircadia.com/community-apps/applications/spectator-camera/spectatorCamera.js

Tested that this works with that URL. The app-more is not updated here yet to pull from the cdn.vircadia.com URL, so manually loading the script will be necessary to test if it loads the related QML. You can test that it will reject an unauthorized QML by loading the spectator cam from the More app which is linked to kasenvr.github.io which is not an approved URL by default.

interface/src/Application.cpp
```diff
@@ -3188,12 +3188,18 @@ void Application::initializeUi() {
...
+            
+            if (AUTHORIZED_EXTERNAL_QML_SOURCE.isParentOf(url)) {
+                isInWhitelist = true;
+                return true;
+            } else {
+                for (const auto& str : safeURLS) {
+                    if (!str.isEmpty() && str.endsWith(".qml") && url.toString().endsWith(".qml") &&
+                        url.toString().startsWith(str)) {
+                        qCDebug(interfaceapp) << "Found matching url!" << url.host();
+                        isInWhitelist = true;
```
daleglass, line 3200: This variable appears to be set but otherwise unused.

kasenvr: This is true, I will remove it, can always add it back if we need it.

ctrlaltdavid:
Merged the PR into a test branch off v0.86.0-k2.1 branch, built, and tested running the following scripts:

https://cdn.vircadia.com/community-apps/applications/spectator-camera/spectatorCamera.js
Script runs and toolbar button successfully opens working dialog.
So the QML whitelisting of community apps is working.
Success.

https://kasenvr.github.io/community-apps/applications/spectator-camera/spectatorCamera.js
(Same script but different URL.)
Script runs but toolbar button opens a blank, non-working dialog.
So non-community QML is blocked.
Success.

PR #309, opened 2020-04-22
Procedural Shader applies to entities other than target

HifiExperiments:
Fixes https://github.com/kasenvr/project-athena/issues/166

Note: depends on the changes in https://github.com/kasenvr/project-athena/pull/276

Test plan:
- Start interface and go to the serverless domain with the default content.  Run [this](https://gist.github.com/HifiExperiments/ae9f742510012d465755d755affd94d4/raw/ccac684fa06b1c103b4ebdb838baf8228148637d/ShaderSwitching.js).
- You'll see a mannequin with a shader on its eyes.  On master, the shader will also apply to some of the other models around you.  Now, it'll only apply to the eyes.
- Press `o` to change the model and `i` to change the shader.  The shader should only ever apply to the model in front of you.


libraries/render-utils/src/MeshPartPayload.cpp
```diff
@@ -493,8 +493,7 @@ void ModelMeshPartPayload::render(RenderArgs* args) {
         batch.setDrawcallUniform(drawcallInfo);
     }
 
-    if (!_drawMaterials.empty() && _drawMaterials.top().material && _drawMaterials.top().material->isProcedural() &&
-            _drawMaterials.top().material->isReady()) {
+    if (_shapeKey.hasOwnPipeline()) {
```
HifiExperiments, line 496: This is the fix.  Before, it was possible for a ModelMeshPartPayload to have a non-custom shape key even though it was supposed to render with a shader.  In that case, it would render in the middle of when all the other models rendered, but would switch the active program to the custom shader, so all models after it would render with the custom shader.  Now, we use the shape key as the voice of truth, and make sure to update the shape key as necessary in RenderableModelEntityItem.cpp

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/ProjectAthena-Alpha-PR309-ff426ff_8f9c1802.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/ProjectAthena-Alpha-PR309-ff426ff.tar_8f9c1802.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/ProjectAthena-Alpha-PR309-ff426ff.tar_8f9c1802.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/ProjectAthena-Alpha-PR309-ff426ff_8f9c1802.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/ProjectAthena-Alpha-PR309-ff426ff_8f9c1802.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/Vircadia-Alpha-PR309-4502e7e.tar_9b47d924.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/Vircadia-Alpha-PR309-4502e7e.tar_9b47d924.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/Vircadia-Alpha-PR309-4502e7e_9b47d924.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/Vircadia-Alpha-PR309-4502e7e_9b47d924.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR309/Vircadia-Alpha-PR309-4502e7e_9b47d924.exe

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -1460,6 +1460,31 @@ void ModelEntityRenderer::doRenderUpdateSynchronousTyped(const ScenePointer& sce
...
+                    break;
+                }
+                material.pop();
+            }
+            if (!allProceduralMaterialsLoaded) {
+                break;
+            }
+        }
+        if (!allProceduralMaterialsLoaded) {
+            emit requestRenderUpdate();
```
daleglass, line 1481: I think this signal is called with the lock held, is that intentional?

daleglass (changes requested):

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
HifiExperiments: yep, that’s fine!

daleglass: Mm, ok then. My understanding is that Qt invokes a signal immediately if the receiver is in the same thread, so the lock would be held however long the handler takes, which seems undesirable.

However upon checking I suppose in this particular case it doesn't matter much, since the one signal handler there is uses a queued connection.

daleglass (approved):

ctrlaltdavid (approved):

JulianGro:
I am not seeing this problem in latest master, the "o" button also doesn't change the model for me. Switching the shaders with the "i" button does work.

kasenvr (approved):
Seems to work fine to me!

PR #310, opened 2020-04-23
[debug crash] unnecessary out-of-bounds array reference on load

odysseus654:
This patch skips a call to blend4 when minSize==0.  In this instance the function doesn't do anything (it's a for() loop that runs "minSize" times) and assembling the parameters retrieves the first value from four vectors, one of which we already know is empty if minSize==0.

### DIFF
--- a/libraries/animation/src/AnimBlendDirectional.cpp
+++ b/libraries/animation/src/AnimBlendDirectional.cpp
@@ -96,7 +96,9 @@ const AnimPoseVec& AnimBlendDirectional::evaluate(const AnimVariantMap& animVars
             }
         }
         _poses.resize(minSize);
-        blend4(minSize, &poseVecs[0][0], &poseVecs[1][0], &poseVecs[2][0], &poseVecs[3][0], &alphas[0], &_poses[0]);
+        if (minSize) {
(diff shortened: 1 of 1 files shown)


libraries/animation/src/AnimBlendDirectional.cpp
```diff
@@ -96,7 +96,9 @@ const AnimPoseVec& AnimBlendDirectional::evaluate(const AnimVariantMap& animVars
             }
         }
         _poses.resize(minSize);
-        blend4(minSize, &poseVecs[0][0], &poseVecs[1][0], &poseVecs[2][0], &poseVecs[3][0], &alphas[0], &_poses[0]);
+        if (minSize) {
```
thoys, line 99: Since it is not a boolean of a pointer I would prefer to see an actual comparison. While your solution probably works, it does make it look like a `if something exists` while you actually want to see right away that it should not be 0 or maybe 0 or lower?
```suggestion
        if (minSize != 0) {
```
```suggestion
        if (minSize > 0) {
```

ctrlaltdavid:
This fix is already in the master branch.

PR #311, opened 2020-04-23
[shutdown-deadlock] Restructure ContextAwareProfile from polling to notifications

odysseus654:
This concerns a class responsible for checking if webpages are restricted from making outside requests.  The previous implementation would do a (cross-thread blocking) retrieval of a QML property (with a one-second cache).  Unfortunately I have been seeing (from debugger, so not necessarily a real-world situation) a consistent deadlock where the value was being retrieved during app shutdown.

The new approach sets up a signal+slot attempting to forward any changes to RestrictFileAccess at the time they occur, the only time the QML value is retrieved is at the point of class construction.  Retrieving the value is done from a QAtomicInt with no blocking.

The only change in behavior should be that RestrictFileAccess is no longer (consistently) changeable from QML script.  Is this something that QML scripts would do?

### DIFF
--- a/libraries/ui/src/ui/types/ContextAwareProfile.cpp
+++ b/libraries/ui/src/ui/types/ContextAwareProfile.cpp
@@ -25,15 +25,15 @@ RestrictedContextMonitor::TMonitorMap RestrictedContextMonitor::gl_monitorMap;
 
 RestrictedContextMonitor::~RestrictedContextMonitor() {
     gl_monitorMapProtect.lock();
-    TMonitorMap::iterator lookup = gl_monitorMap.find(context);
+    TMonitorMap::iterator lookup = gl_monitorMap.find(_context);
     if (lookup != gl_monitorMap.end()) {
         gl_monitorMap.erase(lookup);
     }
     gl_monitorMapProtect.unlock();
 }
 
-RestrictedContextMonitor::TSharedPtr RestrictedContextMonitor::getMonitor(QQmlContext* context, bool createIfMissing) {
-    TSharedPtr monitor;
+RestrictedContextMonitor::TSharedPointer RestrictedContextMonitor::getMonitor(QQmlContext* context, bool createIfMissing) {
+    TSharedPointer monitor;
 
     gl_monitorMapProtect.lock();
     TMonitorMap::const_iterator lookup = gl_monitorMap.find(context);
@@ -41,8 +41,8 @@ RestrictedContextMonitor::TSharedPtr RestrictedContextMonitor::getMonitor(QQmlCo
         monitor = lookup->second.lock();
         assert(monitor);
     } else if(createIfMissing) {
-        monitor = std::make_shared<RestrictedContextMonitor>(context);
-        monitor->selfPtr = monitor;
+        monitor = TSharedPointer::create(context);
+        monitor->_selfPointer = monitor;
         gl_monitorMap.insert(TMonitorMap::value_type(context, monitor));
     }
     gl_monitorMapProtect.unlock();
@@ -55,19 +55,19 @@ ContextAwareProfile::ContextAwareProfile(QQmlContext* context) : ContextAwarePro
     _monitor = RestrictedContextMonitor::getMonitor(context, true);
     assert(_monitor);
     connect(_monitor.get(), &RestrictedContextMonitor::onIsRestrictedChanged, this, &ContextAwareProfile::onIsRestrictedChanged);
-    if (_monitor->isUninitialized) {
-        _monitor->isRestricted = isRestrictedGetProperty();
-        _monitor->isUninitialized = false;
+    if (_monitor->_isUninitialized) {
+        _monitor->_isRestricted = isRestrictedGetProperty();
+        _monitor->_isUninitialized = false;
     }
-    _isRestricted.store(_monitor->isRestricted ? 1 : 0);
+    _isRestricted.store(_monitor->_isRestricted ? 1 : 0);
 }
 
 void ContextAwareProfile::restrictContext(QQmlContext* context, bool restrict) {
-    RestrictedContextMonitor::TSharedPtr monitor = RestrictedContextMonitor::getMonitor(context, false);
+    RestrictedContextMonitor::TSharedPointer monitor = RestrictedContextMonitor::getMonitor(context, false);
 
     context->setContextProperty(RESTRICTED_FLAG_PROPERTY, restrict);
-    if (monitor && monitor->isRestricted != restrict) {
-        monitor->isRestricted = restrict;
+    if (monitor && monitor->_isRestricted != restrict) {
+        monitor->_isRestricted = restrict;
         monitor->onIsRestrictedChanged(restrict);
     }
 }
--- a/libraries/ui/src/ui/types/ContextAwareProfile.cpp
+++ b/libraries/ui/src/ui/types/ContextAwareProfile.cpp
@@ -22,19 +22,19 @@
 
 static const QString RESTRICTED_FLAG_PROPERTY = "RestrictFileAccess";
 
-QReadWriteLock ContextAwareProfile::gl_contextMapProtect;
-ContextAwareProfile::ContextMap ContextAwareProfile::gl_contextMap;
+QReadWriteLock ContextAwareProfile::_global_contextMapProtect;
+ContextAwareProfile::ContextMap ContextAwareProfile::_global_contextMap;
 
 ContextAwareProfile::ContextAwareProfile(QQmlContext* context) : ContextAwareProfileParent(context), _context(context) {
     assert(context);
 
     {   // register our object for future updates
-        QWriteLocker guard(&gl_contextMapProtect);
-        ContextMap::iterator setLookup = gl_contextMap.find(_context);
-        if (setLookup == gl_contextMap.end()) {
-            setLookup = gl_contextMap.insert(_context, ContextAwareProfileSet());
+        QWriteLocker guard(&_global_contextMapProtect);
+        ContextMap::iterator setLookup = _global_contextMap.find(_context);
+        if (setLookup == _global_contextMap.end()) {
+            setLookup = _global_contextMap.insert(_context, ContextAwareProfileSet());
         }
-        assert(setLookup != gl_contextMap.end());
+        assert(setLookup != _global_contextMap.end());
         ContextAwareProfileSet& profileSet = setLookup.value();
         assert(profileSet.find(this) == profileSet.end());
         profileSet.insert(this);
@@ -45,15 +45,15 @@ ContextAwareProfile::ContextAwareProfile(QQmlContext* context) : ContextAwarePro
 
 ContextAwareProfile::~ContextAwareProfile() {
     {  // deregister our object
-        QWriteLocker guard(&gl_contextMapProtect);
-        ContextMap::iterator setLookup = gl_contextMap.find(_context);
-        assert(setLookup != gl_contextMap.end());
-        if (setLookup != gl_contextMap.end()) {
+        QWriteLocker guard(&_global_contextMapProtect);
+        ContextMap::iterator setLookup = _global_contextMap.find(_context);
+        assert(setLookup != _global_contextMap.end());
+        if (setLookup != _global_contextMap.end()) {
             ContextAwareProfileSet& profileSet = setLookup.value();
             assert(profileSet.find(this) != profileSet.end());
             profileSet.remove(this);
             if (profileSet.isEmpty()) {
-                gl_contextMap.erase(setLookup);
+                _global_contextMap.erase(setLookup);
             }
         }
     }
@@ -65,15 +65,13 @@ void ContextAwareProfile::restrictContext(QQmlContext* context, bool restrict) {
     context->setContextProperty(RESTRICTED_FLAG_PROPERTY, restrict);
 
     // broadcast the new value to any registered ContextAwareProfile objects
-    {  // deregister our object
-        QReadLocker guard(&gl_contextMapProtect);
-        ContextMap::const_iterator setLookup = gl_contextMap.find(context);
-        if (setLookup != gl_contextMap.end()) {
-            const ContextAwareProfileSet& profileSet = setLookup.value();
-            for (ContextAwareProfileSet::const_iterator profileIterator = profileSet.begin();
-                 profileIterator != profileSet.end(); profileIterator++) {
-                (*profileIterator)->onIsRestrictedChanged(restrict);
-            }
+    QReadLocker guard(&_global_contextMapProtect);
+    ContextMap::const_iterator setLookup = _global_contextMap.find(context);
+    if (setLookup != _global_contextMap.end()) {
+        const ContextAwareProfileSet& profileSet = setLookup.value();
+        for (ContextAwareProfileSet::const_iterator profileIterator = profileSet.begin();
+                profileIterator != profileSet.end(); profileIterator++) {
+            (*profileIterator)->onIsRestrictedChanged(restrict);
         }
     }
 }
--- a/libraries/ui/src/ui/types/ContextAwareProfile.cpp
+++ b/libraries/ui/src/ui/types/ContextAwareProfile.cpp
@@ -38,12 +38,12 @@ RestrictedContextMonitor::TSharedPointer RestrictedContextMonitor::getMonitor(QQ
     gl_monitorMapProtect.lock();
     TMonitorMap::const_iterator lookup = gl_monitorMap.find(context);
     if (lookup != gl_monitorMap.end()) {
-        monitor = lookup->second.lock();
+        monitor = lookup.value().lock();
         assert(monitor);
     } else if(createIfMissing) {
         monitor = TSharedPointer::create(context);
         monitor->_selfPointer = monitor;
-        gl_monitorMap.insert(TMonitorMap::value_type(context, monitor));
+        gl_monitorMap.insert(context, monitor);
     }
     gl_monitorMapProtect.unlock();
     return monitor;
--- a/libraries/ui/src/ui/types/ContextAwareProfile.cpp
+++ b/libraries/ui/src/ui/types/ContextAwareProfile.cpp
@@ -12,6 +12,7 @@
 #include "ContextAwareProfile.h"
 
 #include <cassert>
+#include <QtCore/QMutexLocker>
 #include <QtCore/QThread>
 #include <QtQml/QQmlContext>
 
@@ -24,18 +25,17 @@ QMutex RestrictedContextMonitor::gl_monitorMapProtect;
 RestrictedContextMonitor::TMonitorMap RestrictedContextMonitor::gl_monitorMap;
 
 RestrictedContextMonitor::~RestrictedContextMonitor() {
-    gl_monitorMapProtect.lock();
+    QMutexLocker locker(&gl_monitorMapProtect);
     TMonitorMap::iterator lookup = gl_monitorMap.find(_context);
     if (lookup != gl_monitorMap.end()) {
         gl_monitorMap.erase(lookup);
     }
-    gl_monitorMapProtect.unlock();
 }
 
 RestrictedContextMonitor::TSharedPointer RestrictedContextMonitor::getMonitor(QQmlContext* context, bool createIfMissing) {
     TSharedPointer monitor;
 
-    gl_monitorMapProtect.lock();
+    QMutexLocker locker(&gl_monitorMapProtect);
     TMonitorMap::const_iterator lookup = gl_monitorMap.find(context);
     if (lookup != gl_monitorMap.end()) {
         monitor = lookup.value().lock();
@@ -45,7 +45,6 @@ RestrictedContextMonitor::TSharedPointer RestrictedContextMonitor::getMonitor(QQ
         monitor->_selfPointer = monitor;
         gl_monitorMap.insert(context, monitor);
     }
-    gl_monitorMapProtect.unlock();
     return monitor;
 }
 
--- a/libraries/ui/src/ui/types/ContextAwareProfile.cpp
+++ b/libraries/ui/src/ui/types/ContextAwareProfile.cpp
@@ -22,19 +22,19 @@
 
 static const QString RESTRICTED_FLAG_PROPERTY = "RestrictFileAccess";
 
-QReadWriteLock ContextAwareProfile::_global_contextMapProtect;
-ContextAwareProfile::ContextMap ContextAwareProfile::_global_contextMap;
+QReadWriteLock ContextAwareProfile::_contextMapProtect;
+ContextAwareProfile::ContextMap ContextAwareProfile::_contextMap;
 
 ContextAwareProfile::ContextAwareProfile(QQmlContext* context) : ContextAwareProfileParent(context), _context(context) {
     assert(context);
 
     {   // register our object for future updates
-        QWriteLocker guard(&_global_contextMapProtect);
-        ContextMap::iterator setLookup = _global_contextMap.find(_context);
-        if (setLookup == _global_contextMap.end()) {
-            setLookup = _global_contextMap.insert(_context, ContextAwareProfileSet());
+        QWriteLocker guard(&_contextMapProtect);
+        ContextMap::iterator setLookup = _contextMap.find(_context);
+        if (setLookup == _contextMap.end()) {
+            setLookup = _contextMap.insert(_context, ContextAwareProfileSet());
         }
-        assert(setLookup != _global_contextMap.end());
+        assert(setLookup != _contextMap.end());
         ContextAwareProfileSet& profileSet = setLookup.value();
         assert(profileSet.find(this) == profileSet.end());
         profileSet.insert(this);
@@ -44,17 +44,16 @@ ContextAwareProfile::ContextAwareProfile(QQmlContext* context) : ContextAwarePro
 }
 
 ContextAwareProfile::~ContextAwareProfile() {
-    {  // deregister our object
-        QWriteLocker guard(&_global_contextMapProtect);
-        ContextMap::iterator setLookup = _global_contextMap.find(_context);
-        assert(setLookup != _global_contextMap.end());
-        if (setLookup != _global_contextMap.end()) {
-            ContextAwareProfileSet& profileSet = setLookup.value();
-            assert(profileSet.find(this) != profileSet.end());
-            profileSet.remove(this);
-            if (profileSet.isEmpty()) {
-                _global_contextMap.erase(setLookup);
-            }
+    // deregister our object
+    QWriteLocker guard(&_contextMapProtect);
+    ContextMap::iterator setLookup = _contextMap.find(_context);
+    assert(setLookup != _contextMap.end());
+    if (setLookup != _contextMap.end()) {
+        ContextAwareProfileSet& profileSet = setLookup.value();
+        assert(profileSet.find(this) != profileSet.end());
+        profileSet.remove(this);
+        if (profileSet.isEmpty()) {
+            _contextMap.erase(setLookup);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


odysseus654:
A slight rephrasing as to what this PR does and its current state:

Origin: I caught this multiple times while testing shutdown-deadlocks in debug build in Visual Studio, repeatedly opening and closing the interface and seeing if at some point it gets into a point where there is no discernible progress on shutting down.

Diagnostic: The focus of this patch is on ContextAwareProfile, a filter catching all web requests from the scripting engine before they're permitted to be made, most likely XHR requests from installed apps in this case.  The existing code (outside of a one-second cached value) issues a QML property request, which can only be executed on the main thread.  An unconditional thread call is executed but never completed as the main thread is in the process of shutting down and is not receiving any requests.

Current state: I have tested this code a few times before pushing this PR out there.  I did not test it after rebasing it to master (my tests were against K2).  I likely need to adjust the variable names to be more consistent with coding requirements (although I'm not sure what the coding says for static-member names, I'm tempted to still use "gl_").  The code is also using std::shared_ptr where QSharedPointer may be preferred.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-aefd20c_58c7f2e9.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-aefd20c_58c7f2e9.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-9899b47_ffe395da.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-9899b47_ffe395da.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-4df9e21_ffe395da.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-4df9e21_ffe395da.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-4df9e21.tar_ffe395da.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-4df9e21.tar_ffe395da.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-4df9e21_ffe395da.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-72cfa66_7ad6dadc.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-72cfa66.tar_7ad6dadc.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-72cfa66_7ad6dadc.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-72cfa66.tar_7ad6dadc.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-72cfa66_7ad6dadc.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-7317d0b_57d8da13.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-7317d0b.tar_57d8da13.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-7317d0b.tar_57d8da13.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/ProjectAthena-Alpha-PR311-7317d0b_57d8da13.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-b3a9432_57d8da13.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-b3a9432.tar_57d8da13.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-b3a9432_57d8da13.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-b3a9432.tar_57d8da13.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-b3a9432_57d8da13.exe

odysseus654:
Just did a force-rebase of this at the git level on to the tip of master, if that helps people that can't do the auto-gen builds.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-acf87f3_47c96fcf.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-acf87f3_47c96fcf.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-acf87f3.tar_47c96fcf.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-acf87f3.tar_47c96fcf.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-acf87f3_47c96fcf.exe

JulianGro:
I am pretty sure that this fixes the problem on my end.
In the last 5 tries, the interface shut down properly.
I will keep using this build for a while and report back if I find any problems.

odysseus654:
While there may still be blocking invokes in this, none of them are accessible from ContextAwareProfile::isRestricted.  The purpose of the monitor is to monitor any changes to the value of isRestricted and update the class with the updated value, rather than doing so on-demand from the scripting engine.

This particular issue is solving an actual deadlock while waiting for script threads to stop.  If one of the script threads hits ContextAwareProfile::isRestricted (say while doing an XHR query), it will attempt to execute a blocking query to the main thread while the main thread is in the middle of its destructor waiting for the script threads to shut down (where it's not willing to entertain new property retrieval requests)

odysseus654:
FWIW I was regularly reproing this in debugger while doing some smoketesting of debug builds, which is how I was able to characterize and write code for this issue. How easy it is to repro may depend on how chatty the buttons you have in your toolbar are.

HifiExperiments:
in theory, couldn't a script thread call any number of other blocking methods during shutdown and cause the same problem?

odysseus654:
Perhaps? I haven't reviewed the scripting API interface to see how many calls are blocking so I don't know how many others are out there really.  This one removes a blocking call that seems relatively unnecessary and often used (and may cause a slight performance boost too).

libraries/ui/src/ui/types/ContextAwareProfile.cpp
```diff
@@ -20,20 +21,60 @@
...
+    _isRestricted.store(_monitor->_isRestricted ? 1 : 0);
+}
 
 void ContextAwareProfile::restrictContext(QQmlContext* context, bool restrict) {
+    RestrictedContextMonitor::TSharedPointer monitor = RestrictedContextMonitor::getMonitor(context, false);
+
     context->setContextProperty(RESTRICTED_FLAG_PROPERTY, restrict);
+    if (monitor && monitor->_isRestricted != restrict) {
+        monitor->_isRestricted = restrict;
+        monitor->onIsRestrictedChanged(restrict);
```
HifiExperiments, line 70: Ok, I understand that the core of the fix is removing the blocking call from ContextAwareProfile::isRestricted(), but I'm still not sure what the purpose of the monitor is.  Couldn't we just directly set ContextAwareProfile::_isRestricted here?

libraries/ui/src/ui/types/ContextAwareProfile.h
```diff
@@ -48,9 +78,13 @@ class ContextAwareProfile : public ContextAwareProfileParent {
 
     ContextAwareProfile(QQmlContext* parent);
     QQmlContext* _context{ nullptr };
-    bool _cachedValue{ false };
-    quint64 _cacheExpiry{ 0 };
-    constexpr static quint64 MAX_CACHE_AGE = MSECS_PER_SECOND;
+    QAtomicInt _isRestricted{ 0 };
```
HifiExperiments, line 81: couldn't you use an `std::atomic<bool>` here?

libraries/ui/src/ui/types/ContextAwareProfile.cpp
odysseus654: Jeez it's been a while since I've looked at this code.  I think the key point of the monitor is this line:

connect(_monitor.get(), &RestrictedContextMonitor::onIsRestrictedChanged, this, &ContextAwareProfile::onIsRestrictedChanged);

I'm thinking I've got one monitor object for each QQmlContext, which serves as something for ContextAwareProfile to subscribe to changes to.  I'm assuming with this approach there may be many more ContextAwareProfile's than QQmlContext's.

The RestrictedContextMonitor does a blocking property-get on construction, after which it assumes that any further change calls can be caught by restrictContext

libraries/ui/src/ui/types/ContextAwareProfile.h
odysseus654: For this exercise I tried to de-STL my code as much as possible, using the Qt versions of code when both were available.

libraries/ui/src/ui/types/ContextAwareProfile.cpp
HifiExperiments: Sure, but couldn't you accomplish the same thing without the monitor by storing the "c++" side version of the bool directly, instead of using the monitor?  You'd still do `_isRestricted =isRestrictedGetProperty();` in the initializer, and you'd still catch the changes here.  The monitor just seems like added complexity (unless the signalling is actually necessary to solve the bug, but from your explanation, it seems like that's not the case).

libraries/ui/src/ui/types/ContextAwareProfile.h
HifiExperiments: it's actually probably in our best interest to decouple ourselves from Qt as much as possible!  we didn't always adhere to this rule (as you can tell since the code is full of Qt data types) but near the end we tried to replace things with STL as we went

libraries/ui/src/ui/types/ContextAwareProfile.cpp
odysseus654: If we could guarantee that we could locate all the ContextAwareProfile  objects that were in existence at the time that the value changed (or guaranteed that there was only one value for that specific QQmlContext) then I would agree with you, otherwise this is a way to distribute the changes to all related objects.

libraries/ui/src/ui/types/ContextAwareProfile.h
odysseus654: If you would prefer this changed to an atomic<bool> then that could be done, there's an awful lot of other Qt'isms in this code though and I don't really expect them all to be changed over.  I'm also not expecting this to turn into a debate of STL vs Qt (of which my opinion has been changing multiple times over the last few weeks)

daleglass: Yes, I'm not sure if it's worthwhile at this point for us to spend a lot of time on being independent from Qt. We're a much smaller project than hifi was, converting code risks introducing bugs without improving functionality, and it makes it harder to examine the code's history if you end up with hifi on one side and STL on the other.

Writing new code based on the STL in places where this is possible without having to convert a bunch of other stuff is probably a good idea, though.

odysseus654: [off-topic] if it helps, my current position is:
- I trust QDeadlockTimer/QElapsed timer much more than the STL versions (the STL "high resolution" objects HiFi uses in the existing UDT code are deprecated as meaningless)
- the STL map<> objects are better than QMap<> when one needs a custom comparator (e.g. case-insensitive) or need std::move support, neither of which Qt does.

HifiExperiments: my original point was also that this could be a bool instead of an int, and I don’t think there’s a QAtomicBool

libraries/ui/src/ui/types/ContextAwareProfile.cpp
HifiExperiments: Could we keep the static map, but make the values the ContextAwareProfiles?

HifiExperiments: Wait, also, wasn’t the original code one value per thing?  How did we keep them synchronized before?

odysseus654: I think prior to this patch the only official value was in QML, each ContextAwareProfile would be querying the QML profile with a one-second cache time.  I'm guessing the cache was there for performance reasons, but that doesn't really help if the cache runs out at the wrong time.

I'm thinking we can't really change the static map to ContextAwareProfiles unless we make it something like map<QQmlContext,list<ContextAwareProfile>>.  We then also start having a potential lock delay as construction/destruction of ContextAwareProfile would require mutexed access to a static object.  Again it depends on how many ContextAwareProfile objects there are vs QQmlContext objects they are associated with.

libraries/ui/src/ui/types/ContextAwareProfile.h
odysseus654: Answer is the same as before, If you would prefer this changed to an atomic then that could be done.

libraries/ui/src/ui/types/ContextAwareProfile.cpp
HifiExperiments: I still don’t really understand the need for the map.  Before, each ContentAwareProfile queried their context for the value.  Can’t we keep that, but only query the values from QML at initialization, and then keep track of the changes in a local member variable?  I see what you’re saying about it not working if there’s multiple Profiles created from one Context, but that case wasn’t handled by the old code, which makes me think it can’t really happen, or isn’t important if it does

odysseus654: The issue is in the "keep track of the changes in a local member variable". ContextAwareProfile::restrictContext() is a static method and only specifies QQmlContext*, so unless there's a way of storing something new in the QQmlContext (which seems like a generic Qt interface), we're gonna need a static map in any case if we're ever expecting to reach ContextAwareProfile.  The only real question is what it should map to, i created a new class to store the value for the QQmlContext that permits a Qt subscription.

HifiExperiments: OH fuck me sorry I didn’t see that it was static, you’re totally right.  can we try something like...std::unordered_map<QQmlContext*, std::unordered_set<ContextAwareProfile*>> (or the Qt equivalents, but at least a QHash instead of a QMap)

odysseus654: Thank you for going down this rabbit hole with me, this really did need a second pair of eyes.

The monitor object in this file is serving a purpose similar to the std::unordered_map you are suggesting.  Pros and cons of each approach:

monitor class:
- uses the Qt signal/slot approach as a pub/sub to handle the change broadcasts for us
- adds another class to the file that could be considered additional complexity
- would need to hold a static mutex whenever a QQmlContext is seen for the first time, or the last ContextAwareProfile using a specific QQmlContext is destructed

std::unordered_map:
 - would be responsible for iterating over all the entries in the map whenever a change needs to be broadcast
 - would need to hold a static mutex whenever a new ContextAwareProfile is constructed or destructed, or a broadcast is in progress (potential lock contention)

HifiExperiments: Can we try the map approach?  I think it will make this code a lot more readable.  I’m not super concerned about the number of these things that we make (looking at the rest of the code, it doesn’t seem like a problem) or the locking, since if there is locking around just the map access, it should be pretty quick.

odysseus654:
Okay, looks like I have change requests from both threads in the preceding review.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-edd6322.tar_4f5f46a6.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-edd6322.tar_4f5f46a6.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-edd6322_4f5f46a6.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-edd6322_4f5f46a6.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-edd6322_4f5f46a6.exe

libraries/ui/src/ui/types/ContextAwareProfile.cpp
```diff
@@ -12,28 +12,76 @@
...
+    }
+}
 
 void ContextAwareProfile::restrictContext(QQmlContext* context, bool restrict) {
+
+    // set the QML property
     context->setContextProperty(RESTRICTED_FLAG_PROPERTY, restrict);
+
+    // broadcast the new value to any registered ContextAwareProfile objects
+    {  // deregister our object
```
HifiExperiments, line 68: looks like this comment was copied over erroneously

libraries/ui/src/ui/types/ContextAwareProfile.cpp
```diff
@@ -47,11 +95,10 @@ bool ContextAwareProfile::isRestrictedInternal() {
     return true;
 }
 
+void ContextAwareProfile::onIsRestrictedChanged(bool newValue) {
```
HifiExperiments, line 98: maybe `setRestricted` instead?

libraries/ui/src/ui/types/ContextAwareProfile.h
```diff
@@ -47,10 +49,18 @@ class ContextAwareProfile : public ContextAwareProfileParent {
...
+
     QQmlContext* _context{ nullptr };
-    bool _cachedValue{ false };
-    quint64 _cacheExpiry{ 0 };
-    constexpr static quint64 MAX_CACHE_AGE = MSECS_PER_SECOND;
+    std::atomic<bool> _isRestricted{ false };
+
+private:
+    typedef QSet<ContextAwareProfile*> ContextAwareProfileSet;
+    typedef QMap<QQmlContext*, ContextAwareProfileSet> ContextMap;
```
HifiExperiments, line 60: a QMap is actually an ordered map, but we don't care about the order so this could just be a QHash.  a QMap is actually a red-black tree, whereas a QHash is really a hash map

libraries/ui/src/ui/types/ContextAwareProfile.h
```diff
@@ -47,10 +49,18 @@ class ContextAwareProfile : public ContextAwareProfileParent {
...
+
+    static QReadWriteLock gl_contextMapProtect;
+    static ContextMap gl_contextMap;
```
HifiExperiments, line 63: I think dropping the `gl_` prefix would be more in line with the coding standard

odysseus654: Almost tempted to raise an objection here -- does the coding standard have anything in it to handle global variables, and should it? Should these variables look the same as global variables? Should they be all caps?

HifiExperiments: Actually one could argue that even though these are static, they are private, so they actually should start with an underscore: https://github.com/kasenvr/project-athena/blob/kasen/core/CODING_STANDARD.md#112-private-class-variables-must-be-in-mixed-case-prefixed-with-an-underscore

One could also argue that this rule implies global variables should NOT start with an underscore (although we actually don’t really follow this rule at all): https://github.com/kasenvr/project-athena/blob/kasen/core/CODING_STANDARD.md#119-global-variables-should-always-be-referred-to-using-the--operator

So...I’m open (but we don’t use gl_ anywhere that I’ve seen so I would vote against that)

odysseus654: I guess if nothing else we could just go with a _global_ prefix

libraries/ui/src/ui/types/ContextAwareProfile.cpp
odysseus654: This is called by the notification broadcaster and strictly from within the object, so setRestricted feels less appropriate.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-e6d95cc_0c7aab15.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-e6d95cc.tar_0c7aab15.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-e6d95cc.tar_0c7aab15.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-e6d95cc_0c7aab15.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-e6d95cc_0c7aab15.exe

libraries/ui/src/ui/types/ContextAwareProfile.cpp
HifiExperiments: it’s called from the static method (otherwise, since it’s just called in one place, I’d recommend getting rid of the method entirely).  but per https://github.com/kasenvr/project-athena/blob/kasen/core/CODING_STANDARD.md#121-the-terms-getset-must-be-used-where-an-attribute-is-accessed-directly it should still be setRestricted

libraries/ui/src/ui/types/ContextAwareProfile.h
```diff
@@ -47,10 +49,19 @@ class ContextAwareProfile : public ContextAwareProfileParent {
...
+    typedef QHash<QQmlContext*, ContextAwareProfileSet> ContextMap;
+
     QQmlContext* _context{ nullptr };
-    bool _cachedValue{ false };
-    quint64 _cacheExpiry{ 0 };
-    constexpr static quint64 MAX_CACHE_AGE = MSECS_PER_SECOND;
+    std::atomic<bool> _isRestricted{ false };
+
+    static QReadWriteLock _global_contextMapProtect;
+    static ContextMap _global_contextMap;
```
HifiExperiments, line 62: eh I’m still not sure about these.  we don’t use the _global_ prefix anywhere else.  I’d settle for _globalContextMap and _globalContextMapProtect though?

odysseus654: So there is nothing in the coding standard that permits global variables to be identified as such? It seems like that's very similar to the reason why all member variables must start with _

libraries/ui/src/ui/types/ContextAwareProfile.cpp
odysseus654: I can see that name being confused by anyone looking at it as an API.  Since it's simplified down to a simple variable assignment at this point I'm tempted to just change it to a direct variable assignment and get rid of the function then.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-04eb2dc_37c613c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-04eb2dc.tar_37c613c3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-04eb2dc_37c613c3.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-04eb2dc.tar_37c613c3.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-04eb2dc_37c613c3.exe

daleglass:
Discussion during meeting: this one is very tricky to test, and few people feel capable of reviewing it.

After code reviews if nothing explodes we'll merge it

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-a1d8fd5_37c613c3.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-a1d8fd5.tar_37c613c3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-a1d8fd5.tar_37c613c3.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-a1d8fd5_37c613c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-a1d8fd5_37c613c3.exe

daleglass (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-190fd26.tar_37c613c3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-190fd26.tar_37c613c3.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-190fd26_37c613c3.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-190fd26_37c613c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR311/Vircadia-Alpha-PR311-190fd26_37c613c3.exe

JulianGro:
Gave it a spin again and didn't see any problems.

kasenvr (approved):
Well, under normal use it appears to be working just fine. Interface doesn't close any faster but it doesn't seem to hang on exit perpetually. Therefore I'll consider this working as intended! :)

odysseus654:
As this is a race condition it would be difficult to completely say "good" but the whole thing is "does the window ever not-close?"  And I think other than me and at least one other person on here that has had problems with the window closing... I'm not really expecting differing behavior other than that no longer happening.

The only real possible malfunction with this code is it not properly flagging web requests as "restricted" or not.  This would be more visible in malfunctions in local apps not being able to read from files (or remote apps somehow able to read from local files?)

PR #334, opened 2020-05-04
Change standard snapshot name to vircadia

jug007:
This should change the name when taking a snapshot (screenshot) to vircadia-snap-by-... rather than hifi-snap-by-...

I just ran `grep -r hifi-snap-by *` to find all related mentions in the source code.


ctrlaltdavid:
Built, tested, and checked JSDoc output. All OK.

PR #338, opened 2020-05-08
Bump packet version size up.

kasenvr:
It is as the comment says in the code. I've tested this on my local machine with a local domain-server and confirmed it to be working.


ctrlaltdavid:
Tested OK.

PR #340, opened 2020-05-09
Attempt to fix #337 by ensuring the object is marked as dead.

### LINKED ISSUE #337: master: PhysicalEntitySimulation::prepareEntityForDelete(EntityItemPointer): Assertion `entity->isDead()' failed.
```interface: /home/dale/git/vircadia/vircadia-master/libraries/physics/src/PhysicalEntitySimulation.cpp:246: virtual void PhysicalEntitySimulation::prepareEntityForDelete(EntityItemPointer): Assertion `entity->isDead()' failed.```

I investigated this slightly. `PhysicalEntitySimulation::prepareEntityForDelete` checks if `entity->isDead()` is true. This seems sensible. The entity can be marked as dead by simply calling `entity->die()`, which simply sets an internal variable.

So I take this is simply a sanity check, and just adding a `die()` in the right place would fix things.

The likely culprit seems to be `OtherAvatar::removeOrb`. It was recently modified in commit ca8d64bc05c which is likely to be the problem.

```
(gdb) bt
#0  0x00007fffe6cc4625 in raise () from /lib64/libc.so.6
#1  0x00007fffe6cad8d9 in abort () from /lib64/libc.so.6
#2  0x00007fffe6cad7a9 in __assert_fail_base.cold () from /lib64/libc.so.6
#3  0x00007fffe6cbca66 in __assert_fail () from /lib64/libc.so.6
#4  0x00007ffff6d6b025 in PhysicalEntitySimulation::prepareEntityForDelete (this=0x1322990, entity=warning: RTTI symbol not found for class 'std::_Sp_counted_deleter<ShapeEntityItem*, ShapeEntityItem::baseFactory(EntityItemID const&, EntityItemProperties const&)::{lambda(EntityItem*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>'
warning: RTTI symbol not found for class 'std::_Sp_counted_deleter<ShapeEntityItem*, ShapeEntityItem::baseFactory(EntityItemID const&, EntityItemProperties const&)::{lambda(EntityItem*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>'
std::shared_ptr<EntityItem> (use count 15, weak count 1) = {...})
    at /home/dale/git/vircadia/vircadia-master/libraries/physics/src/PhysicalEntitySimulation.cpp:246
#5  0x00007ffff4ecffde in EntityTree::processRemovedEntities (this=0x16ab4920, theOperator=...)
    at /home/dale/git/vircadia/vircadia-master/libraries/entities/src/EntityTree.cpp:787
#6  0x00007ffff4ecfc28 in EntityTree::deleteEntitiesByPointer (this=0x16ab4920, entities=std::vector of length 1, capacity 1 = {...})
    at /home/dale/git/vircadia/vircadia-master/libraries/entities/src/EntityTree.cpp:753
#7  0x00007ffff4ea6fbc in EntityScriptingInterface::<lambda()>::operator()(void) const (__closure=0x7fffffff8350)
    at /home/dale/git/vircadia/vircadia-master/libraries/entities/src/EntityScriptingInterface.cpp:997
#8  0x00007ffff4eb3ed0 in ReadWriteLockable::withWriteLock<EntityScriptingInterface::deleteEntity(const QUuid&)::<lambda()> >(EntityScriptingInterface::<lambda()> &&) const
    (this=0x16ab4940, f=...) at /home/dale/git/vircadia/vircadia-master/libraries/shared/src/shared/ReadWriteLockable.h:64
#9  0x00007ffff4ea7136 in EntityScriptingInterface::deleteEntity (this=0x3ba0830, id=...)
    at /home/dale/git/vircadia/vircadia-master/libraries/entities/src/EntityScriptingInterface.cpp:977
#10 0x0000000000c4ae06 in OtherAvatar::removeOrb (this=0x1a58c900) at /home/dale/git/vircadia/vircadia-master/interface/src/avatar/OtherAvatar.cpp:59
#11 0x0000000000b920b6 in AvatarManager::updateOtherAvatars (this=0x3c0eb10, deltaTime=0.0165917613)
    at /home/dale/git/vircadia/vircadia-master/interface/src/avatar/AvatarManager.cpp:315
#12 0x00000000009ee9cb in Application::update (this=0x7fffffffaf30, deltaTime=0.0165917613) at /home/dale/git/vircadia/vircadia-master/interface/src/Application.cpp:6567
#13 0x00000000009e39b7 in Application::idle (this=0x7fffffffaf30) at /home/dale/git/vircadia/vircadia-master/interface/src/Application.cpp:5271
#14 0x00000000009df0da in Application::event (this=0x7fffffffaf30, event=0x7fff0deb89b0) at /home/dale/git/vircadia/vircadia-master/interface/src/Application.cpp:4212
#15 0x00007ffff45042a2 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Widgets.so.5
#16 0x00007ffff450d330 in QApplication::notify(QObject*, QEvent*) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Widgets.so.5
#17 0x00000000009deed0 in Application::notify (this=0x7fffffffaf30, object=0x7fffffffaf30, event=0x7fff0deb89b0)
    at /home/dale/git/vircadia/vircadia-master/interface/src/Application.cpp:4182
#18 0x00007fffe7425d22 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#19 0x00007fffe7428778 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#20 0x00007fffe7479873 in ?? () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#21 0x00007fffe68dc570 in g_main_context_dispatch () from /lib64/libglib-2.0.so.0
#22 0x00007fffe68dc900 in g_main_context_iterate.isra () from /lib64/libglib-2.0.so.0
#23 0x00007fffe68dc9a3 in g_main_context_iteration () from /lib64/libglib-2.0.so.0
#24 0x00007fffe7478f11 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#25 0x00007fffe7424b0b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#26 0x00007fffe742c11e in QCoreApplication::exec() () from /home/dale/git/vircadia/qt5-install/lib/libQt5Core.so.5
#27 0x0000000000cad11e in main (argc=1, argv=0x7fffffffe4c8) at /home/dale/git/vircadia/vircadia-master/interface/src/main.cpp:444
(gdb) frame 4
#4  0x00007ffff6d6b025 in PhysicalEntitySimulation::prepareEntityForDelete (this=0x1322990, entity=warning: RTTI symbol not found for class 'std::_Sp_counted_deleter<ShapeEntityItem*, ShapeEntityItem::baseFactory(EntityItemID const&, EntityItemProperties const&)::{lambda(EntityItem*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>'
warning: RTTI symbol not found for class 'std::_Sp_counted_deleter<ShapeEntityItem*, ShapeEntityItem::baseFactory(EntityItemID const&, EntityItemProperties const&)::{lambda(EntityItem*)#1}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>'
std::shared_ptr<EntityItem> (use count 15, weak count 1) = {...})

daleglass:
Currently interface crashes very quickly on master when built with debug enabled, see #337 

This patch attempts to fix that, and it seems to work, but issues almost instantly crop up elsewhere, somewhere from the scripting engine apparently. So this fix could well be completely incorrect.

Submitting this for feedback.

### DIFF
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -784,6 +784,7 @@ void EntityTree::processRemovedEntities(const DeleteEntityOperator& theOperator)
             }
         }
         if (theEntity->isSimulated()) {
+            theEntity->die();
             _simulation->prepareEntityForDelete(theEntity);
         }
     }


interface/src/avatar/OtherAvatar.cpp
```diff
@@ -56,7 +56,7 @@ OtherAvatar::~OtherAvatar() {
 
 void OtherAvatar::removeOrb() {
     if (!_otherAvatarOrbMeshPlaceholderID.isNull()) {
-        DependencyManager::get<EntityScriptingInterface>()->deleteEntity(_otherAvatarOrbMeshPlaceholderID);
+        DependencyManager::get<EntityScriptingInterface>()->deleteEntity(_otherAvatarOrbMeshPlaceholderID, true);
```
HifiExperiments, line 59: I think there must be something deeper happening here.  Can we find the commit that introduced the crash and determine what changed in the logic?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-fff2044.tar_38b522b7.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-fff2044_38b522b7.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-fff2044.tar_38b522b7.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-fff2044_38b522b7.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-fff2044_38b522b7.exe

interface/src/avatar/OtherAvatar.cpp
daleglass: Yes, almost certainly. The problem is that I don't understand how things work here well enough.

My initial thought was that since `entity->die()` just sets a flag, this must be some sort of sanity check -- don't delete entities that are still present in the domain, or something along those lines. And if you create an entity in the interface, you need to mark it as dead yourself before deleting it.

The problem is that the full intention of the code isn't very clear. 

`die()`/`isDead()` are defined in `SpatiallyNestable.h`. and do nothing besides setting a boolean. Being dead gets checked at a few points: `isReadyToComputeShape`, `shouldBePhysical`, `OtherAvatar::shouldBeInPhysicsSimulation`. So that part seems pretty clear: something dead is something that's about to be cleaned up, so it's not physical and shouldn't be rendered. Considering there are threads involved, cleanup may not be immediate, and meanwhile we don't want people bumping into stuff that's gone.

This code was added back in 2015 in 6eb177091bd and apparently the change there was moving those functions from `AvatarData.h`. Then both asserts and several entity->die() were added to `EntitySimulation.h`. 

That's the point at which it gets confusing for me. The above seems to be just dealing with a delayed cleanup. But in that case just removing an entity outright isn't really a problem. So what is it that the assert and the `entity->die()` in `Entitysimulation.h` are trying to achieve?

HifiExperiments: there must have been a more recent change in the merge with master that is triggering this.  I wonder if it could be specific to local entities?  they should never be physical, so maybe they used to be handled differently.

daleglass: I'm looking into it, yes.

I'm just trying to figure out what's the intent here -- all this die/isDead business would seem to serve some purpose beyond simply getting stuff out of the way when it's scheduled for cleanup.

So far it looks to me like this flag has two purposes: getting removed stuff out of the way, and a second one that performs some sort of correctness checking.

daleglass: I think the crash after this PR is applied is interesting: The backtrace comes from QtScript, which directly calls `EntityScriptingInterface::deleteEntity`. Which I take means some script is deleting something outright, and there's no opportunity there for `entity->die()` to be called.

There's a change in commit 4159bc48 that looks like it could be relevant, I'm looking into it.

daleglass: Another promising looking commit: 9d92546a

daleglass: Okay, pushed an update. There was a single `theEntity->die()` removed in commit 9d92546. I added it back and that seems to fix things.

daleglass: By the way, for testing you need to run a server. This problems can happen on both sides.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-31f3da9_95570d5a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-31f3da9.tar_95570d5a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-31f3da9_95570d5a.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-31f3da9.tar_95570d5a.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR340/ProjectAthena-Alpha-PR340-31f3da9_95570d5a.exe

humbletim:
tested master debug build on Ubuntu 16.04 and crashed 100% of the time whenever clicking on one of the chairs in hub-master.  rebuilt w/Dale's patch and could not actually get interface to crash at all!

ctrlaltdavid:
The two commits mentioned relate to the following three merges to master:

```
11171 (November 14 2019 at 20:10 UTC)
From Pull Request: 16361
DEV-1811 redux: apply entity deletion rules 
This is a second attempt to apply the entity deletion rules: 

(1) deleting an avatar-entity parent doesn't delete any domain-entity children 
(2) deleting a domain-entity parent does delete all of its children 

https://highfidelity.atlassian.net/browse/DEV-1811 

The previous attempt was PR-16307.
```

```
11070 (October 11 2019 at 20:13 UTC)
From Pull Request: 16333
Revert "DEV-1811: apply deletion rules to linked entities" 
Reverts highfidelity/hifi#16307
```

```
11067 (October 10 2019 at 20:24 UTC)
From Pull Request: 16307
DEV-1811: apply deletion rules to linked entities 
This PR is progress toward "more correct" deletion rules on chains of linked entities. In short: children of entities being deleted should always also be deleted... except in cases where they shouldn't (e.g. where the deleter doesn't have the authority to delete). The internal ticket is DEV-1811: 

https://highfidelity.atlassian.net/browse/DEV-1811
```

daleglass:
> This indicates what testing should be done.

Very helpful, thanks!

daleglass:
> (1) deleting an avatar-entity parent doesn't delete any domain-entity children

Why? Like if I'm wearing a hat, shouldn't it disappear when I do?

daleglass:
Testing so far:

> (1) deleting an avatar-entity parent doesn't delete any domain-entity children 

1. Create box1
2. Create box2
3. Parent box2 to my uuid
4. Teleport away
5. Return

Result: box1 remains, box2 is not there.

> (2) deleting a domain-entity parent does delete all of its children 

1. Create box1
2. Create box2
3. Create box3
4. Set box1 as parent of box2 and box3
5. Verify that moving box1 moves both
6. Verify that deleting box1 deletes everything

ctrlaltdavid:
Some test comments from the original PRs...

Emoji stress testing performed
Pointing stress testing performed
Parent-child deletion matrix confirmed

Children are deleted when the parent is deleted both when lifetime expires and when manually deleted. This included manually deleting each of the four entities where I could (unable to select local entities in create, so just had to rely on the lifetime for those).
Tested the following combinations:
avatar-domain-avatar-domain
domain-avatar-domain-avatar
avatar-local-avatar-local
local-avatar-local-avatar
domain-local-domain-local
local-domain-local-domain

PR #343, opened 2020-05-12
Change log file name to "vircadia-.."

jug007:
Just changed the log file name to from "hifi-.." to "vircadia-..".
Already build an tried it on linux, so it should be fine as it is such a small change.


jug007:
I am going to be honest. I am mega new to git and have very little idea what I am doing, so I cannot actually get rid of that.
I guess it is the fact that I updated the branch with a pull request from master, which for some reason created a pull request.

PR #344, opened 2020-05-12
Disable "grabbable" property by default

### LINKED ISSUE #209: Suggestion:  Disable "Create Entities as Grabbable" by default.
This is an option in the edit menu that is checked ON by default.  I can't tell you how many times I've started to work on an elaborate build only to find out nearly all objects are grabbale after someone visits in VR and accidentally redecorates.

jug007:
Changed the code so that no entities are grabbable by default anymore.
As I cannot see why anyone would want this unless this option was easily toggleable, I completely removed that part of the code.
Tested if it works and doesn't break anything that might be related to it.
Closes #209

### DIFF
--- a/scripts/system/create/edit.js
+++ b/scripts/system/create/edit.js
@@ -1409,7 +1409,7 @@ function setupModelMenus() {
         menuItemName: MENU_CREATE_ENTITIES_GRABBABLE,
         afterItem: "Unparent Entity",
         isCheckable: true,
-        isChecked: Settings.getValue(SETTING_EDIT_PREFIX + MENU_CREATE_ENTITIES_GRABBABLE, true)
+        isChecked: Settings.getValue(SETTING_EDIT_PREFIX + MENU_CREATE_ENTITIES_GRABBABLE, false)
     });
 
     Menu.addMenuItem({


AleziaKurdis:
It wipes the setting usage?
Is it not a bit radical ? Or should we consider to remove the setting?

image

Or we remove the setting too
Or we keep the setting and have it by default to false

kasenvr:
I think what should be done is that it's a checkbox in the actual create app (we need settings in the actual UI) and then have this set it to be defaultly false. But you'll be one check away from getting the functionality you need. Most people do not know that checkbox is in the menu, menu items are used primarily to prototype features before you have a good place/functionality for them.

jug007:
"Most people do not know that checkbox is in the menu"
I set it to false first, but then though to myself "there is no checkbox to change this anyways, so might as well remove the whole thing".
Now that I know that there is in fact a setting, I will look into putting that back in.

AleziaKurdis:
That setting was a poor solution. After having thought about it longly, It would have been a good idea to get rid of it. It was to support a behavior the current behavior that practically every creator didn't want (grabbable by default) and offer them an alternative.
But the approach done to have the grabbable option when we create a "Model" Entity is definitely a better solution to this. Not sure we need this for every type of entity. It would probably worth doing it for "Shape" entities, using a create a dialog to select the shape. 
But one thing is clear, it would be preferable to have it "false" by default. 

If you could find out how to remove the setting I would vote that you keep your code change.
(of course, a check would need to be done to make sure the setting is only used for Create App.)
That was my 2 Cents.

jug007:
This now just changes the behavior when generating a new `Interface.json`, so the option will be disable by default.
The setting is still useful if people know that it is there, so it should not just be removed.

thoys:
How about having a settings page that is dedicated to the create app? This would allow you to set some defaults for creating entities.

jug007:
Yeah. Personally I have been thinking, maybe put the setting on the page where you create new entities on. I don't know if we have enough, or too little settings that would go there though.

Nex-Pro:
Toggling grab should by design be right at the point of creation. Desktop users will probably want grab off and HMD users will often want it on.

PR #346, opened 2020-05-13
Remove all uses of QString::null

daleglass:
It's deprecated and produces a warning on Qt 5.14. It produces lots of:

    'QString::null' is deprecated: use QString() [-Wdeprecated-declarations]

when compiling.

### DIFF
--- a/launchers/qt/src/LauncherState.h
+++ b/launchers/qt/src/LauncherState.h
@@ -176,7 +176,7 @@ private:
     QString _displayName;
     QString _applicationErrorMessage;
     QString _currentClientVersion;
-    QString _buildTag { QString::null };
+    QString _buildTag;
     QString _contentCacheURL;
     QString _loginTokenResponse;
     QFile _clientZipFile;
--- a/libraries/shared/src/PathUtils.h
+++ b/libraries/shared/src/PathUtils.h
@@ -58,7 +58,7 @@ public:
     static QString generateTemporaryDir();
     static bool deleteMyTemporaryDir(QString dirName);
 
-    static int removeTemporaryApplicationDirs(QString appName = QString::null);
+    static int removeTemporaryApplicationDirs(QString appName = QString());
 
     static Qt::CaseSensitivity getFSCaseSensitivity();
     static QString stripFilename(const QUrl& url);
--- a/launchers/qt/src/BuildsRequest.h
+++ b/launchers/qt/src/BuildsRequest.h
@@ -4,10 +4,10 @@
 #include <QNetworkAccessManager>
 
 struct Build {
-    QString tag{ QString::null };
+    QString tag;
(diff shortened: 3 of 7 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR346/ProjectAthena-Alpha-PR346-99c6e00_2631b983.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR346/ProjectAthena-Alpha-PR346-99c6e00_2631b983.exe

launchers/qt/src/LauncherState.h
```diff
@@ -176,7 +176,7 @@ private slots:
     QString _displayName;
     QString _applicationErrorMessage;
     QString _currentClientVersion;
-    QString _buildTag { QString::null };
+    QString _buildTag { "" };
```
ctrlaltdavid, line 179:
```suggestion
    QString _buildTag { QString() };
```
`QString()` creates a null string, and a null default is needed for program logic: `Builds::getBuild(...)` tests `.isNull()` which is false if `_buildTag` is initialized to `""`.

Alternatively, remove the initializer which achieves the same thing.

libraries/shared/src/PathUtils.h
```diff
@@ -58,7 +58,7 @@ class PathUtils : public QObject, public Dependency {
     static QString generateTemporaryDir();
     static bool deleteMyTemporaryDir(QString dirName);
 
-    static int removeTemporaryApplicationDirs(QString appName = QString::null);
+    static int removeTemporaryApplicationDirs(QString appName = "");
```
ctrlaltdavid, line 61:
```suggestion
    static int removeTemporaryApplicationDirs(QString appName = QString());
```
Implementation of `removeTemporaryApplicationDirs(…)` tests `.isNull()`.

daleglass: Gah. Good catch, fixing both.

Maybe we should just test for isEmpty instead?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR346/ProjectAthena-Alpha-PR346-66a1200_407b2a4f.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR346/ProjectAthena-Alpha-PR346-66a1200_407b2a4f.exe

libraries/shared/src/PathUtils.h
ctrlaltdavid: Possibly. Though would need to check program logic to ensure there aren't any unintended consequences. So probably easier just to keep using null values per current code.

PR #349, opened 2020-05-15
Fix warnings emitted by GCC

daleglass:
This fixes a whole bunch of build-time warnings emitted by GCC on Linux

* An incorrect fall-through (so an actual bug)
* Suppress fall-through warnings where a fall-through was intended
* Fix an "X will be initialized after Y" warning (possible bug resulting in a variable not being correctly initialized)
* Fix numerical precision warning
* Fix warnings about signed and unsigned comparison
* Fix a set of Wclass-memaccess warnings.

The last one needs a good looking at, because it touches rendering code. Some of this kind still remain, to be addressed later. I fixed the ones relating to glm:: data types, since GLM is an external library we can't modify. The rest belong to Vircadia code which can be changed.

### DIFF
--- a/libraries/octree/src/OctreePacketData.cpp
+++ b/libraries/octree/src/OctreePacketData.cpp
@@ -16,6 +16,7 @@
 
 #include "OctreeLogging.h"
 #include "NumericalConstants.h"
+#include <glm/gtc/type_ptr.hpp>
 
 bool OctreePacketData::_debug = false;
 AtomicUIntStat OctreePacketData::_totalBytesOfOctalCodes { 0 };
@@ -702,17 +703,17 @@ void OctreePacketData::debugBytes() {
 }
 
 int OctreePacketData::unpackDataFromBytes(const unsigned char* dataBytes, glm::vec2& result) {
-    memcpy(&result, dataBytes, sizeof(result));
+    memcpy(glm::value_ptr(result), dataBytes, sizeof(result));
     return sizeof(result);
 }
 
 int OctreePacketData::unpackDataFromBytes(const unsigned char* dataBytes, glm::vec3& result) {
-    memcpy(&result, dataBytes, sizeof(result));
+    memcpy(glm::value_ptr(result), dataBytes, sizeof(result));
     return sizeof(result);
 }
 
 int OctreePacketData::unpackDataFromBytes(const unsigned char* dataBytes, glm::u8vec3& result) {
-    memcpy(&result, dataBytes, sizeof(result));
+    memcpy(glm::value_ptr(result), dataBytes, sizeof(result));
     return sizeof(result);
 }
 
@@ -743,7 +744,12 @@ int OctreePacketData::unpackDataFromBytes(const unsigned char *dataBytes, QVecto
     memcpy(&length, dataBytes, sizeof(uint16_t));
     dataBytes += sizeof(length);
     result.resize(length);
-    memcpy(result.data(), dataBytes, length * sizeof(glm::vec3));
+
+    for(int i=0;i<length;i++) {
+        memcpy(glm::value_ptr(result[i]), dataBytes, sizeof(glm::vec3));
+        dataBytes += sizeof(glm::vec3);
+    }
+
     return sizeof(uint16_t) + length * sizeof(glm::vec3);
 }
 
--- a/libraries/shared/src/BufferParser.h
+++ b/libraries/shared/src/BufferParser.h
@@ -99,7 +99,7 @@ template<>
 inline void BufferParser::readValue(QVector<glm::vec3>& result) {
     uint16_t length; readValue(length);
     result.resize(length);
-    for(int i=0;i<length;i++) {
+    for (int i=0; i<length; i++) {
         memcpy(glm::value_ptr(result[i]), _data + _offset + (sizeof(glm::vec3)*i), sizeof(glm::vec3) * length);
     }
 
--- a/libraries/shared/src/BufferParser.h
+++ b/libraries/shared/src/BufferParser.h
@@ -18,6 +18,7 @@
 #include "GLMHelpers.h"
 #include "ByteCountCoding.h"
 #include "PropertyFlags.h"
+#include <glm/gtc/type_ptr.hpp>
 
 class BufferParser {
 public:
@@ -98,7 +99,10 @@ template<>
 inline void BufferParser::readValue(QVector<glm::vec3>& result) {
     uint16_t length; readValue(length);
     result.resize(length);
-    memcpy(result.data(), _data + _offset, sizeof(glm::vec3) * length);
+    for(int i=0;i<length;i++) {
+        memcpy(glm::value_ptr(result[i]), _data + _offset + (sizeof(glm::vec3)*i), sizeof(glm::vec3) * length);
+    }
+
     _offset += sizeof(glm::vec3) * length;
 }
 
--- a/libraries/shared/src/GeometryUtil.cpp
+++ b/libraries/shared/src/GeometryUtil.cpp
@@ -620,7 +620,9 @@ void PolygonClip::clipToScreen(const glm::vec2* inputVertexArray, int inLength, 
     glm::vec2* tempVertexArrayB = new glm::vec2[maxLength];
 
     // set up our temporary arrays
-    memcpy(tempVertexArrayA, inputVertexArray, sizeof(glm::vec2) * inLength);
+    for(int i=0;i<inLength;i++) {
+        tempVertexArrayA[i] = inputVertexArray[i];
+    }
 
     // Left edge
     LineSegment2 edge;
--- a/libraries/shared/src/GeometryUtil.cpp
+++ b/libraries/shared/src/GeometryUtil.cpp
@@ -620,7 +620,7 @@ void PolygonClip::clipToScreen(const glm::vec2* inputVertexArray, int inLength, 
     glm::vec2* tempVertexArrayB = new glm::vec2[maxLength];
 
     // set up our temporary arrays
-    for(int i=0;i<inLength;i++) {
+    for (int i=0; i<inLength; i++) {
         tempVertexArrayA[i] = inputVertexArray[i];
     }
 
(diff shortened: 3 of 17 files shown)


libraries/octree/src/OctreePacketData.cpp
```diff
@@ -743,7 +744,12 @@ int OctreePacketData::unpackDataFromBytes(const unsigned char *dataBytes, QVecto
     memcpy(&length, dataBytes, sizeof(uint16_t));
     dataBytes += sizeof(length);
     result.resize(length);
-    memcpy(result.data(), dataBytes, length * sizeof(glm::vec3));
+
+    for(int i=0;i<length;i++) {
+        memcpy(glm::value_ptr(result[i]), dataBytes, sizeof(glm::vec3));
+        dataBytes += sizeof(glm::vec3);
+    }
```
HifiExperiments, line 751: what was wrong with all of these places that did it all at once?  it’ll be faster that way

libraries/shared/src/avx2/BlendshapePacking_avx2.cpp
```diff
@@ -269,13 +269,13 @@ void packBlendshapeOffsets_AVX2(float (*unpacked)[9], uint32_t (*packed)[4], int
...
-            case 3: _mm_storeu_si128((__m128i*)packed[i+2], _mm256_castsi256_si128(v2));
-            case 2: _mm_storeu_si128((__m128i*)packed[i+1], _mm256_castsi256_si128(v1));
-            case 1: _mm_storeu_si128((__m128i*)packed[i+0], _mm256_castsi256_si128(v0));
+            case 7: _mm_storeu_si128((__m128i*)packed[i+6], _mm256_extractf128_si256(v2, 1)); /* fall-thru */
+            case 6: _mm_storeu_si128((__m128i*)packed[i+5], _mm256_extractf128_si256(v1, 1)); /* fall-thru */
+            case 5: _mm_storeu_si128((__m128i*)packed[i+4], _mm256_extractf128_si256(v0, 1)); /* fall-thru */
+            case 4: _mm_storeu_si128((__m128i*)packed[i+3], _mm256_castsi256_si128(v3)); /* fall-thru */
+            case 3: _mm_storeu_si128((__m128i*)packed[i+2], _mm256_castsi256_si128(v2)); /* fall-thru */
+            case 2: _mm_storeu_si128((__m128i*)packed[i+1], _mm256_castsi256_si128(v1)); /* fall-thru */
+            case 1: _mm_storeu_si128((__m128i*)packed[i+0], _mm256_castsi256_si128(v0)); /* fall-thru */
```
HifiExperiments, line 278: are we sure these are supposed to fall through?

daleglass: Pretty sure of it, it looks very much like a manually unwrapped loop, with this bit taking care of whatever wasn't evenly divisible by 8

libraries/octree/src/OctreePacketData.cpp
daleglass: Recent versions of GCC emit a lot of warnings like this (this also got added to LLVM):

```
/home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Space.cpp: In member function ‘uint32_t workload::Space::copyProxyValues(workload::Proxy*, uint32_t) const’:
/home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Space.cpp:126:63: warning: ‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘class workload::Proxy’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
  126 |     memcpy(proxies, _proxies.data(), numCopied * sizeof(Proxy));
      |                                                               ^
In file included from /home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Transaction.h:21,
                 from /home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Space.h:22,
                 from /home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Space.cpp:15:
/home/dale/git/vircadia/vircadia-master/libraries/workload/src/workload/Proxy.h:42:7: note: ‘class workload::Proxy’ declared here
   42 | class Proxy {
      |       ^~~~~
```

The issue is that you're only supposed to memcpy over an object that's [trivially copyable](https://en.cppreference.com/w/cpp/types/is_trivially_copyable). If it isn't, this is technically undefined behaviour.

As I understand it there are two main reasons for this:

1. If there's a copy constructor, this would bypass it. That would be obviously bad.
2. There might be something other than the expected data in the class that could get overwritten, such as the vtable.

Admittedly, this probably borders on paranoia. But on the other hand, compiler authors seem to find various creative ways to optimize code that can make invoking undefined behaviour result in very weird things happening, so it might not be that bad idea to steer well clear of it.

Here's some interesting reading on how compilers react to UB: https://c9x.me/compile/bib/ubc.pdf

HifiExperiments: interesting!  at least for structs, I would expect it to be fine and faster to do all at once

daleglass: Yes, I'm quite sure it would be faster. But, considering the compiler is telling me "writing over this with memcpy is undefined behavior", writing over an array seems like a bad idea. 

Here I added the use of glm::value_ptr seems to exist exactly for this purpose, pointing at where inside the glm::vec3 the data is found. This would seem to indicate that &vec3 may not be the same address as glm::value_ptr(vec3).

I of course welcome any better ideas here.

I'm also hoping that the compiler will optimize it back into what it was. The standard may say this is a bad idea, but the compiler knows what it generates and could well figure out it's actually safe in this particular case (I've not tried disassembling it and checking though).

daleglass:
Updated to apply against the current code. I'd like to make another push at this, because the attempt at upgrading Qt produced a whole bunch of new warnings, and that isn't good, because the more of that, the more likely something important gets lost in the noise. So I'd like to get rid of all warnings if at all possible.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-1783e54.tar_5a8be4de.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-1783e54.tar_5a8be4de.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-1783e54_5a8be4de.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-1783e54_5a8be4de.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-1783e54_5a8be4de.dmg

kasenvr:
Is there a test case for this? If not... pretty safe to merge or nah? :)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-8e96bb7_5a8be4de.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-8e96bb7.tar_5a8be4de.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-8e96bb7.tar_5a8be4de.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-8e96bb7_5a8be4de.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-8e96bb7_5a8be4de.exe

JulianGro:
I gave this a semi quick test and cannot find any issues.
For the change in libraries/render-utils/src/RenderPipelines.cpp : I feel like if no one knows that this even is, it cannot be *too* bad even if it does break something.
Of course I will let you know if I do find issues later.

HifiExperiments:
@JulianGro I know what it is because I wrote it and it’s definitely wrong in master lol

libraries/shared/src/BufferParser.h
```diff
@@ -98,7 +99,10 @@ template<>
 inline void BufferParser::readValue(QVector<glm::vec3>& result) {
     uint16_t length; readValue(length);
     result.resize(length);
-    memcpy(result.data(), _data + _offset, sizeof(glm::vec3) * length);
+    for(int i=0;i<length;i++) {
```
ctrlaltdavid, line 102:
```suggestion
    for (int i=0; i<length; i++) {
```

libraries/shared/src/GeometryUtil.cpp
```diff
@@ -620,7 +620,9 @@ void PolygonClip::clipToScreen(const glm::vec2* inputVertexArray, int inLength,
     glm::vec2* tempVertexArrayB = new glm::vec2[maxLength];
 
     // set up our temporary arrays
-    memcpy(tempVertexArrayA, inputVertexArray, sizeof(glm::vec2) * inLength);
+    for(int i=0;i<inLength;i++) {
```
ctrlaltdavid, line 623:
```suggestion
    for (int i=0; i<inLength; i++) {
```

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-a41d31e_fffb6619.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-a41d31e.tar_fffb6619.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-a41d31e.tar_fffb6619.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-a41d31e_fffb6619.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR349/Vircadia-Alpha-PR349-a41d31e_fffb6619.dmg

PR #355, opened 2020-05-17
additional log diagnostics for issue 165

odysseus654:
Adds a few diagnostic warnings in the log to try to help track down "no audio devices" in issue 165

### DIFF
--- a/interface/src/scripting/AudioDevices.cpp
+++ b/interface/src/scripting/AudioDevices.cpp
@@ -385,6 +385,7 @@ void AudioDeviceList::onDevicesChanged(QAudio::Mode mode, const QList<HifiAudioD
         }
     }
 
+    qWarning() << "AudioDeviceList::" << __FUNCTION__ << "(" << mode << ") replacing device list [len=" << rowCount() << "] with new list [len=" << newDevices.length() << "]";
     _devices.swap(newDevices);
     endResetModel();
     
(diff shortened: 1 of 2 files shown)


odysseus654:
Since auto-build doesn't work for K2, this is an unofficial (Windows 64bit) build from off of my workstation: https://odys-asset.s3.amazonaws.com/vircadia-pr355-unofficial.zip

daleglass:
I think a debug line in `AudioDeviceList::onDevicesChanged` could also be useful.

odysseus654:
I'm unsure what you're suggesting, this does add a debug line to AudioDeviceList::onDevicesChanged 

At the moment (based on the reported behavior) I'm actually expecting we'll discover AudioDeviceList::onDevicesChanged is not being called at all (and that debug line would be missing from the log file).  If that happens I'm considering this a race condition likely resolved by https://github.com/kasenvr/project-athena/commit/4f009d884f963a2949c80995ca28f2db668d60b9 (assumed to be already in master/Quantum).  If that turns out to be the case, I'm strongly tempted to cherrypick that commit and confirm that that resolves the issue (if it's ever possible to "confirm" that a race condition is resolved)

I'm not entirely sure what to do with this PR after this is done, I had created it originally to try to leverage the public-build for something interfacing with a non-developer (and likely wouldn't have added the extra load if I had realized it wouldn't work).  My immediate thought is to close the PR, but if some of these diagnostics may have value I could clean them up a bit and rebase them to master branch afterwards.

daleglass:
Er, oops. Yeah, I think that commit looks very promising.

PR #393, opened 2020-05-27
Bug: properly force creating temp domain name if the --get-temp-name

Misterblue:
I found that sometimes the --get-temp-name domain-server parameter did not work. Found that the forcing of temporary domain name creation relied on an unspecified parameter and thus was checking on uninitialized stack memory.
I can't believe a real programming language allows this.

### DIFF
--- a/domain-server/src/DomainServer.cpp
+++ b/domain-server/src/DomainServer.cpp
@@ -263,7 +263,7 @@ DomainServer::DomainServer(int argc, char* argv[]) :
 
     // check for the temporary name parameter
     if (_getTempName) {
-        getTemporaryName();
+        getTemporaryName(true);
     }
 
     // send signal to DomainMetadata when descriptors changed


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR393/ProjectAthena-Alpha-PR393-cdc143c_46f18d08.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR393/ProjectAthena-Alpha-PR393-cdc143c_46f18d08.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR393/ProjectAthena-Alpha-PR393-cdc143c.tar_46f18d08.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR393/ProjectAthena-Alpha-PR393-cdc143c_46f18d08.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR393/ProjectAthena-Alpha-PR393-cdc143c.tar_46f18d08.gz

thoys:
It seems looking at the function that it only doesn't work if there is no domain id set. Is this flag used elsewhere? If so I suggest making an additional --force-get-temp-name  .  since your claim of the undefined default value isn't true. See https://github.com/kasenvr/project-athena/blob/46f18d08a1a67855de4b16e88496d87a8faaf69e/domain-server/src/DomainServer.h#L166 . The default value is false which makes it only work if there is no previous domain id acquired.

Misterblue:
You are correct about the value defaulting "false". My oversight (and will have to find another reason why it seemed to act differently in different runs).

But I would argue for --get-temp-name to be the parameter one specifies to reset a domain-server's domain name. As Vircadia moves to a world with multiple MetaverseServer's, there will be the use case where a domain-server needs to reset its domain registration. At the moment, I believe the only way this can be done is through deleting some (unspecified) configuration files. Before this PR, the parameter is "get a temp domain name if you don't have an existing domain ID" which is what, I believe, a domain-server does anyway.

The conservative resolution is to add the new parameter as suggested or, I think more reasonably, make this existing parameter do what it says: cause the domain-server to get a temp domain name.

thoys:
That is why I asked if it is used for other things than forcing a new domain id. If it isn't then it is all good. But running the domain server with this parameter twice will cause it to get another temp name, which is why I am asking.

Misterblue:
I do not know of another use. It can be dangerous for a running domain (this also gets a new domainID which would undo mappings to place names, etc). I think that, with the previous, single HiFi domain registration server, the existing code tried to not reset registrations.

So, other than changing the parameter to make it scarier, I believe that, in our new use case, this parameter's operation should forceably do what it says: get a new temp domain name.

vegaslon:
The server console uses this parameter every time it starts a domain.

Misterblue:
If server-console indeed uses --get-temp-name on every invocation, then this PR cannot be accepted as it would unregister already registered domains.

I will look into adding a different parameter and/or other solutions.

PR #400, opened 2020-05-28
Bug: ice-server uses the wrong RSA private key decoder.

Misterblue:
The decoder must match the encoder used in libraries/networking/src/AccountManagement.cpp.

Found while developing a new MetaverseServer. A new domain-server generates a domain public/private key pair and sends it to the MetaverseServer. When the domain server later talks to the ice-server, the ice-server gets the domain's public key from the MetaverseServer to verify the identity of the connection.

Discovered the problem that the ice-server was not able to decode the public key it received from the MetaverseServer. Eventually discovered that the key decoding method used by ice-server did not match the encode method used by domain-server's AccountManagement.

This simple fix means that domain-server and ice-server can exchange signed heartbeats.

### DIFF
--- a/ice-server/src/IceServer.cpp
+++ b/ice-server/src/IceServer.cpp
@@ -252,7 +252,7 @@ void IceServer::publicKeyReplyFinished(QNetworkReply* reply) {
                 // convert the downloaded public key to an RSA struct, if possible
                 const unsigned char* publicKeyData = reinterpret_cast<const unsigned char*>(apiPublicKey.constData());
 
-                RSA* rsaPublicKey = d2i_RSA_PUBKEY(NULL, &publicKeyData, apiPublicKey.size());
+                RSA* rsaPublicKey = d2i_RSAPublicKey(NULL, &publicKeyData, apiPublicKey.size());
 
                 if (rsaPublicKey) {
                     _domainPublicKeys[domainID] = { rsaPublicKey, RSA_free };


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR400/ProjectAthena-Alpha-PR400-fc2f991_9e322828.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR400/ProjectAthena-Alpha-PR400-fc2f991.tar_9e322828.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR400/ProjectAthena-Alpha-PR400-fc2f991_9e322828.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR400/ProjectAthena-Alpha-PR400-fc2f991.tar_9e322828.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR400/ProjectAthena-Alpha-PR400-fc2f991_9e322828.exe

vegaslon:
These test builds I am pretty sure are not packaging up the ice server, at least in the case of windows. Lucky the ice server is small. Some one will have to compile it separate and put it in a zip file to test

Misterblue:
continued testing has exposed the fact that the same public key coding problem exists with user public keys: RSAKeypairGenerator.css uses i2d_RSAPublicKey() to create a public key, but DomainGatekeeper.cpp uses d2i_RSA_PUBKEY() to decode it.

I believe this means that the HiFi metaverse-server must be changing the format of the public key -- it receives PKCS formatted public keys but sends out SubjectPublicKeyInfo formatted ones.

At the moment, I haven't found a C# library that can do that conversion. One thought is to fix domain-server so it sends both formats to the metaverse-server. This can be done in a compatible way since the request is already a multi-part form of binary data.

It is starting to look like this PR is the wrong fix for the "problem".

Misterblue:
After changing the Project Apollo sources and testing, it is clear that the real solution is the metaverse-server changing the public-key format. Thus, this PR should not be applied.

PR #410, opened 2020-06-02
Update packaging parameters for Vircadia

kasenvr:
This PR focuses on updating the naming and branding for the packaging and install steps. It also changes where the data is stored in AppData to a "Vircadia" folder.


ctrlaltdavid:
PR build fails because the GitHub Actions scripts need to be updated per the changed file names. For example, to use "Vircadia-Alpha" instead of "ProjectAthena-Alpha".

kasenvr:
I've tested it on Windows, I installed and ran it, checked where the log files, interface settings, and the installation is and it all appears to be in the right place.

It was necessary for me to do this and compile the RC with it even without this being merged so that the change is consistent from here on out.

https://cdn.vircadia.com/dist/release-candidate/Vircadia-Alpha-2020.2.0-Asteria-RC1-INTERNAL1.exe

You can test it using that, though that build was generated without some of the latest PRs, so please take note that the version may remain the same when an updated version is released if you keep this installed.

kasenvr:
Myself and Vegaslon have logged on using a build with this PR, files and folders seem to be correct. :)

PR #417, opened 2020-06-04
Reinstate Instancing

kasenvr:
> This reverts commit 86dbf3f3f957c7453e85f9583bb86b41bdc26f1d, reversing
> changes made to d69c89583f334ef3acdf02e5efc4e0454ad4ef0f. 

This was removed in PR#414 kasenvr/fix/revert-instancing-temporarily.

https://github.com/kasenvr/project-athena/issues/412
https://github.com/kasenvr/project-athena/issues/383

The above issues outlines the current problem with this PR. This PR will be merged to reinstate instancing once this bug is solved.

### DIFF
--- a/libraries/fbx/src/FBXSerializer.cpp
+++ b/libraries/fbx/src/FBXSerializer.cpp
@@ -20,6 +20,7 @@
 #include <BlendshapeConstants.h>
 
 #include <hfm/ModelFormatLogging.h>
+#include <hfm/HFMModelMath.h>
 
 // TOOL: Uncomment the following line to enable the filtering of all the unkwnon fields of a node so we can break point easily while loading a model with problems...
 //#define DEBUG_FBXSERIALIZER
@@ -145,8 +146,9 @@ public:
     bool isLimbNode;  // is this FBXModel transform is a "LimbNode" i.e. a joint
 };
 
+
 glm::mat4 getGlobalTransform(const QMultiMap<QString, QString>& _connectionParentMap,
-        const QHash<QString, FBXModel>& fbxModels, QString nodeID, bool mixamoHack, const QString& url) {
+    const QHash<QString, FBXModel>& fbxModels, QString nodeID, bool mixamoHack, const QString& url) {
     glm::mat4 globalTransform;
     QVector<QString> visitedNodes; // Used to prevent following a cycle
     while (!nodeID.isNull()) {
@@ -166,12 +168,11 @@ glm::mat4 getGlobalTransform(const QMultiMap<QString, QString>& _connectionParen
         }
         QList<QString> parentIDs = _connectionParentMap.values(nodeID);
         nodeID = QString();
-        foreach (const QString& parentID, parentIDs) {
+        foreach(const QString& parentID, parentIDs) {
             if (visitedNodes.contains(parentID)) {
                 qCWarning(modelformat) << "Ignoring loop detected in FBX connection map for" << url;
                 continue;
             }
-
             if (fbxModels.contains(parentID)) {
                 nodeID = parentID;
                 break;
@@ -179,6 +180,21 @@ glm::mat4 getGlobalTransform(const QMultiMap<QString, QString>& _connectionParen
         }
     }
     return globalTransform;
+}
+
+std::vector<QString> getModelIDsForMeshID(const QString& meshID, const QHash<QString, FBXModel>& fbxModels, const QMultiMap<QString, QString>& _connectionParentMap) {
+    std::vector<QString> modelsForMesh;
+    if (fbxModels.contains(meshID)) {
+        modelsForMesh.push_back(meshID);
+    } else {
+        // This mesh may have more than one parent model, with different material and transform, representing a different instance of the mesh
+        for (const auto& parentID : _connectionParentMap.values(meshID)) {
+            if (fbxModels.contains(parentID)) {
+                modelsForMesh.push_back(parentID);
+            }
+        }
+    }
+    return modelsForMesh;
 }
 
 class ExtractedBlendshape {
@@ -404,7 +420,7 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
     QVector<ExtractedBlendshape> blendshapes;
 
     QHash<QString, FBXModel> fbxModels;
-    QHash<QString, Cluster> clusters; 
+    QHash<QString, Cluster> fbxClusters; 
     QHash<QString, AnimationCurve> animationCurves;
 
     QHash<QString, QString> typeFlags;
@@ -515,8 +531,8 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
                     if (object.properties.at(2) == "Mesh") {
                         meshes.insert(getID(object.properties), extractMesh(object, meshIndex, deduplicateIndices));
                     } else { // object.properties.at(2) == "Shape"
-                        ExtractedBlendshape extracted = { getID(object.properties), extractBlendshape(object) };
-                        blendshapes.append(extracted);
+                        ExtractedBlendshape blendshape = { getID(object.properties), extractBlendshape(object) };
+                        blendshapes.append(blendshape);
                     }
                 } else if (object.name == "Model") {
                     QString name = getModelName(object.properties);
@@ -690,8 +706,8 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
 
                     // add the blendshapes included in the model, if any
                     if (mesh) {
-                        foreach (const ExtractedBlendshape& extracted, blendshapes) {
-                            addBlendshapes(extracted, blendshapeIndices.values(extracted.id.toLatin1()), *mesh);
+                        foreach (const ExtractedBlendshape& blendshape, blendshapes) {
+                            addBlendshapes(blendshape, blendshapeIndices.values(blendshape.id.toLatin1()), *mesh);
                         }
                     }
 
@@ -1058,9 +1074,9 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
                             }
                         }
 
-                        // skip empty clusters
+                        // skip empty fbxClusters
                         if (cluster.indices.size() > 0 && cluster.weights.size() > 0) {
-                            clusters.insert(getID(object.properties), cluster);
+                            fbxClusters.insert(getID(object.properties), cluster);
                         }
 
                     } else if (object.properties.last() == "BlendShapeChannel") {
@@ -1214,11 +1230,11 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
     }
 
     // assign the blendshapes to their corresponding meshes
-    foreach (const ExtractedBlendshape& extracted, blendshapes) {
-        QString blendshapeChannelID = _connectionParentMap.value(extracted.id);
+    foreach (const ExtractedBlendshape& blendshape, blendshapes) {
+        QString blendshapeChannelID = _connectionParentMap.value(blendshape.id);
         QString blendshapeID = _connectionParentMap.value(blendshapeChannelID);
         QString meshID = _connectionParentMap.value(blendshapeID);
-        addBlendshapes(extracted, blendshapeChannelIndices.values(blendshapeChannelID), meshes[meshID]);
+        addBlendshapes(blendshape, blendshapeChannelIndices.values(blendshapeChannelID), meshes[meshID]);
     }
 
     // get offset transform from mapping
@@ -1233,13 +1249,13 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
     QVector<QString> modelIDs;
     QSet<QString> remainingFBXModels;
     for (QHash<QString, FBXModel>::const_iterator fbxModel = fbxModels.constBegin(); fbxModel != fbxModels.constEnd(); fbxModel++) {
-        // models with clusters must be parented to the cluster top
+        // models with fbxClusters must be parented to the cluster top
         // Unless the model is a root node.
         bool isARootNode = !modelIDs.contains(_connectionParentMap.value(fbxModel.key()));
         if (!isARootNode) {  
             foreach(const QString& deformerID, _connectionChildMap.values(fbxModel.key())) {
                 foreach(const QString& clusterID, _connectionChildMap.values(deformerID)) {
-                    if (!clusters.contains(clusterID)) {
+                    if (!fbxClusters.contains(clusterID)) {
                         continue;
                     }
                     QString topID = getTopModelID(_connectionParentMap, fbxModels, _connectionChildMap.value(clusterID), url);
@@ -1283,12 +1299,18 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
 
     // convert the models to joints
     hfmModel.hasSkeletonJoints = false;
+    
+    bool needMixamoHack = hfmModel.applicationName == "mixamo.com";
 
-    foreach (const QString& modelID, modelIDs) {
+    std::vector<glm::mat4> transformForClusters;
+    transformForClusters.reserve((size_t)modelIDs.size());
+    for (const QString& modelID : modelIDs) {
         const FBXModel& fbxModel = fbxModels[modelID];
         HFMJoint joint;
         joint.parentIndex = fbxModel.parentIndex;
-        int jointIndex = hfmModel.joints.size();
+        uint32_t jointIndex = (uint32_t)hfmModel.joints.size();
+
+        // Copy default joint parameters from model
 
         joint.translation = fbxModel.translation; // these are usually in centimeters
         joint.preTransform = fbxModel.preTransform;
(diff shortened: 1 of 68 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/ProjectAthena-Alpha-PR417-ae0b40b_1850c184.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/ProjectAthena-Alpha-PR417-ae0b40b.tar_1850c184.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/ProjectAthena-Alpha-PR417-ae0b40b_1850c184.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/ProjectAthena-Alpha-PR417-ae0b40b.tar_1850c184.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/ProjectAthena-Alpha-PR417-ae0b40b_1850c184.exe

Aitolda:
Not sure I have any answers yet, but after doing some testing with the Pillar above, it appears to import visually at the right size in scale, however in the create tools it's showing as exactly twice it's scale on the x and z axis within interface, while the Y axis (up/down) is acurate to scaled size in blender (which is the Z axis within blender). HOWEVER... in blender this axis is actually the ONLY part that is scaled. So... the non-scaled axis are showing up with incorrect dimensions, while the scaled dimension shows up correctly.

In Blender note the scaling and length on the Z axis (Y within interface)
Blender

And note the Y-dimensions within the create menu in interface (again)
https://i.gyazo.com/e58fd49e0bda9b00b9d735a163da0035.png

Both remain at 17.8, but the other two dimensions are doubled in interface. Again, they actually are showing up at the correct visual scale for me, just not in the create dimensions.

kasenvr:
Vegaslon noted these two commits may be of interest:
https://github.com/yozlet/interface/commit/ecdca05679349f9b3add4ed9f1aa4d2f5527cd5c
https://github.com/yozlet/interface/commit/a70f0f27918e315cb295bb2171dcce22445f875c

libraries/fbx/src/FBXSerializer.cpp
```diff
@@ -1299,35 +1321,64 @@ HFMModel* FBXSerializer::extractHFMModel(const hifi::VariantHash& mapping, const
...
+
         glm::quat combinedRotation = joint.preRotation * joint.rotation * joint.postRotation;
+        glm::mat4 localTransform = glm::translate(joint.translation) * joint.preTransform * glm::mat4_cast(combinedRotation) * joint.postTransform;
+
         if (joint.parentIndex == -1) {
-            joint.transform = hfmModel.offset * glm::translate(joint.translation) * joint.preTransform *
-                glm::mat4_cast(combinedRotation) * joint.postTransform;
+            joint.transform = localTransform * joint.geometricOffset;
+            joint.globalTransform = hfmModel.offset * localTransform * joint.geometricOffset;
+            joint.globalTransformForChildren = hfmModel.offset * localTransform;
```
HifiExperiments, line 1350: this is basically the fix...when hifi did the instancing PR, they removed support for geometricOffset

I don't know if there's an analogous property in GLTF so for now this (globalTransformForChildren) is only used here so that each joint has the correct globalTransform (including geometricOffset), but it doesn't pass geometricOffset down to its children

libraries/render-utils/src/Model.cpp
```diff
@@ -232,48 +232,46 @@ void Model::updateRenderItems() {
...
+                    if (useDualQuaternionSkinning) {
+                        data.updateClusterBuffer(meshState.clusterDualQuaternions);
+                    } else {
+                        data.updateClusterBuffer(meshState.clusterMatrices);
                     }
-                }
-                data.updateTransformForSkinnedMesh(renderTransform, modelTransform);
 
-                data.setCauterized(cauterized);
-                data.setRenderWithZones(renderWithZones);
```
HifiExperiments, line 273: I had two merge conflicts when trying to update this PR to master, which reviewers should pay particular attention to.  one was around this area, I think because of the renderWithZones stuff, which I have tried to preserve

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -979,657 +1073,501 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
+                continue;
+            }
 
-                int indicesAccessorIdx = primitive.indices;
+            // Increment the triangle indices by the current mesh vertex count so each mesh part can all reference the same buffers within the mesh
+            int prevMeshVerticesCount = mesh.vertices.count();
+            QVector<uint16_t> clusterJoints;
+            QVector<float> clusterWeights;
 
-                if (indicesAccessorIdx > _file.accessors.size()) {
```
HifiExperiments, line 1058: another was in this file, I think because we added these bound checks

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-a248f48.tar_7bb95a55.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-a248f48.tar_7bb95a55.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-a248f48_7bb95a55.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-a248f48_7bb95a55.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-a248f48_7bb95a55.exe

kasenvr:
While it will be good to review this fully, really all that needs to be considered is what HiFiExperiments added... the rest is stuff High Fidelity kinda added semi-finished but untested fully (how we caught this problem, but maybe they already knew about it and were intending to revisit it, according to the comments, in which case then it was fully considered and of course code-reviewed.) 

I know we're dealing with limited resources, so it's my hope that we'll catch more issues in testing. :)

Aitolda:
Tested.  Scale and position seems to be working as expected. I did notice that drawcalls don't seem to report accurately, however, after switching back to stable this does not seem to be specific to this PR.

vegaslon:
Understand why they removed model offset from this pr. If you are trying to instance a model having geometry offsets on it will cause all but one of the instanced models to be in the wrong place. We will just have to work around that when the times comes for someone try instancing in a scene.

Aitolda:
> Understand why they removed model offset from this pr. If you are trying to instance a model having geometry offsets on it will cause all but one of the instanced models to be in the wrong place. We will just have to work around that when the times comes for someone try instancing in a scene.

You are still having this problem?  I did not. I wonder what we may be doing differently.

vegaslon:
No do not have “this” problem as currently stated. I am just saying that may see “new bugs” in future.

HifiExperiments:
@vegaslon the old code (before the original PR removed it) worked by actually altering the underlying vertices of the meshes, which I assume would lead to the problem you mention.  but with my new changes, the vertices stay unmodified, but the applied transform changes.  I *think* this should fix the issue, but the only way to be sure would be to create a model with instanced parts and geometric offsets and test it

AleziaKurdis:
image

Still have the problem with this PR417 build.

image


image

AleziaKurdis:
There are links on those model in 
https://github.com/kasenvr/project-athena/issues/383
if you need this to investigate.

HifiExperiments:
@daleglass almost all of this code was written by sabrina and was already in hifi master.  only the commits by me, specifically cda77eb388cb8bbd084d88779c627e994b1a55c1, are the actual changes to fix this bug.  I can't really speak to the other changes as I don't understand a lot of them...but theoretically they shouldn't need to be reviewed, since they were already in master before we rolled them back.

except of course that my changes didn't seem to fix the problem, so I need to keep digging

daleglass:
@HifiExperiments Fair enough, but if it's not working right there probably is something wrong somewhere, so I think it's still worthwhile to take a look at the whole thing and see if anything looks suspicious.

daleglass:
In particular it would be helpful if somebody more familiar with this area could say what implications the "TODO" bits have. What are we missing there? An optimization? Are we regressing in functionality?

ctrlaltdavid:
The HiFi merge was late 2019 when they were already winding down development and perhaps being less rigorous in reviewing and testing before merging.

kasenvr:
This PR needs some conflicts to be resolved. :)

libraries/fbx/src/FBXSerializer.cpp
ctrlaltdavid: The top level object in glTF is the "scene". The scene has one or more "nodes". Each node may have a transform plus one or more meshes, and it can have other nodes as children (e.g., to form a skeleton).

kasenvr:
@Aitolda will revisit this in the next week or two to test more thoroughly.

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -979,668 +1074,504 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
+        if (!gltfMesh.primitives.empty()) {
+            for (const auto& attribute : gltfMesh.primitives[0].attributes.values.keys()) {
+                primitiveAttributes.insert(attribute);
+            }
+        }
+        std::vector<QSet<QString>> primitiveAttributeVariants;
 
-            hfmModel.meshes.append(HFMMesh());
-            HFMMesh& mesh = hfmModel.meshes[hfmModel.meshes.size() - 1];
-            mesh.modelTransform = globalTransforms[nodeIndex];
```
HifiExperiments, line 1022: this line was added here: https://github.com/kasenvr/project-athena/pull/699

need to confirm if it is still needed to fix picking and if so how to integrate it here because `globalTransforms` is gone now

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -979,668 +1074,504 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
+            if (!primitive.targets.isEmpty()) {
 
-                    for (int i = 0;; ++i) {
-                        auto blendshapeName = QString(BLENDSHAPE_NAMES[i]);
-                        if (blendshapeName.isEmpty()) {
-                            break;
-                        }
-                        auto mappings = blendshapeMappings.values(blendshapeName);
-                        if (mappings.count() > 0) {
+                // Build list of blendshapes from FST and model.
```
HifiExperiments, line 1462: this blendshape stuff was another conflict but I tried to maintain the code in master from here: https://github.com/kasenvr/project-athena/pull/671

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -979,668 +1074,504 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
+            continue;
+        }
 
-            // Mesh extents must be at least a minimum size, in particular for blendshapes to work on planar meshes.
-            const float MODEL_MIN_DIMENSION = 0.001f;
-            auto delta = glm::max(glm::vec3(MODEL_MIN_DIMENSION) - mesh.meshExtents.size(), glm::vec3(0.0f)) / 2.0f;
-            mesh.meshExtents.minimum -= delta;
-            mesh.meshExtents.maximum += delta;
-            hfmModel.meshExtents.minimum -= delta;
-            hfmModel.meshExtents.maximum += delta;
```
HifiExperiments, line 1639: this sizing code was changed here: https://github.com/kasenvr/project-athena/pull/671

but removed in this original PR?  I think it has moved to thickenFlatExtents in HFMModelMath but now maybe it's lost david's changes?

HifiExperiments (commented):
I've tried to resolve the conflicts here but still don't really understand any of this serializer code.

at the end of the day we need to make sure these two fixes aren't lost: https://github.com/kasenvr/project-athena/pull/699 https://github.com/kasenvr/project-athena/pull/671

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-d46607a_21259aa3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-d46607a_21259aa3.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-d46607a.tar_21259aa3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-d46607a.tar_21259aa3.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-d46607a_21259aa3.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3_21259aa3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3.tar_21259aa3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3.tar_21259aa3.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3_21259aa3.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3_21259aa3.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Quest-Alpha-PR417-c6ad5e3_21259aa3.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR417/Vircadia-Alpha-PR417-c6ad5e3_21259aa3.apk

Aitolda:
The offset issue on instances fbx's appears to be fixed now.  In the models below the origin point of the boxes is one of the bottom corners, so any incorrect rotations would be apparent.
[Blender](https://gyazo.com/58d0316418add22e28d1a222a54761fe)
[Vircadia](https://i.gyazo.com/15b9e5bff9e98fa35dcc5d4fe0df6c8b.png)

It is worth noting that one particular older non-instanced gltf that just happened to be in the same domain was flipped upside down while in the PR, however importing the same gltf into a more modern version of blender and re-exporting resulted in the same gltf model appearing in the correct orientation, leading me to believe the issue is with the older blender export, and not necessarily this PR,  I also duplicated the same issue in another metaverse with instancing.

The takeaway is some older gltf models exported from blender may not appear correctly, but it seems to be rare.  On the other hand, instanced fbx's seem fixed in this PR.

Aitolda:
While I need to investigate further, older models that are scaling incorrectly when imported back into current blender 2.9x, resaved without changes, seam to stop having the issue.

AleziaKurdis:
> While I need to investigate further, older models that are scaling incorrectly when imported back into current blender 2.9x, resaved without changes, seam to stop having the issue.

I suspected that. If this is the case, it might be a lot of rework, but I think we should go ahead.

digisomni:
We should investigate what the Hub looks like in this PR for example.

digisomni:
Per @ctrlaltdavid: Can we find out the difference between the FBX's and insert a compatibility fix then to check to see if a model was exported with the older version and process that differently?

ctrlaltdavid:
If we can get some example "broken before", "not broken before" and "broken before fixed after" models, it might be possible to see what differences there are in the FBX files.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR417/Vircadia-Alpha-PR417-0bc4d1f_21259aa3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR417/Vircadia-Alpha-PR417-0bc4d1f.tar_21259aa3.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR417/Vircadia-Alpha-PR417-0bc4d1f.tar_21259aa3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR417/Vircadia-Alpha-PR417-0bc4d1f_21259aa3.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR417/Vircadia-Alpha-PR417-0bc4d1f_21259aa3.exe

AleziaKurdis:
Maybe this could help:

**BROKEN:**
http://metaverse.bashora.com/objects/ZEDHA/DOORFRAME.fbx
(16 May 2017)
 
http://metaverse.bashora.com/objects/ZEDHA/FENETRE.fbx
(8 May 2017)

http://metaverse.bashora.com/objects/WOTHAL/PILAR.fbx
(2018)

http://metaverse.bashora.com/objects/WOTHAL/PILAR_RE_EDITED_BL281.fbx
(Re-edited in Blender 2.81 (15 dec 2020), Materials get modified)
but with this version it has the same issue!

**NOT BROKEN:**
http://metaverse.bashora.com/objects/ZEDHA/Cafe.fbx 
(15 May 2017)

http://metaverse.bashora.com/objects/WOTHAL/PILAR_RE_EDITED_BL281_WITH_APPLY.fbx
(Re-edited in Blender 2.81 (15 dec 2020), Materials get modified, **AND I did Apply Location, Scale , Rotation**)

--------------------------
I clearly suspect that it has to do with the "**Apply: Location, Scale , Rotation**"
A concept that I wasn't even know 18 months ago... 

so depending if some objects has been rotate or scale in "**Object Mode**" in **Blender**, without have done "Apply"
these objects has the problem. (in the PR417 version)
But sometime, I was starting  to work directly in "Edit Mode" in blender  without have to play with it in "Object Mode",
then those model would have resisted.
Of course, all those model are render correctly in the release 2020.3.2 and are display broken in PR417.

AleziaKurdis:
Here, the "PILAX.blend" file with the scale (Z-UP)
image
In-word, the PILAR.fbx model is rescaled inside its bounding box with opposite proportions.
Those that had the scale of 1,1,1 are untouched. (As observed, being the opposite of them-self)

AleziaKurdis:
This is not a old model or old software issue, but clearly a bug.
I can now reproduce it with model created today with Blender 2.8.1

**2020.3.2 - Demeter**
ROTATE_ONLY_2020_3_2

**PR 417**
ROTATE_ONLY_PR417

**Here's the step to reproduce in Blender** (probably any version)

1- Open Blender

2- Clear the scene and add a mesh of type cube

3- in "**Object Mode**", select the cube and do **R, Z, 45**
(This rotate the Object of 45 degree on the Z axis (blender being Z-UP))

4- Switch to "**Edit Mode**"

5- Click **A** to select All

6- Do **R, Z, -45**
(This rotate the Object of -45 degree on the Z axis.)

7- you can add a material and save this.

8- Export as an FBX (you can use the Metaverse Plugin (Menithal), should not be the problem)

9- Import In-world...
**Results:**
**With 2020.3.2**, you should get a perfect cube that fits its bounding box when selected.

**With PR417**, you should get a long horizontal rectangular prism (Y axis being the longest side) and not fitting the bounding box that is a perfect cube.

**Additional tests**
The second test you see on the pictures (the blue cube)
is when you rotate only it in Object Mode. As you can see, the rotated model is rendered accordingly in it's bounding box, which make sense.

I made also additional test with the scale not being applied, and it's not that, the unique responsible is that double rotation that is badly computed at some point.

Here's the 3 test models:
http://metaverse.bashora.com/objects/broken417/BROKEN_CUBE_ROTATE_OBJECT_ONLY.fbx
http://metaverse.bashora.com/objects/broken417/BROKEN_CUBE_ROTATE_ONLY.fbx
http://metaverse.bashora.com/objects/broken417/GOOD_CUBE.fbx

AleziaKurdis:
Here How it looks inside the FBX of these 3 models:
image

image

image


image

PR #435, opened 2020-06-09
Possible fix for entity server crash

HifiExperiments:
https://github.com/kasenvr/project-athena/issues/434

needs testing on server and client

### DIFF
--- a/assignment-client/src/entities/EntityServer.cpp
+++ b/assignment-client/src/entities/EntityServer.cpp
@@ -376,9 +376,7 @@ void EntityServer::nodeAdded(SharedNodePointer node) {
 
 void EntityServer::nodeKilled(SharedNodePointer node) {
     EntityTreePointer tree = std::static_pointer_cast<EntityTree>(_tree);
-    tree->withWriteLock([&] {
-        tree->deleteDescendantsOfAvatar(node->getUUID());
-    });
+    tree->deleteDescendantsOfAvatar(node->getUUID());
     tree->forgetAvatarID(node->getUUID());
     OctreeServer::nodeKilled(node);
 }
--- a/assignment-client/src/entities/EntityServer.cpp
+++ b/assignment-client/src/entities/EntityServer.cpp
@@ -370,14 +370,18 @@ void EntityServer::entityFilterAdded(EntityItemID id, bool success) {
 
 void EntityServer::nodeAdded(SharedNodePointer node) {
     EntityTreePointer tree = std::static_pointer_cast<EntityTree>(_tree);
-    tree->knowAvatarID(node->getUUID());
+    if (tree) {
+        tree->knowAvatarID(node->getUUID());
+    }
     OctreeServer::nodeAdded(node);
 }
 
 void EntityServer::nodeKilled(SharedNodePointer node) {
     EntityTreePointer tree = std::static_pointer_cast<EntityTree>(_tree);
-    tree->deleteDescendantsOfAvatar(node->getUUID());
-    tree->forgetAvatarID(node->getUUID());
+    if (tree) {
+        tree->deleteDescendantsOfAvatar(node->getUUID());
+        tree->forgetAvatarID(node->getUUID());
+    }
     OctreeServer::nodeKilled(node);
 }
 
--- a/libraries/entities/src/EntityTree.cpp
+++ b/libraries/entities/src/EntityTree.cpp
@@ -2212,16 +2212,24 @@ void EntityTree::fixupNeedsParentFixups() {
             }
 
             entity->postParentFixup();
-        } else if (getIsServer() || _avatarIDs.contains(entity->getParentID())) {
-            std::lock_guard<std::mutex> lock(_childrenOfAvatarsLock);
-            // this is a child of an avatar, which the entity server will never have
-            // a SpatiallyNestable object for.  Add it to a list for cleanup when the avatar leaves.
-            if (!_childrenOfAvatars.contains(entity->getParentID())) {
-                _childrenOfAvatars[entity->getParentID()] = QSet<EntityItemID>();
+        } else {
+            bool needsUpdate = getIsServer();
+            if (!needsUpdate) {
+                std::lock_guard<std::mutex> lock(_avatarIDsLock);
+                needsUpdate = _avatarIDs.contains(entity->getParentID());
             }
-            _childrenOfAvatars[entity->getParentID()] += entity->getEntityItemID();
-            doMove = true;
-            iter.remove(); // and pull it out of the list
+
+            if (needsUpdate) {
+                std::lock_guard<std::mutex> lock(_childrenOfAvatarsLock);
+                // this is a child of an avatar, which the entity server will never have
+                // a SpatiallyNestable object for.  Add it to a list for cleanup when the avatar leaves.
+                if (!_childrenOfAvatars.contains(entity->getParentID())) {
+                    _childrenOfAvatars[entity->getParentID()] = QSet<EntityItemID>();
+                }
+                _childrenOfAvatars[entity->getParentID()] += entity->getEntityItemID();
+                doMove = true;
+                iter.remove(); // and pull it out of the list
+            }
         }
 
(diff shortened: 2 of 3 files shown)


assignment-client/src/entities/EntityServer.cpp
```diff
@@ -376,9 +376,7 @@ void EntityServer::nodeAdded(SharedNodePointer node) {
 
 void EntityServer::nodeKilled(SharedNodePointer node) {
     EntityTreePointer tree = std::static_pointer_cast<EntityTree>(_tree);
-    tree->withWriteLock([&] {
```
HifiExperiments, line 379: I don't think this lock is necessary.  deleteDescendantsOfAvatar sometimes calls deleteEntitiesByID which already locks the tree

libraries/entities/src/EntityTree.cpp
```diff
@@ -2213,6 +2213,7 @@ void EntityTree::fixupNeedsParentFixups() {
 
             entity->postParentFixup();
         } else if (getIsServer() || _avatarIDs.contains(entity->getParentID())) {
+            std::lock_guard<std::mutex> lock(_childrenOfAvatarsLock);
```
HifiExperiments, line 2216: I *think* the problem is that _childrenOfAvatars was modified by different threads, so this just adds a lock in the few places it's used to prevent that

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.sh

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.sh

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f.tar_ae8879a5.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-7bc093f_ae8879a5.exe

daleglass:
I'm told it doesn't fix the problem. Also, the crash happens consistently, which doesn't seem to fit with a locking issue.

libraries/entities/src/EntityTree.cpp
```diff
@@ -2240,8 +2249,19 @@ void EntityTree::fixupNeedsParentFixups() {
...
+
+void EntityTree::forgetAvatarID(const QUuid& avatarID) {
+    std::lock_guard<std::mutex> lock(_avatarIDsLock);
+    _avatarIDs -= avatarID;
+}
+
+void EntityTree::deleteDescendantsOfAvatar(const QUuid& avatarID) {
+    std::lock_guard<std::mutex> lock(_childrenOfAvatarsLock);
+    auto itr = _childrenOfAvatars.find(avatarID);
     if (itr != _childrenOfAvatars.end()) {
```
HifiExperiments, line 2265: ok I think this was the real crash.  I assume `constFind` should check `!= constEnd()`.  In the old code, `constFind()` would NEVER equal `end()`, so it would always fall into this code and then crash.

libraries/entities/src/EntityTree.h
```diff
@@ -364,8 +364,10 @@ class EntityTree : public Octree, public SpatialParentTree {
     QVector<EntityItemWeakPointer> _needsParentFixup; // entites with a parentID but no (yet) known parent instance
     mutable QReadWriteLock _needsParentFixupLock;
 
+    std::mutex _avatarIDsLock;
```
HifiExperiments, line 367: everybody loves mutexes of questionable necessity!

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-78f2e80_ef105e7d.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-78f2e80.tar_ef105e7d.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-78f2e80.tar_ef105e7d.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-78f2e80_ef105e7d.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR435/Vircadia-Alpha-PR435-78f2e80_ef105e7d.exe

libraries/entities/src/EntityTree.cpp
ctrlaltdavid: I wonder how this problem manifested for Darling but seemingly not for anyone else? (I couldn't repro the crash.)

HifiExperiments: I agree, it’s weird!  could be a platform specific thing, maybe on some systems end() does equal constEnd()

kasenvr:
Since this crash is not easy to reproduce, the best thing to do will be to run the server, use it a bit, and if all goes well then it should be considered alright.

daleglass:
Hi, sorry for the delay.

So I looked again at the original report and at this... the code looks OK to me, but does it actually solve the right problem? The bug we're looking at has this:

    Exception thrown at 0x00007FF7DD16C43C in assignment-client.exe: 0xC0000005: Access violation reading location 0x0000000000000014.

This happens in hhash.h, [Inline Frame] assignment-client.exe!QHash<EntityItemID,QHashDummyValue>::isEmpty() Line 269 and the exact code is:

    inline bool isEmpty() const { return d->size == 0; }


So either `this` or `d` are 0x14, which is obviously wrong.

That looks like some sort of buffer overflow somewhere. Is a locking error able to corrupt the hash in this manner?

HifiExperiments:
@daleglass a locking issue could definitely explain the problem, and the code definitely wasn’t threadsafe, but unless we have a solid repro, I don’t have any way to really confirm that this fixes the problem.  it can’t hurt though!

PR #441, opened 2020-06-11
improve crashpad reporting of unhandled c++ exceptions

odysseus654:
- Handle C++ exceptions, passing them to Crashpad before they get to the unknown exception handler
- Attempt to unpack the C++ exception data inside the internal SEH structures to get the variable type (and any superclasses) being thrown.


interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -70,8 +76,99 @@ LONG WINAPI vectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo) {
...
+      return; // broken assumption
+    }
+    CatchableType_internal* pCatchableType = reinterpret_cast<CatchableType_internal*>(moduleBase + pCatchableTypeArray->arrayOfCatchableTypes[0]);
+    if(pCatchableType->pType == 0) {
+      return; // broken assumption
+    }
+    const std::type_info* type = reinterpret_cast<std::type_info*>(moduleBase + pCatchableType->pType);
+
+    // we're crashing, not really sure it matters who's currently holding the lock (although this could in theory really mess things up
+    annotationMutex.try_lock();
```
daleglass, line 153: Return code isn't checked. Mutex isn't ever unlocked.

This seems very dubious. If the idea here is that since we're crashing, the mutex may not ever be unlocked, perhaps we should spin and wait here for some suitable timeout instead, and then abort outright if it doesn't work? A corrupt report is unhelpful anyway.

I'd like to hear a good reason for proceeding without taking the lock. Is it likely to happen? Is the resulting bug report going to be useful?

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -70,8 +76,99 @@ LONG WINAPI vectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo) {
...
+    // now we start breaking the pThrowInfo internal structure apart
+    if(pThrowInfo->pCatchableTypeArray == 0) {
+      return; // broken assumption
+    }
+    CatchableTypeArray_internal* pCatchableTypeArray = reinterpret_cast<CatchableTypeArray_internal*>(moduleBase + pThrowInfo->pCatchableTypeArray);
+    if(pCatchableTypeArray->nCatchableTypes == 0 || pCatchableTypeArray->arrayOfCatchableTypes[0] == 0) {
+      return; // broken assumption
+    }
+    CatchableType_internal* pCatchableType = reinterpret_cast<CatchableType_internal*>(moduleBase + pCatchableTypeArray->arrayOfCatchableTypes[0]);
+    if(pCatchableType->pType == 0) {
```
daleglass, line 147: Can we have some constants here? What's type 0?

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -70,8 +76,99 @@ LONG WINAPI vectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo) {
...
+        fatalCxxException(pExceptionInfo);
+        client->DumpAndCrash(pExceptionInfo);
+    }
+
     return EXCEPTION_CONTINUE_SEARCH;
 }
+
+#pragma pack(push, ehdata, 4)
+
+struct PMD_internal {
```
daleglass, line 89: Where does all this come from? Please link to some documentation

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -70,8 +76,99 @@ LONG WINAPI vectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo) {
...
+// mess at crashpad
+void fatalCxxException(PEXCEPTION_POINTERS pExceptionInfo) {
+    PEXCEPTION_RECORD ExceptionRecord = pExceptionInfo->ExceptionRecord;
+
+    if(ExceptionRecord->NumberParameters != 4 || ExceptionRecord->ExceptionInformation[0] != 0x19930520) {
+        // doesn't match expected parameter counts or magic numbers
+        return;
+    }
+
+    //ULONG_PTR signature = ExceptionRecord->ExceptionInformation[0];
```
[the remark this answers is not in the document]
odysseus654: I think I added this line here at least partially for documentation as to what the parameters are for C++ exceptions (likely could comment out the next line too, we're not using it).

signature in this case is a "magic number" described above.  There are only 3 parameters in 32bit applications.

I'll try to get some of the pages I used to construct this but the info is a bit... scattered.

interface/src/CrashHandler_Crashpad.cpp
odysseus654: This is an "RVA pointer" (offset from image base) so 0 in this case is "NULL".  I'm just trying to do extra sanity checks as I descend through a barely-documented structure to make sure things make as much sense as I can make it.

interface/src/CrashHandler_Crashpad.cpp
odysseus654: There may not be an include file for these, they're generally defined internally by the compiler -- apparently improperly though (the compiler versions use 64bit numbers that aren't compatible with the actual structures).  I did copy some of this stuff from... somewhere though

interface/src/CrashHandler_Crashpad.cpp
odysseus654: The only contention here would be if a different thread was trying to set an annotation while the system was crashing.  I would have no problem with a spinlock with a timeout, but considering the system is in a degraded state I would want a timeout with "proceed anyhow without lock".

I'd appreciate suggestions here for improving this.

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -70,8 +76,99 @@ LONG WINAPI vectoredExceptionHandler(PEXCEPTION_POINTERS pExceptionInfo) {
         client->DumpAndCrash(pExceptionInfo);
     }
 
+    if (pExceptionInfo->ExceptionRecord->ExceptionCode == 0xE06D7363) {
```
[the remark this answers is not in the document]
odysseus654: I can likely make it a constant somewhere, was kinda split on that as this would be the only place it would be used.

odysseus654:
This patch intends two changes:
- Make sure that C++ crashes crash as early as possible.  This is done with the addition of DumpAndCrash done in vectoredExceptionHandler
- Try to retrieve the C++ exception type.  This is done through the fatalCxxException function, which currently does a lot of work for just getting 1-2 annotation strings out of the deal

odysseus654:
This is what a C++ exception currently looks like in Sentry.  The exception is not caught until well after the exception information is lost (when terminate() is called) and the entire stack is saturated by the exception handling (the bottom/earliest call here is to RtlRaiseException)
image

This is what an exception looks like after this patch is applied, where the exception is caught earlier and the stack is more appropriate to the location where the issue occurred:
image

In addition the fatalCxxException function tries to extract the type of exception that is thrown along with a comma-delimited list of classes that could potentially catch it:
image

interface/src/CrashHandler_Crashpad.cpp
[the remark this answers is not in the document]
daleglass: Right, what does `0xE06D7363` stand for? Surely this has some name to it.

Also it's in the coding standard.

odysseus654:
This is a description of classes predefined in the compiler, some of the code in this patch was copied from "Exception Handling (C++ Only)": http://www.geoffchappell.com/studies/msvc/language/predefined/

StackOverflow of someone trying to catch 64bit exceptions -- This is not the first time I have learned about RVA pointers inside the exception structures, but this is the page I found describing it this time (a fair amount of code came from here): https://stackoverflow.com/questions/39113168/c-rtti-in-a-windows-64-bit-vectoredexceptionhandler-ms-visual-studio-2015

Exploration of how C++ exceptions are thrown: https://www.codeproject.com/Articles/175482/Compiler-Internals-How-Try-Catch-Throw-are-Interpr

Another stackoverflow on catching C++ exceptions from an SEH context: https://stackoverflow.com/questions/21888076/how-to-find-the-context-record-for-user-mode-exception-on-x64

interface/src/CrashHandler_Crashpad.cpp
daleglass: Mm. I'll admit that I'm a bit out of my depth here, but this PR is for exception handling, so probably the application is otherwise in an usable state, and all the other threads are running, right? So all existing threads should be proceeding forward without trouble?

Also, if a thread stops with the mutex held, the likelihood of it corrupting the crashpad_info state is high, and if that happens, we would seem to be screwed anyway.


But (very rough idea here), how about an atomic swap scheme?

odysseus654:
Note: if the exception types are recognized (like "std::exception") I am planning on reinterpret_cast'ing them to get the exception message.  Kinda got messed up with my crash-on-start (I've also got a UI change in the wings to add "throw c++ exception" to the "crash me" menu)

interface/src/CrashHandler_Crashpad.cpp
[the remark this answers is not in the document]
odysseus654: as far as I can tell it means "Microsoft C++ Exception".

I tried to take an ascii chart to it in case it came up with a word but I didn't find anything.

interface/src/CrashHandler_Crashpad.cpp
odysseus654: Atomic swap wouldn't really work as this mutex is protecting crashpadAnnotations which is a "crashpad::SimpleStringDictionary"

odysseus654: I'm kinda arguing in my brain about this, yes it's unlikely we could deadlock inside the crash logic waiting for the mutex but... it does seem to be a possibility.  At the moment I think I'd prefer a spinlock with a timeout of like a couple seconds (the lock really shouldn't be getting held that long)

odysseus654: Hmm, there is a second, more complex way of adding annotations to crashpad that we aren't currently using; I could look into using that for this which may bypass the need for locking the map altogether (as we wouldn't be using it)

interface/src/CrashHandler_Crashpad.cpp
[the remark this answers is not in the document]
odysseus654: Would you suggest a different/better way of documenting what each of the "exception parameters" this thing is called with are?

interface/src/CrashHandler_Crashpad.cpp
daleglass: The rough idea I have is:

1. Have a crashpad::SimpleStringDictionary* current_annotations
2. Take the lock.
3. Create a new crashpad::SimpleStringDictionary
4. Copy the entirety of what's in current_annotations
5. Apply the change.
6. Atomically overwrite current_annotations with the pointer to the new, modified dictionary.
7. Unlock.

Then the crash code can go by whatever is in current_annotations, without having to take a lock.

interface/src/CrashHandler_Crashpad.cpp
daleglass: Does that mean this is compiler specific?

interface/src/CrashHandler_Crashpad.cpp
odysseus654: Can't change the pointer to annotations or move it around memory like that I don't think, it's part of the glue that connects this to the external crashpad executable.

There is a second type of annotation that doesn't use this map that I may be able to use.  Registering it with the external executable involves creating a linked list.  Nowhere else in the program is using the second methodology so we may not even need to take out a mutex (If the client code were properly written then THIS would be a good place for atomic variables)

interface/src/CrashHandler_Crashpad.cpp
odysseus654: Yes this is very compiler-specific, if anyone wants to compile this for windows without using the Microsoft C++ compiler then all this code would need to be chucked.  There is a "#ifdef Q_OS_WIN" earlier in the file which I'm hoping would catch this.  I'm also making the assumption that we will not be making 32bit builds anytime soon.

interface/src/CrashHandler_Crashpad.cpp
daleglass: Hmm. I poked at it, and realized that crashpad_info->set_simple_annotations(crashpadAnnotations) is called in setCrashAnnotation, so this whole scheme I had in mind may not work.

But upon further thought, I still don't think ignoring the mutex is going to be any good. It's needed to protect Crashpad's own data, and if that's corrupt then it's unlikely to do much good. It'll either crash there, or produced a garbled report in the best case, and that doesn't seem like an useful thing.

Plus, you're talking about exception handling here, and an exception shouldn't involve such dramatic malfunction that other threads suddenly grind to a halt, right?

odysseus654:
I just tried again to get a version of this code working without mutexes -- by implementing the missing functions that are declared in the .h include files but apparently not linked in.  Unfortunately the additional information appears to have been ignored in the submission, suggesting that the crashpad .exe file is much older than the include .h files would suggest and don't implement everything listed there.

This is starting to become a longer road here.  My next step would be to try to bring in a more modern crashpad into the mix, but is there something we should be doing here in the meantime to get this clear?

daleglass:
@odysseus654 Well, maybe there's no need to complicate it that much either. I'd say give it a try with using mutexes the way they're supposed to first. My guess so far is that the cases where mutex interferes with crash reporting in such a way that ignoring it would do any good will be very small if any, anyway.

Here's my list of things to improve still:

* Link to the documentations in comments, so that if somebody looks at this years later they can find at least a starting point.
* Perhaps make it clearer that this is compiler (and probably compiler version) specific.
* I think it would be also good to use the annotations to store where the code has bailed out when one of the checks fails, so that if things go wrong, the resulting report tells us what assumption broke. This seems likely to happen eventually with new compiler versions.

odysseus654:
Hmm, thinking about this more, I may be more comfortable with testing the return value of the mutex and just bailing early if the instant-try fails.  Would be nice if there were a timeout value that could be passed to the function.

I'm not sure how to make things clearer that this is very windows (Microsoft C++)-specific, the code is already #ifdef'ed out of the way.

I can see about adding annotations about where assumption failures go (assuming the mutex can be retrieved)

daleglass:
@odysseus654 I'm still not quite getting what you're concerned about with the mutex. It protects Crashpad's own data, so if you can't take it because something is stuck, it seems to me you're unlikely to be able to create an useful report anyway. So what's the harm in just taking it like a normal lock?

Besides, we have an exception and not a crash, so all threads should be in working order, unless I'm missing something here.

odysseus654:
I don't like introducing something that may potentially cause a deadlock in the middle of the crash mechanism, unexpected though it might be.

Right now I think the options are:
- grab the lock, waiting indefinitely for the lock to be available
- attempt to grab the lock, bail if it's not immediately available
- attempt to grab the lock, if not immediately available just bypass it and write the structure.

Right now the code is doing the third one, I'm thinking of moving it to the second (we lose the thrownObject and thrownObjectLike extras if this happens), and it sounds like you would prefer the first.

daleglass:
@odysseus654 Well, I'm just not seeing what good could come out of scenario 3. The only place where the lock is taken is setCrashAnnotation.

The first thing we have there is a 'new' that'll only ever be run once during execution. We can ignore this part, because I'm going to bet >99% of the time the crash won't be on the very first invocation.

Then there's a std::replace that should really be outside the lock.

This leaves `crashpadAnnotations->SetKeyValue(name, value);` as the place where it's pretty much guaranteed to get stuck, if that ever happens. And that's exactly the very structure that needs protecting and needs to be in working order.

So in the case where we lock up while this hashtable is mid-modification, we're going to modify that again from another thread? That doesn't seem particularly likely to go well.

I'd go for the first, or for just spinning around trying to take the lock for a second or two before bailing out if you're concerned about the process just hanging forever.

odysseus654:
Spinlocks don't seem to be natively defined under either Qt or STL, but there does seem to be some classes out there with people creating their classes.  I can take a look at doing something like this.

And you're right, the only thing the lock is protecting is the call to SetKeyValue.

daleglass:
Why complicate it? Why not:

    QTime timeout;
    bool acquired = false;

    timeout.start();
    while( !acquired && timeout.elapsed() < 1000 ) {
        acquired = annotationMutex.try_lock();
    }
    if ( !acquired )
        // bail out

odysseus654:
Gack I forgot to add the links to random documentation pages that you suggested.

Take a look at what's here in any case, I ended up doing things "the hard way" and did a "proper" implementation of spinlocks (or at least as close as you can get with Qt).

daleglass:
Well, uh...

Personally I think you're complicating this more than it needs be, and don't see a reason to do more than try_lock in a loop for a while, if that even needs doing at all.

But on the other hand I can't help but appreciate the dedication, and this is an obscure enough corner of the codebase that I figure that in the worst case we'll have to poke at the problem some more.

odysseus654:
Heh, "approved due to excessive dedication".

Should I try to get the error-annotations and documentation links into here as suggested before it gets merged?

JulianGro:
Is there anything specific that would need to be tested here?
Is there a specific crash to test this with? *Unfortunately* the Interface has been very stable for me.

odysseus654:
A couple responses to this:
- The code I've been using to develop with this has some unknown problem that causes it to crash-on-start (ref: https://sentry.vircadia.dev/organizations/vircadia/issues/86/?query=is%3Aunresolved ).  Because of this I have crashed on every change before pushing it into the repository.
- I have (on another branch) added a "do a C++ exception" menu item to the secret "CRASHME" menu.  However... I haven't included it here as I have been unable to test it because umm my viewer crashes on start.

daleglass:
@JulianGro @odysseus654 You can make a crash menu appear by setting this environment variable:

`HIFI_SHOW_DEVELOPER_CRASH_MENU`

odysseus654:
...except that crash menu doesn't (currently) have an option for throwing a C++ exception.

I'm gonna try to rebase this to v2020.2.2-rc, see if I can get a working Interface, and merge/test the crash-menu changes I had queued.  Along with documentation links that were suggested.

odysseus654:
Ugh, looks like my build was crashing because I was hooking first-chance exceptions.  This isn't very well documented.  Trying a different hook now.

odysseus654:
Okay, getting seriously confused here.  I did intentionally throw a c++ exception and crashpad recorded it as such, which means the problem this is trying to solve... is more complicated than it looks like on the surface.

With the most recent commit I'm overriding the crashpad handler and extracting the thrownObject class name which... might be enough for this PR to still be useful.

My focus now is to figure out where the unhandled exception filter is apparently being disabled in so many exceptions that we've been seeing entering Sentry.  Doing some searches suggest this can happen when a DLL with a foreign CRT is loaded and resets things back to default.  I'll be adding code to re-establish the crash handler for the Steam check (which includes a LoadLibrary) command, otherwise... where else are we loading foreign DLLs after the crashpad hook is created? Oculus?

odysseus654:
Okay, added hook to run once a minute and re-establish the unhandled exception filter if it gets broken (and log something in the log file about it, if it ever does happen).  Also added some more comments into the mix that hopefully makes things clearer.

I'm hoping this is the last code commit I have to do on this.  So I guess do a code review of recent work, push it to testing (there's a new option in the crash menu), and see if we can move forward on it.

Not really sure what we can do about testing other than make sure it doesn't break something, especially as we don't have a proper repro on the  0x40000015 exceptions littered through Sentry.  I suspect I would not be able to repro this on my machine in any case.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-44cb8bb_af6d2b2e.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-44cb8bb_af6d2b2e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-44cb8bb.tar_af6d2b2e.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-44cb8bb.tar_af6d2b2e.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-44cb8bb_af6d2b2e.exe

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -58,29 +141,179 @@ static const QString CRASHPAD_HANDLER_NAME { "crashpad_handler" };
...
+
+    if (ExceptionRecord->NumberParameters != 4 || ExceptionRecord->ExceptionInformation[0] != MSVC_CPP_EXCEPTION_SIGNATURE) {
+        // doesn't match expected parameter counts or magic numbers
+        return;
+    }
+
+    // get the ThrowInfo struct from the exception arguments
+    ThrowInfo_internal* pThrowInfo = reinterpret_cast<ThrowInfo_internal*>(ExceptionRecord->ExceptionInformation[2]);
+    ULONG_PTR moduleBase = ExceptionRecord->ExceptionInformation[3];
+    if(moduleBase == 0 || pThrowInfo == NULL) {
```
ctrlaltdavid, line 258:
```suggestion
    if (moduleBase == 0 || pThrowInfo == NULL) {
```
Space after "if", here and elsewhere.

interface/src/CrashHandler_Crashpad.cpp
```diff
@@ -58,29 +141,179 @@ static const QString CRASHPAD_HANDLER_NAME { "crashpad_handler" };
...
+      return; // broken assumption
```
ctrlaltdavid, line 259:
```suggestion
        return; // broken assumption
```
Indent 4 spaces, here and elsewhere.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-6c82ff8_2f00c3e0.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-6c82ff8_2f00c3e0.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-6c82ff8.tar_2f00c3e0.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-6c82ff8.tar_2f00c3e0.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-6c82ff8_2f00c3e0.exe

kasenvr:
Which crash should I perform to best test the reporting effectiveness of this? What I will do is crash myself on a release build with debug artifacts already uploaded and see what happens, then do the same on a compiled and uploaded build of this.

kasenvr:
This needs a code re-review.

odysseus654:
The only entries on the crash menu relevant to this are the ones that throw C++ exceptions.

Is there a comment that I missed here that I need to resolve?

kasenvr:
Sentry is giving trouble so this is not testable right at this moment.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-2b974ba.tar_2f00c3e0.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-2b974ba.tar_2f00c3e0.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-2b974ba_2f00c3e0.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-2b974ba_2f00c3e0.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-efcbc04.tar_2f00c3e0.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-efcbc04.tar_2f00c3e0.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-efcbc04_2f00c3e0.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-efcbc04_2f00c3e0.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-efcbc04_2f00c3e0.exe

daleglass (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-ccee9ee_b8181d92.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-ccee9ee.tar_b8181d92.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-ccee9ee.tar_b8181d92.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-ccee9ee_b8181d92.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR441/Vircadia-Alpha-PR441-ccee9ee_b8181d92.exe

PR #442, opened 2020-06-11
Fix models not scaling correctly when loading due to timeout.

kasenvr:
**Issue:** Loading a model through the Create App would cause its dimensions to be set to 1m^3. This would happen because the app would wait for the model to load before checking its dimensions and setting them, however the app had a strict timeout of 1s. So, if the model is in your cache then it would work. If the model was cachebusted or just in general new, it would not work as the loading would *likely* take more than 1 second.

**Solution:** I raised the timeout to 10 seconds and also put a message if the model gets wrongly sized due to running out of time that way the user will at least have some indication of what happened, and why, and so will developers.

**Test Plan:** Load a model entity into the world that is not cached, cachebust one if you would like. If it comes in under 10 seconds, it should be set to the correct dimensions. If it takes more than 10 seconds, it should be 1m^3.

### DIFF
--- a/scripts/system/create/edit.js
+++ b/scripts/system/create/edit.js
@@ -621,6 +621,10 @@ var toolBar = (function () {
                         dimensions: naturalDimensions
                     })
                     dimensionsCheckCallback();
+                    // We want to update the selection manager again since the script has moved on without us.
+                    selectionManager.clearSelections(this);
+                    entityListTool.sendUpdate();
+                    selectionManager.setSelections([entityID], this);
                     return;
                 }
                 Script.setTimeout(entityIsLoadedCheck, LOADED_CHECK_INTERVAL);
--- a/scripts/system/create/edit.js
+++ b/scripts/system/create/edit.js
@@ -603,15 +603,19 @@ var toolBar = (function () {
                 Script.setTimeout(dimensionsCheckFunction, DIMENSIONS_CHECK_INTERVAL);
             }
             // Make sure the model entity is loaded before we try to figure out
-            // its dimensions.
-            var MAX_LOADED_CHECKS = 10;
+            // its dimensions. We need to give ample time to load the entity.
+            var MAX_LOADED_CHECKS = 100; // 100 * 100ms = 10 seconds.
             var LOADED_CHECK_INTERVAL = 100;
             var isLoadedCheckCount = 0;
             var entityIsLoadedCheck = function() {
                 isLoadedCheckCount++;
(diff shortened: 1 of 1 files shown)


ctrlaltdavid:
Need a description of the problem / test plan.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-f69b051.tar_39104483.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-f69b051_39104483.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-f69b051.tar_39104483.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-f69b051_39104483.exe

SilverfishVR:
Models come in at correct scale but if the user is in create mode and tries to rescale immediately (with gizmo) it pops back to 1x1x1.
does not happen if you deselect, and then select and rescale.

kasenvr:
That's good feedback, I'll take a look. :)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-9c77dbc_3e585267.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-9c77dbc.tar_3e585267.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-9c77dbc_3e585267.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-9c77dbc.tar_3e585267.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR442/Vircadia-Alpha-PR442-9c77dbc_3e585267.exe

SilverfishVR:
Testing _PR442-9c77dbc_

It seems to all work now, no issues found 👍

PR #505, opened 2020-07-03
Increase maximum texture size to 8192

daleglass:
There is interest in increasing the maximum texture size. This seems reasonable, since most of our users own their own domain, and decide which kinds of users to target, so there seems little need to restrict them.

According to https://feedback.wildfiregames.com/report/opengl/feature/GL_MAX_TEXTURE_SIZE the 8192 size should be supported by 90% of the surveyed hardware, which includes 10 year old cards. That seems quite reasonable.

This needs testing to see if:

1. A 8192x8192 texture actually works
2. What happens when it's used on hardware that doesn't support it (if we can find hardware old enough)

### DIFF
--- a/interface/src/ModelPackager.cpp
+++ b/interface/src/ModelPackager.cpp
@@ -24,7 +24,7 @@
 #include "ModelPropertiesDialog.h"
 #include "InterfaceLogging.h"
 
-static const int MAX_TEXTURE_SIZE = 1024;
+static const int MAX_TEXTURE_SIZE = 8192;
 
 void copyDirectoryContent(QDir& from, QDir& to) {
     for (auto entry : from.entryInfoList(QDir::Dirs | QDir::Files | QDir::NoDotAndDotDot |
--- a/libraries/hfm/src/hfm/HFM.h
+++ b/libraries/hfm/src/hfm/HFM.h
@@ -51,7 +51,7 @@ using ColorType = glm::vec3;
 #define HFM_COLOR_ELEMENT gpu::Element::VEC3F_XYZ
 #endif
 
-const int MAX_NUM_PIXELS_FOR_FBX_TEXTURE = 2048 * 2048;
+const int MAX_NUM_PIXELS_FOR_FBX_TEXTURE = 8192 * 8192;
 
 
 using ShapeVertices = std::vector<glm::vec3>;
--- a/libraries/image/src/image/TextureProcessing.cpp
+++ b/libraries/image/src/image/TextureProcessing.cpp
@@ -41,7 +41,7 @@ using namespace gpu;
 
 static const glm::uvec2 SPARSE_PAGE_SIZE(128);
 static const glm::uvec2 MAX_TEXTURE_SIZE_GLES(2048);
-static const glm::uvec2 MAX_TEXTURE_SIZE_GL(4096);
+static const glm::uvec2 MAX_TEXTURE_SIZE_GL(8192);
 bool DEV_DECIMATE_TEXTURES = false;
 std::atomic<size_t> DECIMATED_TEXTURE_COUNT{ 0 };
 std::atomic<size_t> RECTIFIED_TEXTURE_COUNT{ 0 };


JulianGro:
Frankly speaking, I don't think hardware that old can run Vircadia anyways. My first graphics card supported 8192 textures and that probably would have had problems running Vircadia at 190x80 even.

kasenvr:
What model was that? If it's pre say a GTX 260, then we know support for that is 100% deprecated. Lol

JulianGro:
Nvidia Geforce 9800 GT

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR505/Vircadia-Alpha-PR505-515b527.tar_9f258916.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR505/Vircadia-Alpha-PR505-515b527_9f258916.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR505/Vircadia-Alpha-PR505-515b527.tar_9f258916.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR505/Vircadia-Alpha-PR505-515b527_9f258916.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR505/Vircadia-Alpha-PR505-515b527_9f258916.exe

SilverfishVR:
* It runs on my laptop gtx920mx, the 8k texture look crisp 👍 framerate is not that bad (but not good, never was).
* both image entities and materials displays at full resolution 👍 

1/8 of a texture on a 20x20m plane, 2k vs 8k:
Screenshot 2020-07-04 20 20 08

Could not test on my servers integrated graphics (i5 4690k) as interface requires minimum OpenGL 4.1 to run

JulianGro:
Here is a skybox texture to test if it shows up in this PR. In current master it is too big and the renderer will just not show it.
http://motofckr9k.ddns.net/m/vircadia/purple_nebula_skybox_4096.png

kasenvr:
That's a beautiful example of a skybox with the stars, so I really hope that this bump can make more beautiful skyboxes possible.

daleglass:
Reading some documentation, I found that OpenGL 4 requires 16k texture support, so we could probably go that far as well.

The concern I have there is what kind of user experience will that provide. The "maximum texture memory" setting is set to "auto" by default, which happens to be a hard coded 1 GB (so no auto whatsoever). And a 16K texture takes 1GB VRAM when uncompressed, so that's a problem.

Then there's the issue of that we stream everything, so what could be reasonable in a modern game where everything is on-disk beforehand, having to download even a fraction of 1GB is quite painful, not to mention expensive to serve. S3 costs $0.09 per GB, which adds up quickly. That's less of a concern but probably warrants some testing.

kasenvr:
I feel like for people, they'll know when their costs get that big. 16k is *way* above 4-8k in actual pixel count. I imagine they would still compress it and use those high-res textures only where needed. Especially if they are paying for their own hosting. ;)

daleglass:
I mean more of the end-user experience. Like what happens when somebody on a 128 Kbps connection tries to download one of those monster textures? What happens with Quest users?

At that point we should probably consider implementing some sort of downscaling proxy that limited users could take advantage of to request lower resolution versions.

kasenvr:
I sense this is where we really need to think about actual LOD architecture... Something along the lines of actually specifying different models and different resolution targets so different clients and graphical settings can request and use things as they see fit.

JulianGro:
Maybe we should just keep to 8k for now and think about 16k once we have some way to control this.

kasenvr:
Can someone try going to a large world with this PR? For existing worlds, any unoptimized textures might end up being bigger than was originally planned for, no? That's fine as long as the adverse effects are minimal. This *should* be controlled by the max texture memory I'd assume.

SilverfishVR:
I have visited a few domains on this PR (hub, hyperspace, spot) no big surprises, I think most creators knows not to silly size textures, I can shop around a bit more.

on a side note, aside from skyboxes, you have always been able to use textures that were bigger than this max size, it just gets downscaled, so once the graphics settings gets the much needed overhaul, _maximum texture size_ could be exposed as a user setting.

Aitolda:
Seems to work as intended, however combining a material with Diffuse, Roughness, Occlusion, and Normals all 8k is not recommended.

kasenvr:
> 
> 
> Should we stick this in a constant somewhere so we can change it easily in one place and eventually change it to a user setting?
> 
> It would also be good to check how the oven handles this, I don’t remember if it has a limit. 8k textures might take MUCH longer to bake. I know Quest + Android have their own limit (2048) as they should which is good.
> 
> Otherwise looks good.

All the graphics settings are going to be improved and put into a proper menu in due time, this will definitely be one of those options. :)

Issue #398, opened 2020-05-28
Zones with scripts don't get enterEntity calls on first try


AleziaKurdis:
I have a Client Entity Script that creates a zone and then fails to update the "keyLight.direction" properties using Entities.editEntity. 
The result is that the zone looks like completely as inheriting the higher zone over it, despite that the Create app display the expected values

To reproduce:
In a Test domain or in the Serverless domain, import this json:
**http://metaverse.bashora.com/scripts/hytrion/galax_Edit.json**
This create a zone that run a script _(http://metaverse.bashora.com/scripts/hytrion/galactic_sky_edit.js)_ that generates a sub zone and update it's "keyLight.direction" each 3 seconds.
But that sub zone (named "DEEPSPACE") appears completely black. (as the zone over it)

I discovered after a couple of hour of test, that "Entities.editEntity", to update correctly a zone's attribute, needs that we force an update of the Mode attributes ("keyLightMode", "ambientLightMode", "skyboxMode",  "hazeMode" and "bloomMode") **with at least one** attribute of that mode.
Without this, the display ignores the update and behave as if the mode was set to "inherited".

I was expecting to be able to update only one specific attribute using the Entities.editEntity and have the existing context to be assumed as it is.
To me, this is clearly a bug.

Also to workarround the issue, I had to include this in my update:
                "keyLightMode": "enabled",
                "ambientLightMode": "enabled",
                "skyboxMode": "enabled",
                "hazeMode": "enabled",
                "bloomMode": "enabled",  
in addition of at least one attributes under each of these Mode, even if I didn't need that.

It's clearly not intuitive.

AleziaKurdis:
Seems to be related to events like enterEntities that are not triggered when a script is attached to an entities. Apparently for the session where this script runs for the first time.
The impact might affect lot of things. This script will never work well when a visitor come the first time in a domain. The owner might conclude it working when he visit a second time. (until a cache refresh) 
Not sure this is low severity.

HifiExperiments:
I looked into this and I think you're right that the issue is related to enterEntity, not the zone properties specifically.  I can create a zone and edit its keyLight direction from the console without changing the other properties.

[...]

PR #507, opened 2020-07-05
JS Doc update for enterEntity signal.

AleziaKurdis:
To address issue #398
"**Zones with scripts don't get enterEntity calls on first try**"

It has been agreed that the current solution wasn't a real issue at the moment we documented the behavior to advise the user and give him a solution.
This PR simply add a note in the documentation of the "enterEntity" signal.

From this:
**Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event.** 

to this:
**Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event. Note: At the initial loading of the script, if the avatar is already present inside the entity, it might be too late to catch this event when the script runs, so it won't trigger. The 'preload' signal can be used to handle those cases.**


libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -2399,7 +2399,9 @@ public slots:
 
 
     /**jsdoc
-     * Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event.
+     * Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event.
```
ctrlaltdavid, line 2402:
```suggestion
     * Triggered when an avatar enters an entity. 
```
The "but only if" text isn't needed because it's use is explained in the two "see also" links included in the description. Or if you think it needs expanding, then it needs expanding for all entity method signals.

libraries/entities/src/EntityScriptingInterface.h
```diff
@@ -2399,7 +2399,9 @@ public slots:
 
 
     /**jsdoc
-     * Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event.
+     * Triggered when an avatar enters an entity, but only if the entity has an entity method exposed for this event. 
+     * Note: At the initial loading of the script, if the avatar is already present inside the entity, it might be too late 
+     * to catch this event when the script runs, so it won't trigger. The 'preload' signal can be used to handle those cases.
```
ctrlaltdavid, line 2404:
```suggestion
     * Note: At the initial loading of the script, if the avatar is already present inside the entity, it might be too late 
     * to catch this event when the script runs, so it won't trigger. The {@link Entities.preload|preload} signal can be used 
     * to handle those cases.
```
Let's link to the "preload" signal in the docs.

libraries/entities/src/EntityScriptingInterface.h
AleziaKurdis: I can remove that, but there is the same "but only if" for "leaveEntity"
Should I remove it there too? (see Line 2413)

AleziaKurdis: Clearly. After second thought. I went with removing the 2 "but only if..." (line 2402 and 2413)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR507/Vircadia-Alpha-PR507-06307c7_6c24f646.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR507/Vircadia-Alpha-PR507-06307c7.tar_6c24f646.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR507/Vircadia-Alpha-PR507-06307c7.tar_6c24f646.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR507/Vircadia-Alpha-PR507-06307c7_6c24f646.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR507/Vircadia-Alpha-PR507-06307c7_6c24f646.exe

PR #547, opened 2020-07-17
Fix lag on restart with scaled avatar

HifiExperiments:
This fixes https://github.com/kasenvr/project-athena/issues/48, not https://github.com/kasenvr/project-athena/issues/77

### DIFF
--- a/interface/src/avatar/MyCharacterController.cpp
+++ b/interface/src/avatar/MyCharacterController.cpp
@@ -407,13 +407,13 @@ void MyCharacterController::clearDetailedMotionStates() {
 }
 
 void MyCharacterController::buildPhysicsTransaction(PhysicsEngine::Transaction& transaction) {
-    for (size_t i = 0; i < _detailedMotionStates.size(); i++) {
-        _detailedMotionStates[i]->forceActive();
+    for (auto& detailedMotionState : _detailedMotionStates) {
+        detailedMotionState->forceActive();
     }
     if (_pendingFlags & PENDING_FLAG_REMOVE_DETAILED_FROM_SIMULATION) {
         _pendingFlags &= ~PENDING_FLAG_REMOVE_DETAILED_FROM_SIMULATION;
-        for (size_t i = 0; i < _detailedMotionStates.size(); i++) {
-            transaction.objectsToRemove.push_back(_detailedMotionStates[i]);
+        for (auto& detailedMotionState : _detailedMotionStates) {
+            transaction.objectsToRemove.push_back(detailedMotionState);
         }
         // NOTE: the DetailedMotionStates are deleted after being added to PhysicsEngine::Transaction::_objectsToRemove
         // See AvatarManager::handleProcessedPhysicsTransaction()
--- a/interface/src/avatar/OtherAvatar.cpp
+++ b/interface/src/avatar/OtherAvatar.cpp
@@ -116,10 +116,10 @@ void OtherAvatar::updateSpaceProxy(workload::Transaction& transaction) const {
 
 int OtherAvatar::parseDataFromBuffer(const QByteArray& buffer) {
     int32_t bytesRead = Avatar::parseDataFromBuffer(buffer);
-    for (size_t i = 0; i < _detailedMotionStates.size(); i++) {
+    for (auto& detailedMotionState : _detailedMotionStates) {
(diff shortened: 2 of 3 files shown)


libraries/physics/src/CharacterController.cpp
```diff
@@ -543,12 +543,8 @@ void CharacterController::setLocalBoundingBox(const glm::vec3& minCorner, const
         _minStepHeight = DEFAULT_MIN_STEP_HEIGHT_FACTOR * (_halfHeight + _radius);
         _maxStepHeight = DEFAULT_MAX_STEP_HEIGHT_FACTOR * (_halfHeight + _radius);
 
-        if (_physicsEngine) {
```
HifiExperiments, line 546: The issue was that on startup, before the physics engine loaded, we would set the ADD flags repeatedly as we scaled to the correct size without setting the REMOVE flags, so we'd end up with a ton of duplicate motion states.  The number would depend on loading times and the target scale (smaller scales take longer to get to, so the problem is worse).  In one case, I had 1500 motion states when it should have been 50.

kasenvr:
Wait, so this will solve my avatar lagging on load until such a time that I resize?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR547/Vircadia-Alpha-PR547-1346a15_cbce911e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR547/Vircadia-Alpha-PR547-1346a15.tar_cbce911e.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR547/Vircadia-Alpha-PR547-1346a15.tar_cbce911e.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR547/Vircadia-Alpha-PR547-1346a15_cbce911e.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR547/Vircadia-Alpha-PR547-1346a15_cbce911e.exe

HifiExperiments:
@kasenvr yeah!

interface/src/avatar/MyCharacterController.cpp
```diff
@@ -407,13 +407,13 @@ void MyCharacterController::clearDetailedMotionStates() {
 }
 
 void MyCharacterController::buildPhysicsTransaction(PhysicsEngine::Transaction& transaction) {
-    for (size_t i = 0; i < _detailedMotionStates.size(); i++) {
```
daleglass, line 410: This doesn't appear to make any functional change, and to just be a cleanup.

Perhaps this should be in a separate PR?

interface/src/avatar/OtherAvatar.cpp
```diff
@@ -116,10 +116,10 @@ void OtherAvatar::updateSpaceProxy(workload::Transaction& transaction) const {
 
 int OtherAvatar::parseDataFromBuffer(const QByteArray& buffer) {
     int32_t bytesRead = Avatar::parseDataFromBuffer(buffer);
-    for (size_t i = 0; i < _detailedMotionStates.size(); i++) {
```
daleglass, line 119: Same here

libraries/physics/src/CharacterController.cpp
daleglass: Hmm, what could be the original reasoning for the `if (_physicsEngine)`? Is that an attempt at optimization, or could it be protecting something?

interface/src/avatar/MyCharacterController.cpp
HifiExperiments: these are leftover from when I was trying to track down the problem.  since this is a short PR and these changes theoretically improve performance, I would vote to keep them

libraries/physics/src/CharacterController.cpp
HifiExperiments: my guess is that instead of setting flags, this code used to do actual work here.  I don’t see any way that removing this if could cause problems.  perhaps if we actually added an if around the ADD flags, we could optimize this to do less work on startup, but that seems a lot riskier, since I could imagine it leading to cases where we never create the motion states

PR #583, opened 2020-08-02
Requesting Code Review of ByteSlice

odysseus654:
As part of the UDT rewriting I intend to be using two utility new utility classes named ByteSlice and ByteRope.  ByteSlice is the most mature of the classes at this point.  This PR is my request for eyes on this to ensure that (-) it meets coding standards and/or has no major issues with it, and (-) adding classes like this to the codebase makes sense, even with all the string and bytestring classes offered in Qt and STL.


libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // next check to see if we're "lucky" and we already have the data within ourselves
+    size_t rhsLength = rhs.length();
+    if (_offset + _length + rhsLength >= _content->_fullLength && memcmp(&_content->_content[_offset+_length], rhs.constData(), rhsLength) == 0) {
+        // "the answer was within you the whole time"
+        return ByteSlice(_content, _offset, _length + rhsLength);
+    }
+    if (rhs._offset >= _length && memcmp(&rhs._content->_content[rhs._offset - _length], constData(), _length) == 0) {
+        // "the answer was within *them* the whole time"
+        return ByteSlice(_content, rhs._offset - _length, _length + rhsLength);
+    }
```
HifiExperiments, line 79: I'm not sure I understand the logic in these two if blocks...

the first one seems to be saying "if my size + rhs size is bigger than my full length, and the data after me matches rhs, make myself bigger to contain that data".  but if you're looking at data that's out of your bounds, even if the data happens to match rhs in value, there's no guarantee about what that other data is.  it could be used by something else and change out from under us, right?  I'm also confused about the logic for the second one.

but also, it seems like these memcmp checks would often be more expensive than just always allocating a new slice?

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    // okay, didn't work.  Just create a new one from scratch
+    ByteSlice newSlice;
+    quint8* newContent = static_cast<quint8*>(newSlice.create(_length + rhsLength));
+    memcpy(newContent, constData(), _length);
+    memcpy(&newContent[_length], rhs.constData(), rhsLength);
```
HifiExperiments, line 85: shouldn't these copies be taking into account the offsets?

libraries/networking/src/ByteSlice.inl
```diff
@@ -0,0 +1,70 @@
...
+
+inline bool ByteSlice::empty() const {
+    return !_length;
+}
+
+inline size_t ByteSlice::length() const {
+    return _length;
+}
+
+inline const quint8& ByteSlice::operator[](size_t idx) const {
```
HifiExperiments, line 54: should this take into account _offset?

libraries/networking/src/ByteSlice.inl
```diff
@@ -0,0 +1,70 @@
+//
```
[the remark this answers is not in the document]
odysseus654: I tend to disagree with the "compilers are smarter" with regard to inline vs not; unless you are doing "whole program optimization" or "generate code at link time" the compiler wouldn't have access to the code body if it were included in the .cpp file to outside classes.  While the compiler can choose not to inline functions, this also flags the compiler to permit multiple identical definitions of a function with the same name if they appear in multiple compile units during linking.

In terms of whether or not to inline, I'm considering it based on the complexity of the code and whether the compiler might be able to optimize it better having access to the implementation.  As it is the line for "too complex to be inline" is a fuzzy one and I wouldn't be against moving some of the larger functions out of inline.

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+quint8 ByteSlice::pop_front() {
+    if (!_length) {
+        return 0;
+    }
+    --_length;
+    return _content->_content[_offset++];
+}
+
+ByteSlice ByteSlice::substring(size_t offset, size_t length /* = static_cast<size_t>(-1) */) const {
```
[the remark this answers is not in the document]
odysseus654: mostly "my tradition" considering this internal documentation as the default value the function was defined with cannot be included in the implementation.  I can understand if the code style doesn't permit this though.

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (offset >= _length) {
+        return ByteSlice();
+    }
+    size_t maxLen = _length - offset;
+    if (length == static_cast<size_t>(-1)) {
```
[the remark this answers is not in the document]
odysseus654: is there a reason why one form of cast is preferred over the other? I've been trying to remove function-style casting from my style.  Is use of static_cast / const_cast deprecated?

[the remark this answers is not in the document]
odysseus654: I suspect unsigned(-1) is being used as a magic value here, similar to how it's used in functions such as substr -- this is the same value as std::string::npos

libraries/networking/src/ByteSlice.inl
odysseus654: yes, both operator[] and constData() should be taking _offset into account, this is a bug.

libraries/networking/src/ByteSlice.cpp
odysseus654: constData() should be taking the offsets into account (but apparently isn't)

libraries/networking/src/ByteSlice.cpp
odysseus654: okay looks like this particular code may need more checking of the comparisons, some of the inequality checks may be reversed.  This function was added late as the networking code needed some way to concatenate the contents of two packets together.  The memcmp can be skipped if it takes too long; I've been considering malloc() to be fairly intensive though (and in at least older heap implementations it would hold a global lock while allocating).

The idea was kinda if you had an original string named "the quick brown fox jumped over the lazy dog" and you had a ByteSlice representing "lazy" from that string and tried to concatenate " dog" to that then it would be able to reuse the ByteString and not have to allocate something new.  The arguments on this may be similar to that for inline functions though -- cost of passing all the checks (vs skipping once the first one fails) vs just going through and allocating a new object.

Depend on how other code works I may end up removing this function, if another (more specialized) class would be more appropriate (but then it's also a balance between complexity vs performance)

libraries/networking/src/ByteSlice.inl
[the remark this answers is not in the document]
HifiExperiments: my hesitation here is that I don't yet know where or how often this code is being called, which I'm sure you've got a better idea of.  personally I do trust the compiler to make these decisions 99% of the time, and you can see from the rest of the codebase that we use inline fairly infrequently.  but like I said, I would be in favor of moving these short methods to the .h so the compiler would have access to them, and the larger methods to the .cpp, which is also what the coding standard suggests

libraries/networking/src/ByteSlice.cpp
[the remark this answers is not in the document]
HifiExperiments: ah I gotcha.  I agree this can be useful but yeah let's avoid the clutter in the .cpp

libraries/networking/src/ByteSlice.cpp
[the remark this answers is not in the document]
HifiExperiments: the coding standard recommends: https://github.com/kasenvr/project-athena/blob/kasen/core/CODING_STANDARD.md#3121-primitive-types-should-use-c-style-casting

c style casts for primitive, c++ style casts for classes

ah ok I guess std::string uses -1 in a size_t too so yeah it's fine

libraries/networking/src/ByteSlice.cpp
HifiExperiments: I definitely see what this code is trying to accomplish, but yeah it seems like a tradeoff between how often we think that's going to actually happen and how expensive the checks are (I also assume the malloc is more expensive than the compare, but if we're doing both 90% of the time, is it really worth it?)

libraries/networking/src/ByteSlice.inl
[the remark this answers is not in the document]
odysseus654: FWIW this class takes the place of QByteArray in most of the new networking code (and likely will be used more if/when we move away from using QDataStream for our packet encoding).  QByteArray does implicit data sharing when the same string is passed around but does reallocation on any substrings (which this class avoids).  It's possible (with a single-packet message and no obfuscation) that the network message passed to octree (or wherever) has a QByteString initially allocated when calling QUdpSocket::readDatagram (and containing the entire raw UDP packet).

Part of me still feels that I need to emphasize that the compiler has no means to inline a function that is implemented in a different .cpp file than the one being compiled, so the decision here is at least partially based on how many places we want the user of a QByteSlice to have the option to inline when calling a function.  I'm saying this without having looked at the code recently though -- I'm not trying to argue against de-inlining here, only against the possibility that the compiler can choose to inline a function in a different .cpp file without [using proprietary Microsoft extensions](https://docs.microsoft.com/en-us/cpp/build/reference/gl-whole-program-optimization?view=vs-2019).

libraries/networking/src/ByteSlice.cpp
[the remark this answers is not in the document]
odysseus654: thinking I need to constexpr this in the .h file to get away from magic numbers in the code (which I think might be in the style rules somewhere)

libraries/networking/src/ByteSlice.cpp
odysseus654: I'm hoping the "if" statements outside of the call fail early the majority of the time so it never really gets to the memcmp() actually.

odysseus654:
Pushed an update related to feedback on this issue.  Most of the inlines ended up collapsible to a single line of code -- which may or may not have objections (I'm not sure I really like trinary operations and would prefer the more verbose description in many places)

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,96 @@
...
+        return ByteSlice(_content, _offset + offset, std::min(maxLen, length));
+    }
+}
+
+// create a new buffer and return a pointer to it
+void* ByteSlice::create(size_t length) {
+    _offset = 0;
+    _length = length;
+    _content = ByteStringPointer::create(length);
+    return const_cast<quint8*>(_content->_content);
```
HifiExperiments, line 65: this method returns a void* so this cast isn’t necessary

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (empty()) {
+        return rhs;
+    }
+    if (rhs.empty()) {
+        return *this;
+    }
+
+    // next check to see if we're "lucky" and we already have the data within ourselves
+    size_t rhsLength = rhs.length();
+    if (_offset + _length + rhsLength <= _content->_fullLength && memcmp(&_content->_content[_offset+_length], rhs.constData(), rhsLength) == 0) {
```
HifiExperiments, line 81: shouldn’t it just be comparing &_content->_content[_offset], no + _length?

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,96 @@
...
+        // "the answer was within you the whole time"
+        return ByteSlice(_content, _offset, _length + rhsLength);
+    }
+    if (rhs._offset >= _length && memcmp(&rhs._content->_content[rhs._offset - _length], constData(), _length) == 0) {
```
HifiExperiments, line 85: I still don’t understand this check, why does comparing their offset to our length tell us we have enough space?

libraries/networking/src/ByteSlice.h
```diff
@@ -0,0 +1,67 @@
...
+    inline ByteSlice(const QByteArray& data) : _offset(0), _length(data.length()), _content(ByteStringPointer::create(reinterpret_cast<const quint8*>(data.constData()), data.length())) {}
+    inline ByteSlice(const std::string& data) : _offset(0), _length(data.length()), _content(ByteStringPointer::create(reinterpret_cast<const quint8*>(data.c_str()), data.length())) {}
+    inline ByteSlice(const ByteSlice& data) : _offset(data._offset), _length(data._length), _content(data._content) {}
+    inline ByteSlice(ByteSlice&& data) noexcept : _offset(data._offset), _length(data._length) { _content.swap(data._content); }
+    void* create(size_t length);  // create a new buffer and return a pointer to it
+
+    inline size_t length() const { return _length; }
+    inline bool empty() const { return !_length; }
+    void clear();
+    inline const quint8* constData() const { return _content.isNull() ? NULL : _content->_content + _offset; }
```
HifiExperiments, line 40: nullptr instead of NULL

libraries/networking/src/ByteSlice.cpp
```diff
@@ -0,0 +1,96 @@
...
+//  libraries/networking/src
+//
+//  Created by Heather Anderson on 2020-05-01.
+//  Copyright 2020 Vircadia.
+//
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+//
+
+#include "Byteslice.h"
```
HifiExperiments, line 12: ByteSlice.h with uppercase S

libraries/networking/src/ByteSlice.cpp
HifiExperiments: oh wait I get it, this is saying "do they have enough space at their beginning" so this one looks good

libraries/networking/src/ByteSlice.cpp
HifiExperiments: WAIT no I think I get this one too

libraries/networking/src/ByteSlice.cpp
odysseus654: 1) The compiler diagrees:
```
10>C:\devel\vircadia-udt3\libraries\networking\src\ByteSlice.cpp(65,30): error C2440: 'return': cannot convert from 'const quint8 *' to 'void *'
10>C:\devel\vircadia-udt3\libraries\networking\src\ByteSlice.cpp(65,20): message : Conversion loses qualifiers
```

2) This function is the *only* place that a non-const pointer can be received from a ByteString (i.e. the const_cast is very intentional here).  Every other member outside of create() would return an immutable pointer.

HifiExperiments: ah!  my bad, I had misread this, you’re totally right

HifiExperiments (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR583/Vircadia-Alpha-PR583-140ebc5.tar_4819c00b.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR583/Vircadia-Alpha-PR583-140ebc5_4819c00b.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR583/Vircadia-Alpha-PR583-140ebc5.tar_4819c00b.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR583/Vircadia-Alpha-PR583-140ebc5_4819c00b.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR583/Vircadia-Alpha-PR583-140ebc5_4819c00b.exe

kasenvr:
Looks like your PR passed preliminary review, but I don't think you want to actually merge this as is, do you? 🤔

odysseus654:
Thanks so much to @HifiExperiments for the time spent doing this code review for me :-) I have incorporated these changes into the branch I have been working on as well as propagating some of these changes (where applicable) to other files.  Hoping this reduces the amount of review required later on :-)

PR #591, opened 2020-08-05
OAuth2 WordPress

ctrlaltdavid:
Initial release.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4d63a79_5ba33a52.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4d63a79.tar_5ba33a52.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4d63a79.tar_5ba33a52.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4d63a79_5ba33a52.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4d63a79_5ba33a52.exe

kasenvr:
This functionality will be expanded on later to be more generalized for OAuth providers. This feature is especially helpful and useful for domain operators wanting to have their own authentication based on potential paid tiers, groups, membership, etc. Or just have their own user system (a roleplay group?) that is separate from the Metaverse API.

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -1061,3 +1207,91 @@ Node::LocalID DomainGatekeeper::findOrCreateLocalID(const QUuid& uuid) {
...
+
+bool DomainGatekeeper::domainHasLogin() {
+    // The domain may have its own users and groups in a WordPress site.
+    return _server->_settingsManager.valueForKeyPath(AUTHENTICATION_ENABLE_OAUTH2).toBool()
+        && !_server->_settingsManager.valueForKeyPath(AUTHENTICATION_OAUTH2_URL_PATH).toString().isEmpty()
+        && !_server->_settingsManager.valueForKeyPath(AUTHENTICATION_WORDPRESS_URL_BASE).toString().isEmpty()
+        && !_server->_settingsManager.valueForKeyPath(AUTHENTICATION_PLUGIN_CLIENT_ID).toString().isEmpty();
+}
+
+void DomainGatekeeper::requestDomainUser(const QString& username, const QString& accessToken, const QString& refreshToken) {
```
odysseus654, line 1220: There doesn't seem to be any use made of refresh tokens, and no expiration date set for access tokens.  I'm thinking the access token expiration date would need to be included in processConnectRequestPacket (unless there's some way to query for its expiration date?), otherwise this may prompt the user for a login more often than expected.

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -155,6 +175,27 @@ NodePermissions DomainGatekeeper::setPermissionsForUser(bool isLocalUser, QStrin
...
 
+    // If this user is a known member of a domain group, give them the implied permissions.
+    // Do before processing verifiedUsername in case user is logged into the metaverse and is a member of a blacklist group.
+    if (!verifiedDomainUserName.isEmpty()) {
+        auto userGroups = _domainGroupMemberships[verifiedDomainUserName];
+        foreach (QString userGroup, userGroups) {
+            // Domain groups may be specified as comma- and/or space-separated lists of group names.
+            // For example, "@silver @Gold, @platinum".
+            auto domainGroups = _server->_settingsManager.getDomainGroupNames()
+                .filter(QRegularExpression("^(.*[\\s,])?" + userGroup + "([\\s,].*)?$",
```
odysseus654, line 186: do we need to escape special characters in userGroup at all?

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -155,6 +175,27 @@ NodePermissions DomainGatekeeper::setPermissionsForUser(bool isLocalUser, QStrin
...
+                    << "so:" << userPerms;
+#endif
+            }
+        }
+
+    }
+
     if (verifiedUsername.isEmpty()) {
         userPerms |= _server->_settingsManager.getStandardPermissionsForName(NodePermissions::standardNameAnonymous);
 #ifdef WANT_DEBUG
```
odysseus654, line 201: Did we want to add a check to prevent permissions for groups starting with @ to be granted if domainHasLogin() (perhaps wrapping calls to isGroupMember in this function with additional checks?)

(seeing a check in DomainServerSettingsManager::apiRefreshGroupInformation that may satisfy this)

interface/src/Application.cpp
```diff
@@ -7069,19 +7081,23 @@ void Application::updateWindowTitle() const {
...
-
     QString connectionStatus = isInErrorState ? " (ERROR CONNECTING)" :
         nodeList->getDomainHandler().isConnected() ? "" : " (NOT CONNECTED)";
-    QString username = accountManager->getAccountInfo().getUsername();
 
-    setCrashAnnotation("sentry[user][username]", username.toStdString());
+    QString metaverseUsername = accountManager->getAccountInfo().getUsername();
+    QString domainUsername = domainAccountManager->getUsername();
+
+    setCrashAnnotation("sentry[user][metaverseUsername]", metaverseUsername.toStdString());
```
odysseus654, line 7100: changing the attribute from [username] to [metaverseUsername] may be an issue here, [username] was chosen here as it is a recognized keyword that is interpreted in Sentry as per https://develop.sentry.dev/sdk/event-payloads/user/

kasenvr: Good point, we can just use secondary metadata later to explain a domain user and assume the master "user" is a metaverse one. :)

odysseus654: If you do make up your own tags then replace the entire field and don't start it with "sentry[", that way they'll end up at the bottom of any error reports.  Unrecognized keywords in the "sentry" namespace appear to be eaten.

domain-server/src/DomainGatekeeper.cpp
ctrlaltdavid: Will do.

domain-server/src/DomainGatekeeper.cpp
ctrlaltdavid: Yes, the check in `apiRefreshGroupInformation()` prevents this. However, I'll make the test in that method more targeted: filter out group setting values that start with a "@" but let through group setting values that include a "@" just in case metaverse group names do allow special characters after the first character at some stage.

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -419,7 +489,9 @@ SharedNodePointer DomainGatekeeper::processAgentConnectRequest(const NodeConnect
...
             qDebug() << "stalling login because we have no username-signature:" << username;
 #endif
-            return SharedNodePointer();
+            if (!domainHasLogin() || domainUsername.isEmpty()) {
```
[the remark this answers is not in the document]
ctrlaltdavid: No, the user can be logged into both the metaverse and the local domain at the same time.

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -419,7 +489,9 @@ SharedNodePointer DomainGatekeeper::processAgentConnectRequest(const NodeConnect
...
+                return SharedNodePointer();
+            }
         } else if (verifyUserSignature(username, usernameSignature, nodeConnection.senderSockAddr)) {
             // they sent us a username and the signature verifies it
             getGroupMemberships(username);
```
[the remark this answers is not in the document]
ctrlaltdavid: The domain owner can use the domain server settings to set permissions to achieve a variety of scenarios (e.g., disable anonymous, localhost, friends, logged-in from connecting and only enable connecting by domain groups). But we don't want to necessarily restrict access to people logged in via the metaverse on domains that have their own log-in.

domain-server/src/DomainGatekeeper.cpp
ctrlaltdavid: Handling of token expiry and refresh, unfortunately, hasn't made it into this cut of the code. It will be handled in due course.

interface/src/Application.cpp
ctrlaltdavid: Good catch! Will revert to using "[username]".

ctrlaltdavid:
PR for CR changes: https://github.com/kasenvr/project-athena/pull/597

domain-server/src/DomainGatekeeper.cpp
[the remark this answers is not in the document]
odysseus654: Yes, I'm just wondering if there's any "please authenticate" packets being sent when the user has already logged into the domain, I don't clearly understand what "sendConnectionTokenPacket" means.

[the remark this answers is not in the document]
ctrlaltdavid: Ah. `sendConnectionTokenPacket()` etc. will only be called if the metaverse `username` is not empty. The metaverse `username` is kept separate from the domain `domainUserName`. ... So if they're logged into the domain, this code won't be called unless they use the File > Login menu item to log into the metaverse, in which case the metaverse login dialog will open and once they've entered their username etc., this code will get executed to execute the metaverse login request.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4931e08_0897ef4f.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4931e08.tar_0897ef4f.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4931e08_0897ef4f.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4931e08.tar_0897ef4f.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-4931e08_0897ef4f.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-49d3b08_95e0907b.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-49d3b08.tar_95e0907b.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-49d3b08_95e0907b.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-49d3b08.tar_95e0907b.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR591/Vircadia-Alpha-PR591-49d3b08_95e0907b.exe

kasenvr:
If there are further updates and revisions to be made, a new PR should be made for that. :)

Issue #30, opened 2019-12-19
"Auto" Developer/Rendering/Maximum Texture Memory just sets it to 1 GB.


daleglass:
The setting is not "auto" whatsoever and just sets it to a hardcoded value of 1GB. This is incredibly wasteful on modern hardware, with cards having memory amounts of 8GB quite often.

I already looked into it, and intend to try to submit a fix soon-ish. Submitting the bug here for tracking purposes.

kasenvr:
After looking into this, I think it might be a good idea to remove the auto option entirely and put this setting as part of the graphics settings so the user can handle this. Let the "auto" type function be instead having the max texture memory scale with the presets.

Users are used to having to fine-tune graphics settings for all matters of games and 3D apps so they can definitely set the custom amount if need be. If anything, a notice can be given that the amount of memory the world wants is higher than their settings is allowing perhaps?

To make an almost sort of black box algorithm to have it dynamically scale, especially where there are few official/reliable cross-platform ways to ascertain enough information about available, total, used video memory seems quite a bit of an undertaking at this time, too early on.

Though, the app initially setting the max memory to be perhaps 70% of the maximum available to the system (suggested by Dale) would be a good idea where it's possible to ascertain the necessary parameters.

PR #599, opened 2020-08-06
Initial version of automatic texture memory size

daleglass:
This fixes the automatic texture memory size option.

In the original code, "auto" simply hard-codes texture size to 1 GB, and isn't "auto" whatsoever.

This PR implements 3 policies, with the one we use depending on what we can know:

* If we know how much VRAM is free (NVIDIA and ATI report this), then we query that. The code will then use available VRAM, leaving some amount (currently hardcoded as 64MB).
* If we don't know how much VRAM is free, but we do know how much there is in total, we set the limit to a percentage of that. This  happens on Intel. Currently the max use is hardcoded to 80%. The assumption here is that this is a low-spec machine, and unlikely to be running anything else at the same time, so we just need to leave some margin for system use.
* If we don't know anything, fallback to the old method, which just goes with 1GB. That seems reasonable considering my ancient work laptop that's absolutely not fit for running Vircadia reports 1.5 GB.

The first case needs testing. It's not a fixed memory limit, but tries to use whatever memory is free, so it's also going to be influenced by anything else on the system that allocates VRAM.

This fixes bug #30

### DIFF
--- a/libraries/gl/src/gl/Config.cpp
+++ b/libraries/gl/src/gl/Config.cpp
@@ -74,6 +74,10 @@ static void* getGlProcessAddress(const char *namez) {
 
 #else
 
+
+typedef Bool ( *PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC) (int attribute, unsigned int *value);
+PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC QueryCurrentRendererIntegerMESA;
+
 static void* getGlProcessAddress(const char *namez) {
     return (void*)glXGetProcAddressARB((const GLubyte*)namez);
 }
@@ -90,6 +94,10 @@ void gl::initModuleGl() {
         wglGetSwapIntervalEXT = (PFNWGLGETSWAPINTERVALEXTPROC)getGlProcessAddress("wglGetSwapIntervalEXT");
         wglChoosePixelFormatARB = (PFNWGLCHOOSEPIXELFORMATARBPROC)getGlProcessAddress("wglChoosePixelFormatARB");
         wglCreateContextAttribsARB = (PFNWGLCREATECONTEXTATTRIBSARBPROC)getGlProcessAddress("wglCreateContextAttribsARB");
+#endif
+
+#if defined(Q_OS_LINUX)
+        QueryCurrentRendererIntegerMESA = (PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC)getGlProcessAddress("glXQueryCurrentRendererIntegerMESA");
 #endif
 
 #if defined(USE_GLES)
@@ -123,4 +131,15 @@ void gl::setSwapInterval(int interval) {
 #else
     Q_UNUSED(interval);
 #endif
+}
+
+bool gl::queryCurrentRendererIntegerMESA(int attr, unsigned int *value) {
+    #if defined(Q_OS_LINUX)
+    if (QueryCurrentRendererIntegerMESA) {
+        return QueryCurrentRendererIntegerMESA(attr, value);
+    }
+    #endif
+
+    *value = 0;
+    return false;
 }
--- a/libraries/gl/src/gl/Config.cpp
+++ b/libraries/gl/src/gl/Config.cpp
@@ -75,7 +75,7 @@ static void* getGlProcessAddress(const char *namez) {
 #else
 
 
-typedef Bool ( *PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC) (int attribute, unsigned int *value);
+typedef Bool (*PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC) (int attribute, unsigned int *value);
 PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC QueryCurrentRendererIntegerMESA;
 
 static void* getGlProcessAddress(const char *namez) {
--- a/libraries/gl/src/gl/Config.h
+++ b/libraries/gl/src/gl/Config.h
@@ -52,7 +52,7 @@ namespace gl {
     void initModuleGl();
     int getSwapInterval();
     void setSwapInterval(int swapInterval);
-    bool  queryCurrentRendererIntegerMESA(int attr, unsigned int *value);
+    bool queryCurrentRendererIntegerMESA(int attr, unsigned int *value);
 }
 
 #endif // hifi_gpu_GPUConfig_h
--- a/libraries/gl/src/gl/Config.h
+++ b/libraries/gl/src/gl/Config.h
@@ -52,6 +52,7 @@ namespace gl {
     void initModuleGl();
     int getSwapInterval();
     void setSwapInterval(int swapInterval);
+    bool  queryCurrentRendererIntegerMESA(int attr, unsigned int *value);
 }
 
 #endif // hifi_gpu_GPUConfig_h
--- a/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
@@ -122,9 +122,9 @@ GLint GLBackend::GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX{ 0 };
 GLint GLBackend::GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX{ 0 };
 GLint GLBackend::TEXTURE_FREE_MEMORY_ATI{ 0 };
 
-size_t GLBackend::_total_memory{ 0 };
-size_t GLBackend::_dedicated_memory{ 0 };
-GLBackend::VideoCardType GLBackend::_video_card{ GLBackend::Unknown };
+size_t GLBackend::_totalMemory{ 0 };
+size_t GLBackend::_dedicatedMemory{ 0 };
+GLBackend::VideoCardType GLBackend::_videoCard{ GLBackend::Unknown };
 
 
 #define GLX_RENDERER_VIDEO_MEMORY_MESA 0x8187
@@ -148,7 +148,7 @@ void GLBackend::init() {
         GL_GET_INTEGER(MAX_UNIFORM_BLOCK_SIZE);
         GL_GET_INTEGER(UNIFORM_BUFFER_OFFSET_ALIGNMENT);
 
-        if ( vendor.contains("NVIDIA") ) {
+        if (vendor.contains("NVIDIA") ) {
             qCDebug(gpugllogging) << "NVIDIA card detected";
             GL_GET_INTEGER(GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX);
             GL_GET_INTEGER(GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX);
@@ -157,44 +157,48 @@ void GLBackend::init() {
             qCDebug(gpugllogging) << "GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX: " << GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX;
             qCDebug(gpugllogging) << "GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX: " << GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX;
             qCDebug(gpugllogging) << "GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX: " << GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX;
-            qCDebug(gpugllogging) << "sz: " << sizeof(_total_memory);
+            qCDebug(gpugllogging) << "sz: " << sizeof(_totalMemory);
 
-            _total_memory = GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX * BYTES_PER_KIB;
-            _dedicated_memory = GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX * BYTES_PER_KIB;
-            _video_card = NVIDIA;
+            _totalMemory = GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX * BYTES_PER_KIB;
+            _dedicatedMemory = GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX * BYTES_PER_KIB;
+            _videoCard = NVIDIA;
 
 
-        } else if ( vendor.contains("ATI")) {
+        } else if (vendor.contains("ATI")) {
             qCDebug(gpugllogging) << "ATI card detected";
             GL_GET_INTEGER(TEXTURE_FREE_MEMORY_ATI);
 
-            _total_memory = TEXTURE_FREE_MEMORY_ATI * BYTES_PER_KIB;
-            _dedicated_memory = _total_memory;
-            _video_card = ATI;
-        } else if ( vendor.contains("Intel")) {
+            _totalMemory = TEXTURE_FREE_MEMORY_ATI * BYTES_PER_KIB;
+            _dedicatedMemory = _totalMemory;
+            _videoCard = ATI;
+        } else if (vendor.contains("Intel")) {
             unsigned int mem;
 
            if ( ::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem) ) {
-                _total_memory = mem * BYTES_PER_MIB;
-                _dedicated_memory = _total_memory;
+                _totalMemory = mem * BYTES_PER_MIB;
+                _dedicatedMemory = _totalMemory;
+                _videoCard = Intel;
             }
         } else {
             qCCritical(gpugllogging) << "Don't know how to get memory for OpenGL vendor " << vendor;
-            _video_card = Unknown;
-            _dedicated_memory = 0;
-            _total_memory = 0;
+            _videoCard = Unknown;
+            _dedicatedMemory = 0;
+            _totalMemory = 0;
         }
 
-        qCDebug(gpugllogging) << "dedicated: " << _dedicated_memory;
-        qCDebug(gpugllogging) << "total: " << _total_memory;
+        qCDebug(gpugllogging) << "dedicated: " << _dedicatedMemory;
+        qCDebug(gpugllogging) << "total: " << _totalMemory;
 
 
         LOG_GL_CONTEXT_INFO(gpugllogging, contextInfo);
+        GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer);
 
         // From here on, GPUIdent::getInstance()->getMumble() should efficiently give the same answers.
         qCDebug(gpugllogging) << "GPU:";
-        qCDebug(gpugllogging) << "\ttotal memory:" << (_total_memory / BYTES_PER_KIB) << "KB";
-        qCDebug(gpugllogging) << "\tdedicated memory:" << (_dedicated_memory / BYTES_PER_KIB) << "KB";
+        qCDebug(gpugllogging) << "\tcard:" << gpu->getName();
+        qCDebug(gpugllogging) << "\tdriver:" << gpu->getDriver();
+        qCDebug(gpugllogging) << "\ttotal memory:" << (_totalMemory / BYTES_PER_KIB) << "KB";
+        qCDebug(gpugllogging) << "\tdedicated memory:" << (_dedicatedMemory / BYTES_PER_KIB) << "KB";
         qCDebug(gpugllogging) << "\tavailable memory:" << (getAvailableMemory()  / BYTES_PER_KIB) << "KB";
         qCDebug(gpugllogging) << "Limits:";
         qCDebug(gpugllogging) << "\tmax textures:" << MAX_TEXTURE_IMAGE_UNITS;
@@ -212,13 +216,15 @@ void GLBackend::init() {
 size_t GLBackend::getAvailableMemory() {
     GLint mem;
 
-    switch( _video_card ) {
+    switch( _videoCard ) {
         case NVIDIA:
             glGetIntegerv(GL_GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX, &mem);
             return mem * BYTES_PER_KIB;
         case ATI:
             glGetIntegerv(GL_TEXTURE_FREE_MEMORY_ATI, &mem);
             return mem * BYTES_PER_KIB;
+        case Intel:
+            return 0; // Don't know the current value
         case Unknown:
             break;
     }
--- a/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
@@ -10,11 +10,13 @@
 //
 #include "GLBackend.h"
 
+
 #include <mutex>
 #include <queue>
 #include <list>
 #include <functional>
 #include <glm/gtc/type_ptr.hpp>
+#include "gl/Config.h"
 
 #if defined(NSIGHT_FOUND)
 #include "nvToolsExt.h"
@@ -105,13 +107,27 @@ GLBackend::CommandCall GLBackend::_commandCalls[Batch::NUM_COMMANDS] =
 };
 
 #define GL_GET_INTEGER(NAME) glGetIntegerv(GL_##NAME, &const_cast<GLint&>(NAME)); 
-    
+
+#define BYTES_PER_KIB 1024L
+#define BYTES_PER_MIB (1024L * BYTES_PER_KIB)
+
 GLint GLBackend::MAX_TEXTURE_IMAGE_UNITS{ 0 };
 GLint GLBackend::MAX_UNIFORM_BUFFER_BINDINGS{ 0 };
 GLint GLBackend::MAX_COMBINED_UNIFORM_BLOCKS{ 0 };
 GLint GLBackend::MAX_COMBINED_TEXTURE_IMAGE_UNITS{ 0 };
 GLint GLBackend::MAX_UNIFORM_BLOCK_SIZE{ 0 };
 GLint GLBackend::UNIFORM_BUFFER_OFFSET_ALIGNMENT{ 1 };
+GLint GLBackend::GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX{ 0 };
+GLint GLBackend::GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX{ 0 };
+GLint GLBackend::GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX{ 0 };
+GLint GLBackend::TEXTURE_FREE_MEMORY_ATI{ 0 };
+
+size_t GLBackend::_total_memory{ 0 };
+size_t GLBackend::_dedicated_memory{ 0 };
+GLBackend::VideoCardType GLBackend::_video_card{ GLBackend::Unknown };
+
+
+#define GLX_RENDERER_VIDEO_MEMORY_MESA 0x8187
 
 void GLBackend::init() {
     static std::once_flag once;
@@ -132,13 +148,54 @@ void GLBackend::init() {
         GL_GET_INTEGER(MAX_UNIFORM_BLOCK_SIZE);
         GL_GET_INTEGER(UNIFORM_BUFFER_OFFSET_ALIGNMENT);
 
+        if ( vendor.contains("NVIDIA") ) {
+            qCDebug(gpugllogging) << "NVIDIA card detected";
+            GL_GET_INTEGER(GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX);
+            GL_GET_INTEGER(GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX);
+            GL_GET_INTEGER(GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX);
+
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX: " << GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX: " << GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX: " << GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "sz: " << sizeof(_total_memory);
+
+            _total_memory = GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX * BYTES_PER_KIB;
+            _dedicated_memory = GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX * BYTES_PER_KIB;
+            _video_card = NVIDIA;
+
+
+        } else if ( vendor.contains("ATI")) {
+            qCDebug(gpugllogging) << "ATI card detected";
+            GL_GET_INTEGER(TEXTURE_FREE_MEMORY_ATI);
+
+            _total_memory = TEXTURE_FREE_MEMORY_ATI * BYTES_PER_KIB;
+            _dedicated_memory = _total_memory;
+            _video_card = ATI;
+        } else if ( vendor.contains("Intel")) {
+            unsigned int mem;
+
+           if ( ::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem) ) {
+                _total_memory = mem * BYTES_PER_MIB;
+                _dedicated_memory = _total_memory;
+            }
+        } else {
+            qCCritical(gpugllogging) << "Don't know how to get memory for OpenGL vendor " << vendor;
+            _video_card = Unknown;
+            _dedicated_memory = 0;
+            _total_memory = 0;
+        }
+
+        qCDebug(gpugllogging) << "dedicated: " << _dedicated_memory;
+        qCDebug(gpugllogging) << "total: " << _total_memory;
+
+
         LOG_GL_CONTEXT_INFO(gpugllogging, contextInfo);
-        GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer); 
+
         // From here on, GPUIdent::getInstance()->getMumble() should efficiently give the same answers.
         qCDebug(gpugllogging) << "GPU:";
-        qCDebug(gpugllogging) << "\tcard:" << gpu->getName();
-        qCDebug(gpugllogging) << "\tdriver:" << gpu->getDriver();
-        qCDebug(gpugllogging) << "\tdedicated memory:" << gpu->getMemory() << "MB";
+        qCDebug(gpugllogging) << "\ttotal memory:" << (_total_memory / BYTES_PER_KIB) << "KB";
+        qCDebug(gpugllogging) << "\tdedicated memory:" << (_dedicated_memory / BYTES_PER_KIB) << "KB";
+        qCDebug(gpugllogging) << "\tavailable memory:" << (getAvailableMemory()  / BYTES_PER_KIB) << "KB";
         qCDebug(gpugllogging) << "Limits:";
         qCDebug(gpugllogging) << "\tmax textures:" << MAX_TEXTURE_IMAGE_UNITS;
         qCDebug(gpugllogging) << "\tmax texture binding:" << MAX_COMBINED_TEXTURE_IMAGE_UNITS;
@@ -150,6 +207,24 @@ void GLBackend::init() {
         qCDebug(gpugllogging, "V-Sync is %s\n", (::gl::getSwapInterval() > 0 ? "ON" : "OFF"));
 #endif
     });
+}
+
+size_t GLBackend::getAvailableMemory() {
+    GLint mem;
+
+    switch( _video_card ) {
+        case NVIDIA:
+            glGetIntegerv(GL_GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX, &mem);
+            return mem * BYTES_PER_KIB;
+        case ATI:
+            glGetIntegerv(GL_TEXTURE_FREE_MEMORY_ATI, &mem);
+            return mem * BYTES_PER_KIB;
+        case Unknown:
+            break;
+    }
+
+    return 0;
+
 }
 
 GLBackend::GLBackend(bool syncCache) {
--- a/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
@@ -149,6 +149,7 @@ void GLBackend::init() {
         GL_GET_INTEGER(UNIFORM_BUFFER_OFFSET_ALIGNMENT);
 
         GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer);
+        unsigned int mem;
 
         if (vendor.contains("NVIDIA") ) {
             qCDebug(gpugllogging) << "NVIDIA card detected";
@@ -173,21 +174,15 @@ void GLBackend::init() {
             _totalMemory = TEXTURE_FREE_MEMORY_ATI * BYTES_PER_KIB;
             _dedicatedMemory = _totalMemory;
             _videoCard = ATI;
-        } else if (vendor.contains("Intel")) {
-            unsigned int mem;
-
-           if ( ::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem) ) {
+        } else if ( ::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem) ) {
+                // This works only on Linux. queryCurrentRendererIntegerMESA will return false if the
+                // function is not supported because we're not on Linux, or for any other reason.
+                qCDebug(gpugllogging) << "MESA card detected";
                 _totalMemory = mem * BYTES_PER_MIB;
                 _dedicatedMemory = _totalMemory;
-                _videoCard = Intel;
-            } else {
-                qCWarning(gpugllogging) << "Intel card on non-Linux system, trying GPUIdent fallback";
-                _videoCard = Unknown;
-                _dedicatedMemory = gpu->getMemory();
-                _totalMemory = _dedicatedMemory;
-            }
+                _videoCard = MESA;
         } else {
-            qCCritical(gpugllogging) << "Don't know how to get memory for OpenGL vendor " << vendor;
+            qCCritical(gpugllogging) << "Don't know how to get memory for OpenGL vendor " << vendor << "; renderer " << renderer << ", trying fallback";
             _videoCard = Unknown;
             _dedicatedMemory = gpu->getMemory();
             _totalMemory = _dedicatedMemory;
@@ -230,7 +225,7 @@ size_t GLBackend::getAvailableMemory() {
         case ATI:
             glGetIntegerv(GL_TEXTURE_FREE_MEMORY_ATI, &mem);
             return mem * BYTES_PER_KIB;
-        case Intel:
+        case MESA:
             return 0; // Don't know the current value
         case Unknown:
             break;
@@ -244,7 +239,7 @@ bool GLBackend::availableMemoryKnown() {
     switch( _videoCard ) {
         case NVIDIA: return true;
         case ATI: return true;
-        case Intel: return false;
+        case MESA: return false;
         case Unknown: return false;
     }
 
--- a/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
@@ -148,6 +148,8 @@ void GLBackend::init() {
         GL_GET_INTEGER(MAX_UNIFORM_BLOCK_SIZE);
         GL_GET_INTEGER(UNIFORM_BUFFER_OFFSET_ALIGNMENT);
 
+        GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer);
+
         if (vendor.contains("NVIDIA") ) {
             qCDebug(gpugllogging) << "NVIDIA card detected";
             GL_GET_INTEGER(GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX);
@@ -178,12 +180,17 @@ void GLBackend::init() {
                 _totalMemory = mem * BYTES_PER_MIB;
                 _dedicatedMemory = _totalMemory;
                 _videoCard = Intel;
+            } else {
+                qCWarning(gpugllogging) << "Intel card on non-Linux system, trying GPUIdent fallback";
+                _videoCard = Unknown;
+                _dedicatedMemory = gpu->getMemory();
+                _totalMemory = _dedicatedMemory;
             }
         } else {
             qCCritical(gpugllogging) << "Don't know how to get memory for OpenGL vendor " << vendor;
             _videoCard = Unknown;
-            _dedicatedMemory = 0;
-            _totalMemory = 0;
+            _dedicatedMemory = gpu->getMemory();
+            _totalMemory = _dedicatedMemory;
         }
 
         qCDebug(gpugllogging) << "dedicated: " << _dedicatedMemory;
@@ -191,7 +198,7 @@ void GLBackend::init() {
 
 
         LOG_GL_CONTEXT_INFO(gpugllogging, contextInfo);
-        GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer);
+
 
         // From here on, GPUIdent::getInstance()->getMumble() should efficiently give the same answers.
         qCDebug(gpugllogging) << "GPU:";
@@ -231,6 +238,17 @@ size_t GLBackend::getAvailableMemory() {
 
     return 0;
 
+}
+
+bool GLBackend::availableMemoryKnown() {
+    switch( _videoCard ) {
+        case NVIDIA: return true;
+        case ATI: return true;
+        case Intel: return false;
+        case Unknown: return false;
+    }
+
+    return false;
 }
 
 GLBackend::GLBackend(bool syncCache) {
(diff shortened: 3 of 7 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-c36ae7e_f4f50899.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-c36ae7e.tar_f4f50899.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-c36ae7e.tar_f4f50899.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-c36ae7e_f4f50899.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-c36ae7e_f4f50899.exe

libraries/gl/src/gl/Config.cpp
```diff
@@ -74,6 +74,10 @@ static void* getGlProcessAddress(const char *namez) {
 
 #else
 
+
+typedef Bool ( *PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC) (int attribute, unsigned int *value);
```
HifiExperiments, line 78:
```suggestion
typedef Bool (*PFNGLXQUERYCURRENTRENDERERINTEGERMESAPROC) (int attribute, unsigned int *value);
```

libraries/gl/src/gl/Config.cpp
```diff
@@ -124,3 +132,14 @@ void gl::setSwapInterval(int interval) {
     Q_UNUSED(interval);
 #endif
 }
+
+bool gl::queryCurrentRendererIntegerMESA(int attr, unsigned int *value) {
+    #if defined(Q_OS_LINUX)
```
HifiExperiments, line 137: do you need this ifdef, since you already check if the function pointer below?

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,54 @@ void GLBackend::init() {
         GL_GET_INTEGER(MAX_UNIFORM_BLOCK_SIZE);
         GL_GET_INTEGER(UNIFORM_BUFFER_OFFSET_ALIGNMENT);
 
+        if ( vendor.contains("NVIDIA") ) {
```
HifiExperiments, line 151:
```suggestion
        if (vendor.contains("NVIDIA") ) {
```

HifiExperiments: is the vendor always in all caps?

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,54 @@ void GLBackend::init() {
...
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX: " << GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX: " << GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "sz: " << sizeof(_total_memory);
+
+            _total_memory = GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX * BYTES_PER_KIB;
+            _dedicated_memory = GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX * BYTES_PER_KIB;
+            _video_card = NVIDIA;
+
+
+        } else if ( vendor.contains("ATI")) {
```
HifiExperiments, line 167:
```suggestion
        } else if (vendor.contains("ATI")) {
```

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,54 @@ void GLBackend::init() {
...
+            qCDebug(gpugllogging) << "ATI card detected";
+            GL_GET_INTEGER(TEXTURE_FREE_MEMORY_ATI);
+
+            _total_memory = TEXTURE_FREE_MEMORY_ATI * BYTES_PER_KIB;
+            _dedicated_memory = _total_memory;
+            _video_card = ATI;
+        } else if ( vendor.contains("Intel")) {
```
HifiExperiments, line 174:
```suggestion
        } else if (vendor.contains("Intel")) {
```

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,54 @@ void GLBackend::init() {
...
+            unsigned int mem;
+
+           if ( ::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem) ) {
```
HifiExperiments, line 177:
```suggestion
           if (::gl::queryCurrentRendererIntegerMESA(GLX_RENDERER_VIDEO_MEMORY_MESA, &mem)) {
```

HifiExperiments: can you explain this?  right now it looks like this will only work on linux with Intel GPUs.  why not other OSes?

also should it set _video_card?

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -105,13 +107,27 @@ GLBackend::CommandCall GLBackend::_commandCalls[Batch::NUM_COMMANDS] =
...
 GLint GLBackend::MAX_UNIFORM_BLOCK_SIZE{ 0 };
 GLint GLBackend::UNIFORM_BUFFER_OFFSET_ALIGNMENT{ 1 };
+GLint GLBackend::GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX{ 0 };
+GLint GLBackend::GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX{ 0 };
+GLint GLBackend::GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX{ 0 };
+GLint GLBackend::TEXTURE_FREE_MEMORY_ATI{ 0 };
+
+size_t GLBackend::_total_memory{ 0 };
+size_t GLBackend::_dedicated_memory{ 0 };
+GLBackend::VideoCardType GLBackend::_video_card{ GLBackend::Unknown };
```
HifiExperiments, line 127: these should be camel case, e.g. _videoCard

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,54 @@ void GLBackend::init() {
...
+
+
         LOG_GL_CONTEXT_INFO(gpugllogging, contextInfo);
-        GPUIdent* gpu = GPUIdent::getInstance(vendor, renderer); 
+
         // From here on, GPUIdent::getInstance()->getMumble() should efficiently give the same answers.
         qCDebug(gpugllogging) << "GPU:";
-        qCDebug(gpugllogging) << "\tcard:" << gpu->getName();
-        qCDebug(gpugllogging) << "\tdriver:" << gpu->getDriver();
-        qCDebug(gpugllogging) << "\tdedicated memory:" << gpu->getMemory() << "MB";
```
HifiExperiments, line 141: can we put back these old logs (I guess not the old dedicated memory one?)

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -152,6 +209,24 @@ void GLBackend::init() {
     });
 }
 
+size_t GLBackend::getAvailableMemory() {
```
HifiExperiments, line 212: seems like there's some repetition between this code and the above.  could we DRY it up by getting the video card first, and then calling this method?

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -185,7 +185,11 @@ void GLTextureTransferEngineDefault::updateMemoryPressure() {
 
     size_t allowedMemoryAllocation = gpu::Texture::getAllowedGPUMemoryUsage();
     if (0 == allowedMemoryAllocation) {
-        allowedMemoryAllocation = DEFAULT_ALLOWED_TEXTURE_MEMORY;
+        allowedMemoryAllocation = GLBackend::getTotalMemory();
```
HifiExperiments, line 188: should this be getAvailableMemory?  and if so, what happens below if we actually DO have 0 available memory

JulianGro:
Some issues I found while testing:
-When switching to the automatic texture memory setting I get 
```
[08/07 01:12:42] [DEBUG] [hifi.gpu] New MAX texture memory  0  MB
``` 
in console. (The setting still works)
-`hifi.gpu` is not available as secondary filter in the log viewer and you cannot put one in manually.
-When actually using all the texture memory, the my operating system gets unstable. (whole system freezing when something additional get loaded)

The last issue is a game breaker for me, so I would be against merging this until we got that adressed.
Also if anyone wants to test: My `MotoApartment` domain on explore currently has a ton of 8k images for testing this PR.

libraries/gl/src/gl/Config.cpp
daleglass: The variable that holds the pointer is declared in an #ifdef and may not exist

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
daleglass: I looked, so far I've not seen a way to do this for Intel other than on Linux. I got this from the glxinfo program.

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
daleglass: Fixed

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
daleglass: What do you mean? It performs a different function (available vs total)

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
daleglass: This is slightly unfinished, sorry about that. I rushed a bit to push this out to get some feedback and testing on other systems. This is going to be slightly more complex.

The original code just sets a limit and keeps things under that. So if we have 4 GB VRAM, we try not to exceed that amount of textures.

But systems use VRAM for other things, and some cards like NVidia and ATI let you ask how much memory is free. So if we're doing things automatically, it'd be better to actually ask how much is free, and go by that.

But for Intel so far I can only find the total memory, so we still have to do things the old way.

For now the code is 100% the old way of doing things, but the improvement's coming soon.

daleglass:
@JulianGro 
> Some issues I found while testing:
> -When switching to the automatic texture memory setting I get
> 
> ```
> [08/07 01:12:42] [DEBUG] [hifi.gpu] New MAX texture memory  0  MB
> ```

That should be normal, "auto" works as a limit of 0 MB internally.

> -`hifi.gpu` is not available as secondary filter in the log viewer and you cannot put one in manually.

Which log viewer? I think this should be possible to filter using Qt functionality

> -When actually using all the texture memory, the my operating system gets unstable. (whole system freezing when something additional get loaded)

This is very much a prototype. The first pending thing is track free VRAM and use that when available (can't do that on Intel currently). Also there should probably be some sort of reserve for the OS and such (maybe limit it to 90%?)

> The last issue is a game breaker for me, so I would be against merging this until we got that adressed.

Oh, certainly, this still needs work.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-25cfd9e.tar_051d7c0a.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-25cfd9e.tar_051d7c0a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-25cfd9e_051d7c0a.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-25cfd9e_051d7c0a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-25cfd9e_051d7c0a.exe

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
daleglass: Okay, improvements made in the last commit

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-91bf5c9_a507b0c4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-91bf5c9.tar_a507b0c4.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-91bf5c9_a507b0c4.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-91bf5c9.tar_a507b0c4.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-91bf5c9_a507b0c4.exe

JulianGro:
I just tried a507b0c on my system with 1GB VRAM and it froze while loading the serverless, presumably because the VRAM filled completely.
I cannot really test, as my system completely freezes, but it does feel similar to the vram filling up: cursor moving slow, keyboard inputs becoming unreliable and then everything completely freezing.

daleglass:
@JulianGro Which card? Can you try running glxinfo before and during? For instance, on nvidia I get:

```
    Dedicated video memory: 8192 MB
    Total available memory: 8192 MB
    Currently available dedicated video memory: 5710 MB
```

daleglass:
@JulianGro Please try adding this GLTextureTransfer.cpp, at line 253:

```
    qCDebug(gpugllogging) << "UseAvail" << useAvailableGlMemory <<
        "; Pressure" << pressure <<
        "; Allowed " << allowedMemoryAllocation <<
        "; Variable " << totalVariableMemoryAllocation <<
        "; Free " << GLBackend::getAvailableMemory() << "/" << GLBackend::getTotalMemory() <<
        "; canDemote " << canDemote;
```

JulianGro:
Just noticed that I am actually running nouveau drivers. This is an Nvidia GTX 650 Ti.
glxinfo does not show a lot of useful info. Here is what seems somewhat useful:
```
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: nouveau (0x10de)
    Device: NVE6 (0x11c6)
    Version: 18.3.6
    Accelerated: yes
    Video memory: 994MB
    Unified memory: no
    Preferred profile: core (0x1)
    Max core profile version: 4.3
    Max compat profile version: 4.3
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
OpenGL vendor string: nouveau
OpenGL renderer string: NVE6
OpenGL core profile version string: 4.3 (Core Profile) Mesa 18.3.6
```
@daleglass Should I try your addition to the source code anyways? Nouveau has always been pretty much useless for 3D for me.

daleglass:
@JulianGro Ah crap, I had forgotten about nouveau. I have no idea if this works right with it.

Yes, I'd be very interested to see what happens with that

daleglass:
@JulianGro It's likely it's going through the "else" pathway and just doing the old thing of having a fixed 1 GB limit

daleglass:
@JulianGro Ok, attempt at improvement made. Please test!

JulianGro:
With commit a507b0c and your additional changes my Interface froze this time rather than my full system (though it was close)
Here are that parts of the log that seem relevant:
```
[08/09 01:35:10] [CRITICAL] [hifi.gpu.gl] Don't know how to get memory for OpenGL vendor  "nouveau"
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Version:  
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Shader Language Version:  
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Vendor:  
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Renderer:  
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Version:  4.3 (Core Profile) Mesa 18.3.6
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Shader Language Version:  4.30
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Vendor:  nouveau
[08/09 01:35:10] [DEBUG] [hifi.gl] GL Renderer:  NVE6
[08/09 01:35:10] [CRITICAL] [hifi.gpu.gl] Don't know how to get memory for OpenGL vendor  "nouveau"
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] dedicated:  0
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] total:  0
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] GL Version:  4.3 (Core Profile) Mesa 18.3.6
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] GL Shader Language Version:  4.30
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] GL Vendor:  nouveau
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] GL Renderer:  NVE6
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] GPU:
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	card: ""
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	driver: ""
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	total memory: 0 KB
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	dedicated memory: 0 KB
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	available memory: 0 KB
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] Limits:
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	max textures: 32
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	max texture binding: 192
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	max uniforms: 84
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	max uniform binding: 84
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	max uniform size: 65536
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 	uniform alignment: 256
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] V-Sync is ON
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] 
[08/09 01:35:10] [DEBUG] [hifi.gpu.gl] Using OpenGL 4.1 backend
[08/09 01:35:10] [DEBUG] [hifi.interface] Initialized GL
```

daleglass:
Yup, `Don't know how to get memory for OpenGL vendor  "nouveau"`

So it's not really doing anything for you, it should just be going with 1 GB. Please try the latest addition.

JulianGro:
I think we have very much hit the limits of the nouveau driver: Bildschirmfoto vom 2020-08-09 02-15-04
But while the former auto setting often froze the system, this one only seems to slow the system down for a little time and then freeze the interface (with visible graphical glitches).
```
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Version:  
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Shader Language Version:  
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Vendor:  
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Renderer:  
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Version:  4.3 (Core Profile) Mesa 18.3.6
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Shader Language Version:  4.30
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Vendor:  nouveau
[08/09 02:14:46] [DEBUG] [hifi.gl] GL Renderer:  NVE6
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] MESA card detected
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] dedicated:  1042284544
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] total:  1042284544
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] GL Version:  4.3 (Core Profile) Mesa 18.3.6
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] GL Shader Language Version:  4.30
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] GL Vendor:  nouveau
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] GL Renderer:  NVE6
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] GPU:
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	card: ""
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	driver: ""
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	total memory: 1017856 KB
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	dedicated memory: 1017856 KB
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	available memory: 0 KB
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] Limits:
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	max textures: 32
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	max texture binding: 192
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	max uniforms: 84
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	max uniform binding: 84
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	max uniform size: 65536
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 	uniform alignment: 256
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] V-Sync is ON
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] 
[08/09 02:14:46] [DEBUG] [hifi.gpu.gl] Using OpenGL 4.1 backend
[08/09 02:14:46] [DEBUG] [hifi.interface] Initialized GL
```

According to my experiences with nouveau in the past, I don't think there I much point in troubleshooting this.

daleglass:
@JulianGro the MESA pathway it's now using only provides a way to get the total VRAM present, with no way to tell how much is actually being used. Currently the patch goes with 80% of that. So that could be the reason for the improvement, it's being less aggressive.

Maybe try messing with `MAX_AUTO_FRACTION_OF_TOTAL_MEMORY` in GLTextureTransfer.cpp and see if a lower fraction will perform better. Manually setting VRAM to say, 512MB or lower should do pretty much the same.

daleglass:
Also, please add the logging code and see what it spits out, that may help figure out some usable limit.

JulianGro:
My system just froze without Vircadia running..
This could either be because of Nouveau (wouldn't be a first), or the graphics card is starting to fail because of old age. Nonetheless, I will stop the nouveau experimentation here and install the proprietary drivers again as those worked fine before.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-f035c0c.tar_ea97c577.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-f035c0c.tar_ea97c577.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-f035c0c_ea97c577.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-f035c0c_ea97c577.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-f035c0c_ea97c577.exe

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
daleglass: Update: Changed slightly in the last commit. Now the `queryCurrentRendererIntegerMESA` bit is done always when not NVIDIA or ATI, which should work better with the nouveau driver.

Unfortunately there doesn't seem to be a way to get the free VRAM.

Still Linux specific though, as so far I've not found a way to get this information on other systems. So on Windows/OSX this will fail and go to the last resort of using GPUIdent.

JulianGro:
So the proprietary nvidia drivers work, but it does seem like there might be too little memory left over. The serverless works fine and nvidia-smi reports ~100megabytes for VRAM still left, but when I go to my video memory test world, the whole system locks up. I will send the full log once my system is rebootet.

JulianGro:
Oh it seems like the Alt+F4 actually worked this time.
I ran commit ea97c57 with the logging code.
https://gist.github.com/JulianGro/444d37dafdfb5f58ed657a77242c8cf7

I also feel like the automatic setting should be reported as that and not as "0MB".
Similarly we might want the logging code in there by standard, as there could always be problems with unorthodox hardware.

daleglass:
@JulianGro There's nothing from the logging code in there, are you sure you built with it?

The logging code isn't there by standard because it's going to spit out a LOT of that. I think it makes more sense to try to include it in the stats display instead.

JulianGro:
I did copy the source, edit the code and then point vircadia builder at it.
But it does seem like my vram test world makes the system unstable even on 256 MB memory. So there might be a different issue there.

daleglass:
@JulianGro Oh, that explains it. Vircadia builder resets the source tree on every build. You want to pass `--keep-source` if you really want to do that. The intention of it is to do builds automatically and reliably, not to work as a development environment.

But really, once you ran it once, you can just run `make` from `~/Vircadia/build` and skip vircadia-builder entirely.

JulianGro:
According to nvidia-smi, the interface fills up the vram even with 256 mb of texture memory selected. Maybe it loads the textures into vram to have them downscaled by the graphics card? It looks like it spikes with every big texture being loaded and at some point just cannot keep up anymore.

daleglass:
Mm, it's certainly possible that there's just something broken elsewhere. Can you try building master to see if it behaves any better?

Also, I'd like to see what the logging code says. Here I think it might be pointing at a leak, because my usage seems to slowly go up, but I've not dug into it yet.

JulianGro:
So Vircadia builder seems to override the source even when using --keep-souce. Not sure what is happening there as I did the same thing in the past and it worked. Also I don't see how `make` would work in the build directory, as that directory has been created with the old source?

daleglass:
Ack, really? I'll look into that. 

I mean, `~/Vircadia/source` is just a git repo. You can pull and checkout whatever you want there. 
Then go to `~/Vircadia/build` and `make` there.

But in general, vircadia-builder was never intended to be used for actual development, the idea was making it easy to build stuff for distributions where we don't have packages, or when somebody wants to test a specific branch.

daleglass:
Just tested the flag, works just fine for me...

JulianGro:
Yeah not sure what is up with that.
I did find out what the system instability on the proprietary drivers is about. I set texture memory to 4mb (which seems to be respected as much as possible) and noticed that the system was still hanging even though there was multiple hundred megabytes of free vram. Turns out that the *system* memory is actually being used up when I go into my vram test world.. I assume the engine is trying to downscale all textures at once and fills up all memory with that (which is 8gb ram *and* 8gb swap). I will open an issue for that.

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -218,7 +236,19 @@ void GLTextureTransferEngineDefault::updateMemoryPressure() {
 
     Backend::textureResourceIdealGPUMemSize.set(idealMemoryAllocation);
     size_t unallocated = idealMemoryAllocation - totalVariableMemoryAllocation;
-    float pressure = (float)totalVariableMemoryAllocation / (float)allowedMemoryAllocation;
+    float pressure = 0;
+
+    if ( useAvailableGlMemory ) {
+        float total_mem = GLBackend::getTotalMemory();
+        float avail_mem = GLBackend::getAvailableMemory() - AUTO_RESERVE_TEXTURE_MEMORY;
+        if ( avail_mem < 0 ) {
```
HifiExperiments, line 244:
```suggestion
        if (avail_mem < 0) {
```

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -218,7 +236,19 @@ void GLTextureTransferEngineDefault::updateMemoryPressure() {
 
     Backend::textureResourceIdealGPUMemSize.set(idealMemoryAllocation);
     size_t unallocated = idealMemoryAllocation - totalVariableMemoryAllocation;
-    float pressure = (float)totalVariableMemoryAllocation / (float)allowedMemoryAllocation;
+    float pressure = 0;
+
+    if ( useAvailableGlMemory ) {
+        float total_mem = GLBackend::getTotalMemory();
```
HifiExperiments, line 242: camel case

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
HifiExperiments: also getAvailableMemory returns a size_t so this can never be negative

daleglass: The issue is the  `- AUTO_RESERVE_TEXTURE_MEMORY;` bit on line 243. My thought is that we should ensure some memory is always left unallocated, because something else on the system might need some.

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,60 @@ void GLBackend::init() {
...
+        if (vendor.contains("NVIDIA") ) {
+            qCDebug(gpugllogging) << "NVIDIA card detected";
+            GL_GET_INTEGER(GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX);
+            GL_GET_INTEGER(GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX);
+            GL_GET_INTEGER(GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX);
+
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX: " << GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX: " << GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX: " << GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "sz: " << sizeof(_totalMemory);
```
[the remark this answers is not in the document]
daleglass: Accidentally snuck in when I was trying to figure out why I was getting a weird result. Turned out I needed an 'L' suffix in the constant.

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -218,7 +236,19 @@ void GLTextureTransferEngineDefault::updateMemoryPressure() {
...
+    if (useAvailableGlMemory) {
+        float totalMem = GLBackend::getTotalMemory();
+        float availMem = GLBackend::getAvailableMemory() - AUTO_RESERVE_TEXTURE_MEMORY;
+        if (availMem < 0) {
+            availMem = 0;
+        }
+
+        pressure = (totalMem - availMem) / totalMem;
+    } else {
+        pressure = (float)totalVariableMemoryAllocation / (float)allowedMemoryAllocation;
```
HifiExperiments, line 250: can either of the denominators (totalMem or allowedMemoryAllocation) be 0?  should we detect that and do something else?  otherwise looks good

daleglass: allowedMemoryAllocation is dealt with early on starting on line 191. If zero (this signals an "auto" setting), it sets useAvailableGlMemory only if the backend says we know how much free VRAM we have, and in that case there is no fixed allowed allocation, we just keep watch on free memory. totalMem is guaranteed to be known, because if we know how much is free, we definitely know the total.

If free memory is not known (eg, Intel), then we pick 80% of the maximum as a target (idea being this is an Intel card, and so probably nothing else is running. This is just a guess, but I don't see any better options).

If nothing is known, fallback to the old system of assuming a 1 GB limit.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-2bd3da7_2314b6fb.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-2bd3da7.tar_2314b6fb.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-2bd3da7.tar_2314b6fb.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-2bd3da7_2314b6fb.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-2bd3da7_2314b6fb.exe

kasenvr:
Found one issue, as I max out my GPU memory by running Firestorm and Vircadia, nametags do not show. Closing Firestorm allows them to be shown.

So my question is this, does it have a set minimum? Automatic should have a minimum of 1GB perhaps? Setting the max texture memory to 1GB does allow nametags to be shown, for example.

daleglass:
There's no minimum, no. It tries to fit in dedicated VRAM. Exceeding that likely carries a significant performance penalty (I've not measured it though).

The goal here is picking a reasonable setting without user intervention. If you have 8 GB free, it'll use that. If you are running other things and have 512MB free, it'll try to squeeze into that. Formerly the default was 1 GB in both cases.

kasenvr:
Well since everyone gets set to auto by default, can it be a minimum of 1GB, or 512MB, whatever it takes to ensure the basic experience has a minimum maintained (e.g. nametags work?). If someone wants to give it a go, max their texture memory with another app and then set to auto and see if they get nametags on auto?

daleglass:
I think nametags are probably not entirely fixable that way. What if you're in a domain that has 4 GB of textures, wouldn't that still affect nametags even with the minimum? I think for things like that we should consider some sort of texture priority system, where you can specify how important it is to keep a texture at full res.

kasenvr:
Completely agree. I'm just wondering why it works then with 1GB hard cap instead of auto? I sense auto is just going way too low for some reason... So figured for now a 1GB minimum would serve the purpose of ensuring a minimum experience? But I'm not sure the exact cause, someone else can give it a try and see if they see similar behavior perhaps? Maybe 512MB is enough, not entirely sure.

I think something that prioritizes would of course be good but probably not something we can do easily at the moment.

Settings: High
Max Texture: Auto
image

Settings: High
Max Texture: 1GB
image

SilverfishVR:
On my laptop with a gtx920m (2GB) "auto" looks very blurry, like selecting 4MB, selecting 1 or 2GB looks way better and performs well

"auto" on PR599-2bd3da7:
Skærmbillede 2020-08-14 00 11 16

"auto" on 2020.2.4RC1 (it's what was on there):
Skærmbillede 2020-08-14 00 12 56

daleglass:
@SilverfishVR What video card, how much total VRAM, and can you find out the free amount at the time when you're running?

Also, a run with the logging patch above would be very appreciated.

libraries/gl/src/gl/Config.h
```diff
@@ -52,6 +52,7 @@ namespace gl {
     void initModuleGl();
     int getSwapInterval();
     void setSwapInterval(int swapInterval);
+    bool queryCurrentRendererIntegerMESA(int attr, unsigned int *value);
```
ctrlaltdavid, line 55:
```suggestion
    bool queryCurrentRendererIntegerMESA(int attr, unsigned int* value);
```

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -105,13 +107,27 @@ GLBackend::CommandCall GLBackend::_commandCalls[Batch::NUM_COMMANDS] =
 };
 
 #define GL_GET_INTEGER(NAME) glGetIntegerv(GL_##NAME, &const_cast<GLint&>(NAME)); 
-    
+
+#define BYTES_PER_KIB 1024L
+#define BYTES_PER_MIB (1024L * BYTES_PER_KIB)
```
ctrlaltdavid, line 112: I suggest using `BYTES_PER_KILOBYTE` and `BYTES_PER_MEGABYTE`. The rest of the codebase has "kilobyte" and "megabyte" spelt out, and it's not clear what a "KIB" or "MIB" is (e.g, whether these are special things).

Or use/update NumericalContstants.h.

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -132,13 +148,59 @@ void GLBackend::init() {
...
+
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX: " << GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX: " << GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX;
+            qCDebug(gpugllogging) << "GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX: " << GPU_MEMORY_INFO_CURRENT_AVAILABLE_VIDMEM_NVX;
+
+            _totalMemory = GPU_MEMORY_INFO_TOTAL_AVAILABLE_MEMORY_NVX * BYTES_PER_KIB;
+            _dedicatedMemory = GPU_MEMORY_INFO_DEDICATED_VIDMEM_NVX * BYTES_PER_KIB;
+            _videoCard = NVIDIA;
+
+
```
ctrlaltdavid, line 168:
```suggestion
```

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -218,7 +236,19 @@ void GLTextureTransferEngineDefault::updateMemoryPressure() {
 
     Backend::textureResourceIdealGPUMemSize.set(idealMemoryAllocation);
     size_t unallocated = idealMemoryAllocation - totalVariableMemoryAllocation;
-    float pressure = (float)totalVariableMemoryAllocation / (float)allowedMemoryAllocation;
+    float pressure = 0;
+
+    if (useAvailableGlMemory) {
+        float totalMem = GLBackend::getTotalMemory();
+        float availMem = GLBackend::getAvailableMemory() - AUTO_RESERVE_TEXTURE_MEMORY;
```
ctrlaltdavid, line 243: Why the `- AUTO_RESERVER_TEXUTRE_MEMORY` / why is the 64MB value correct? Perhaps add a comment?

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
JulianGro: KiB and MiB are traditional Kilobytes and Megabytes. This is a common thing in the Linux world, as a lot of software (like windows) uses traditional and decimal interchangeably without the end user seeing that.

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
daleglass: In that branch, we're trying to just available memory. The 64M is there because I'm supposing completely filling it might be undesirable (eg, the desktop UI or a web browser might need some).

64M is kind of a randomly picked value. My logic was that on the low end you get 1 GB-ish, and ideally little should be left unused.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-6cf8f6e_abbace87.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-6cf8f6e.tar_abbace87.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-6cf8f6e.tar_abbace87.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-6cf8f6e_abbace87.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-6cf8f6e_abbace87.exe

JulianGro:
I have retested on my GTX 650 ti (1gb) machine and am not sure what to make of my findings:
The used video memory before launching the interface decreases every time I run it.
~246 MiB before the first test, ~202 MiB before the second test, 184 MiB before the third test. That is with opening additinal application in between like the gnome-system-monitor and the xed text editor.
In the first test I started the interface in the serverless and walked around, changed window size, looked around until the VRAM usage did not increase anymore at 645 MiB.
When starting the interface in the hub, the VRAM usage stabilizes around 870 MiB without moving around or anything, just starting the interface in that location.
When looking around the video memory usage goes up to 978 MiB (of 979 MiB total video memory) and stays close to that border. Once the video memory get filled up like that, the framerate drops to ~6 fps and the LOD goes to minimum while seemingly not improving performance or freeing up system ressources. The LOD then continues to go up and down.
Once the video memory filled up, the RAM usage makes a good jump and slowly but steadily increases. I stopped the test on ~9GiB of memory usage (on a machine with only 8gb of real system memory).
The GPU Texture memory statistics in the interface speak of 1497 mb allocated and populated memory.

If I set the texture memory to the old standard which is 1gb, the memory fills up very fast, but the framerate stays more stable, only dropping down to 18 fps. The RAM also doesn't fill up nearly as fast. RAM usage seemingly stopping to increase at 6,4 GiB. The LOD also doesn't go down visibly.
The GPU Texture memory statistics in the interface show 891 mb allocated (1621 mb ideal) and 891 mb populated.
Overall the hard set value of 1 gb feels a lot more stable.

kasenvr:
This one I think needs further attention because if there's a lower amount of VRAM on the computer it will give a subpar experience without gaining back the performance needed. This is a similar effect to the LOD tools where it will wipe out rendered objects and still have little impact on improving the experience.

JulianGro:
@daleglass This is the log when running on the auto setting with the debug code added: https://gist.github.com/JulianGro/39f76d190613b7b2d777a2a3934bb61f
This is the log on 6GB texture memory setting: https://gist.github.com/JulianGro/9ff9d28ac924d72be86c94a18c9c6799
If there is some weird errors in there: the shader thingee in the hub was pretty much "corrupting" all graphics.

daleglass:
@JulianGro Looked at the log, fixed an overflow issue, please test again.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-aa60982.tar_7a186bba.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-aa60982.tar_7a186bba.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-aa60982_7a186bba.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-aa60982_7a186bba.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-aa60982_7a186bba.exe

JulianGro:
I can confirm that it works as expected now. In the new hub I get a slight performance improvement in the sense that there is less frametime spikes.
One thing that is interesting though is the "taking" video memory from other applications. When I run Interface on auto, it keeps 153MiB of video memory free by only using 71MB for textures. This makes a lot of textures very blurry. If I run the 1024MB setting it uses more video memory that is available and pushes some stuff into system memory. If I now switch to auto again, it uses 140MB of video memory for textures and while still leaving 153MiB empty. Textures are looking a lot better at that point.

I feel like there must be some way we could use this. Then again, it might be to risky on system with low system memory or less intelligent video drivers. And this problem might already be solved when someone uses a 1,5GB card instead of a 1GB one.
I do actually have the same GPU with 2GB instead of 1GB laying around, so I could test that as well.

TLDR: works as expected on GTX 650 Ti with Nvidia 450.80.02 proprietary on Linux.

daleglass:
@JulianGro Currently it's trying to reserve 64MB VRAM, I could get rid of that. My testing so far suggests this should be harmless.

Overall though I think that even if graphically suboptimal this should be a pretty good setting in that it should maximize performance, which is probably a good default on low end hardware, and on high end hardware it'll make better use of the available VRAM, so it makes better use of high end hardware's capabilities. The user can always switch to a fixed allocation if they want to ignore that.

daleglass:
@JulianGro Please test the latest revision, which should leave text entities untouched.

@ctrlaltdavid or @HifiExperiments: Am I on the right track here? What I think is happening is that text entities get made by having a font get all its glyphs rendered to a texture, and then text entities are assembled by creating geometry that is textured with the right bits of that font texture. So protecting that one texture from demotion ensures all text entities that use it will remain crisp.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-3f64897_32843133.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-3f64897.tar_32843133.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-3f64897.tar_32843133.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-3f64897_32843133.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR599/Vircadia-Alpha-PR599-3f64897_32843133.exe

HifiExperiments:
that’s a great catch and sounds reasonable to me

ctrlaltdavid:
Looks good to me.

JulianGro:
I can confirm this working on text entities.
On 7a186bb:
vircadia-snap-by--on-2020-10-27_22-27-00

On the new 3284313:
vircadia-snap-by--on-2020-10-27_22-23-41

kasenvr (approved):
Seems to be working pretty well for me on Windows 10. :)

vegaslon:
On my Radeon card this is reporting "[11/18 09:15:04] [DEBUG] [hifi.gpu] New MAX texture memory  0  MB" when I turn on automatic texture memory.

JulianGro:
@vegaslon That is fine. The debug message is arguably misleading, but 0 MB is the automatic setting. It should work as expected.

vegaslon:
Still using this PR for me makes the world blurry, have to change my max to at least 512 to for it to stop being blurry.

JulianGro:
@vegaslon What graphics card with how much video memory do you have?
Now that you report this issue, I am not sure if this was tested on AMD actually. Personally I only tested it on a low end and a high end Nvidia card.

vegaslon:
GPU - AMD Radeon RX 5700 XT - Primary/Discrete
VRAM - 8176 MB - GDDR6 1750 MHz
Graphics Card Manufacturer - Built by AMD
Graphics Chipset - AMD Radeon RX 5700 XT
Device ID - 731F
Vendor ID - 1002
SubSystem ID - 0B36
SubSystem Vendor ID - 1002
Revision ID - C1
Bus Type - PCI Express 4.0
Current Bus Settings - PCI Express 4.0 x16
BIOS Version - 017.001.000.042
BIOS Part Number - 113-D1820501-101
BIOS Date - 2019/06/16 22:38
Usable Memory Size - 8176 MB
Memory Type - GDDR6
Memory Clock - 1750 MHz
Core Clock - 1755 MHz
Total Memory Bandwidth - 448 GByte/s
Memory Bit Rate - 14.00 Gbps
2D Driver File Path - /REGISTRY/MACHINE/SYSTEM/CurrentControlSet/Control/Class/{4d36e968-e325-11ce-bfc1-08002be10318}/0000
OpenGL® API Version - 4.6
OpenCL™ API Version - 2.0

JulianGro:
@vegaslon Can you open the Statistics by pressing / or pressing the button in the development menu, press on the statistics to make them bigger and send a screenshot of that on the auto setting? Ideally while looking at a texture that is blurry.
Unfortunately I don't currently have an AMD card to test it myself.

vegaslon:
https://i.gyazo.com/d655f7096296bc9edd857a685130df06.jpg

JulianGro:
Yeah that is pretty bad

Issue #575, opened 2020-07-27
Feature Request - copying an entity with .001 naming scheme names the new entity .002, and so on...


Aitolda:
This is handy when you went to drag/copy multiple entities.  For example entities set to render with zone.  In my test.  I have zone.001, box.001, Material.001 which are all parented to the hiarchy.  If I drag and copy the new entities will have the same name as the original currently.

HifiExperiments:
I think one danger here is that if there are attached scripts that rely on things being named specific names, those will break on the duplicated objects

Aitolda:
> I think one danger here is that if there are attached scripts that rely on things being named specific names, those will break on the duplicated objects

Hmm...  I wonder what might be a good alternative.

kasenvr:
If it's liable to be useful to a great majority of people, simple optional parameters/controls would do.

Aitolda:
> If it's liable to be useful to a great majority of people, simple optional parameters/controls would do.

The main thing is to be able to visually distinguish different models in the entity list when duplicates are made.  What I proposed is how it works in blender, but HFExperiments brought up a valid point.  I feel something still needs to indicate these differences, but I guess the question is which issue will we run into more, and is there an alternative?

AleziaKurdis:
> I think one danger here is that if there are attached scripts that rely on things being named specific names, those will break on the duplicated objects

After duplication, a script can also be in problem because there are 2 entities with the same name. Duplicates can always generate such undesired consequences, with or without a name change.
To have used duplicate often, I can tell you that  it's annoying to have to check the id to figure which one is the duplicate (to avoid issues with parenting for example)
Plus, you can always rename it directly in the list to remove the duplication mark, which is already easier than now, where you need to figure which one is the copy in order to rename it.

HifiExperiments:
that’s a great point.  in that case, I think this should be solved in the create script itself, and I’d vote for append “(number)“ instead of trying to detect a specific naming scheme and modifying it.

so “Box” would have a duplicate “Box (2)” which would have a duplicate “Box (3)”

if there’s an option to toggle the behavior within the create UI, even better

Aitolda:
> that’s a great point. in that case, I think this should be solved in the create script itself, and I’d vote for appending “(number)“ instead of trying to detect a specific naming scheme and modifying it.
> 
> so “Box” would have a duplicate “Box (2)” which would have a duplicate “Box (3)”
> 
> if there’s an option to toggle the behavior within the create UI, even better

[...]

PR #603, opened 2020-08-07
Incremental suffix to cloned entities's name (issue #575)

AleziaKurdis:
This enhancement adds an incremental suffix to the name 
of the cloned entities resulting from "Duplicate" and "Paste".
The goal is to help to figure which entities is the coned one.
This addresses issue kasenvr#575


scripts/system/create/entitySelectionTool/entitySelectionTool.js
```diff
@@ -31,6 +33,55 @@ function deepCopy(v) {
...
+    var rippedOriginalName = removeNameAppendice(originalName);
+
+    do {
+        duplicateNumber++;
+                if (rippedOriginalName === "") {
+                    duplicateName = "(" + duplicateNumber + ")";
+                } else {
+                    duplicateName = rippedOriginalName + " (" + duplicateNumber + ")";
+                }
+                existingNames = Entities.findEntitiesByName(duplicateName, Vec3.ZERO, ENTIRE_DOMAIN_SCAN_RADIUS, false);
```
HifiExperiments, line 49: I think this is at the core of why I don’t love this solution.  1. it’s not a good idea to call this method in a loop (it can be expensive, it’s blocking, and this loop could be huge), 2. what if you’ve got Box, Box (2), and Box (3), and you delete Box (2), and then copy Box again?

AleziaKurdis: What other solution do we have?
a sequence per name kept for the session?

Or simply not doing this functionality.
For sure, this can rapidly become an annoying feature when used for repetitive cloning.

HifiExperiments: keeping track of it per name won’t work because the script could be reloaded or stopped

like I said in the issue, I would vote to have Box always become Box (2), even if you clone it multiple times

AleziaKurdis:
Just pushed a simpler solution without any hazardous uniqueness check.
Cloned entities will systematically get the suffix " (2)", unless their name has already a suffix. 
In that case, we simply increment the suffix.

scripts/system/create/entitySelectionTool/entitySelectionTool.js
```diff
@@ -485,6 +528,7 @@ SelectionManager = (function() {
             } else {
                 delete properties.position;
             }
+            properties.name = getDuplicateAppendedName(properties.name);
```
HifiExperiments, line 531: what's the difference between pasting and duplicating?  do we want only duplicating to do this new logic?

scripts/system/create/entitySelectionTool/entitySelectionTool.js
```diff
@@ -31,6 +32,46 @@ function deepCopy(v) {
...
+                        rippedName = name.substr(0, i-1);
+                        existingSequence = parseInt(sequenceReader);
+                        break;
+                    } else {
+                        sequenceReader = currentChar + sequenceReader;
+                    }
+                }
+            }
+        }
+    }
```
HifiExperiments, line 65: this logic *seems* sound to me, but it's a little hard to follow, which means it will also be a little hard to debug if something ever goes wrong.

I wonder if it would be easier to use regexs?  I don't know what JS regex support looks like

scripts/system/create/entitySelectionTool/entitySelectionTool.js
AleziaKurdis: Duplicate is Move + CTRL... generating a clone of the entity. (this one is right I think)

But just realize that it was not a good implementation for "Paste" (when you do Copy > Paste) because you might also have done "Cut > Paste" and this is going to be wrong in that second case. 
We don't want a cut-paste to increment...  (I would need to check that.)

I envision that more for the "Duplicate" since it's what I do 100% of the time when I use this.
It's probably not too difficult to address Copy versus Cut.

scripts/system/create/entitySelectionTool/entitySelectionTool.js
AleziaKurdis: I've simply used something I trust... regex is just not intuitive "for me" and I don't know what surprise it could bring.
I will check that option.

scripts/system/create/entitySelectionTool/entitySelectionTool.js
AleziaKurdis: I will check for that

scripts/system/create/entitySelectionTool/entitySelectionTool.js
HifiExperiments: yeah not super important, just wanted to mention it in case it would be easier

scripts/system/create/entitySelectionTool/entitySelectionTool.js
HifiExperiments: could you also copy from another domain and paste in a new one?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-024980e_91ebd378.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-024980e_91ebd378.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-024980e.tar_91ebd378.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-024980e.tar_91ebd378.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-024980e_91ebd378.exe

scripts/system/create/entitySelectionTool/entitySelectionTool.js
AleziaKurdis: "Duplicate" makes sense to get the suffix (and have it incremented)

"Paste" doesn't. It's not possible for the script to be sure of the origin of what is in the clipboard.
It would not make sense to add the suffix if you paste something copied on another domain.
It would not  make sense to add the suffix to something we cut and then paste (to let say move it somewhere else  in the same or in another domain)
we can paste the cut object N time. If we can track this for the current run, it's impossible the second we close and reopen the creat app.

I will adjust the fix to have this only for "Duplicate".
No suffix for pasted entities.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-ee5de58.tar_2cd1fc84.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-ee5de58.tar_2cd1fc84.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-ee5de58_2cd1fc84.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-ee5de58_2cd1fc84.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR603/Vircadia-Alpha-PR603-ee5de58_2cd1fc84.dmg

kasenvr:
image
Appears to be working as expected for me. :)

PR #613, opened 2020-08-09
[enhancement] transition listener registration from member string name to member string pointer

odysseus654:
Qt permits signals to be connected to slots using either a method pointer or the string name of the method.  This page describes the strengths of both: https://doc.qt.io/qt-5/signalsandslots-syntaxes.html

Registering interest in specific packet types is done through registering a listener with PacketReceiver.  Up to this point all registrations have been made by string, which required digging into the Qt reflection layer to confirm the method was compatible and indirectly invoke the method.

This patch changes all listener registrations to use functors referencing member pointers.  This enforces the existence and parameters of the method at compile-time rather than run-time.  There may be a small performance benefit from the reduced processing required, but the focus here is in getting away from using strings to identify members.

### DIFF
--- a/domain-server/src/DomainGatekeeper.cpp
+++ b/domain-server/src/DomainGatekeeper.cpp
@@ -17,7 +17,8 @@
 #include <openssl/x509.h>
 #include <random>
 
-#include <QDataStream>
+#include <QtCore/QDataStream>
+#include <QtCore/QMetaMethod>
 
 #include <AccountManager.h>
 #include <Assignment.h>
--- a/domain-server/src/DomainServerSettingsManager.h
+++ b/domain-server/src/DomainServerSettingsManager.h
@@ -211,6 +211,8 @@ private:
 
     /// guard read/write access from multiple threads to settings 
     QReadWriteLock _settingsLock { QReadWriteLock::Recursive };
+
+    friend class DomainServer;
 };
 
 #endif // hifi_DomainServerSettingsManager_h
--- a/interface/src/ui/DomainConnectionModel.cpp
+++ b/interface/src/ui/DomainConnectionModel.cpp
@@ -9,7 +9,8 @@
 //
 
 #include "DomainConnectionModel.h"
-#include <QLoggingCategory>
+#include <QtCore/QLoggingCategory>
+#include <QtCore/QMetaMethod>
 
 #include <NodeList.h>
 #include <NumericalConstants.h>
--- a/libraries/networking/src/PacketReceiver.h
+++ b/libraries/networking/src/PacketReceiver.h
@@ -16,8 +16,6 @@
 #include <vector>
 #include <unordered_map>
 
-#include <QtCore/QMap>
-#include <QtCore/QMetaMethod>
 #include <QtCore/QMutex>
 #include <QtCore/QObject>
 #include <QtCore/QPointer>
@@ -29,6 +27,7 @@
 #include "udt/PacketHeaders.h"
 
 class EntityEditPacketSender;
+class Node;
 class OctreePacketProcessor;
 
 namespace std {
@@ -43,6 +42,22 @@ namespace std {
 class PacketReceiver : public QObject {
     Q_OBJECT
 public:
+    class ListenerReference {
+    public:
+        virtual bool invokeDirectly(const QSharedPointer<ReceivedMessage>& receivedMessagePointer, const QSharedPointer<Node>& sourceNode) = 0;
+        bool invokeWithQt(const QSharedPointer<ReceivedMessage>& receivedMessagePointer, const QSharedPointer<Node>& sourceNode);
+        virtual bool isSourced() const = 0;
+        virtual QObject* getObject() const = 0;
+    };
+    typedef QSharedPointer<ListenerReference> ListenerReferencePointer;
+
+    template<class T>
+    static ListenerReferencePointer makeUnsourcedListenerReference(T* target, void (T::*slot)(QSharedPointer<ReceivedMessage>));
+
+    template <class T>
+    static ListenerReferencePointer makeSourcedListenerReference(T* target, void (T::*slot)(QSharedPointer<ReceivedMessage>, QSharedPointer<Node>));
+
+public:
     using PacketTypeList = std::vector<PacketType>;
     
     PacketReceiver(QObject* parent = 0);
@@ -55,8 +70,8 @@ public:
     // If deliverPending is false, ReceivedMessage will only be delivered once all packets for the message have
     // been received. If deliverPending is true, ReceivedMessage will be delivered as soon as the first packet
     // for the message is received.
-    bool registerListener(PacketType type, QObject* listener, const char* slot, bool deliverPending = false);
-    bool registerListenerForTypes(PacketTypeList types, QObject* listener, const char* slot);
+    bool registerListener(PacketType type, const ListenerReferencePointer& listener, bool deliverPending = false);
+    bool registerListenerForTypes(PacketTypeList types, const ListenerReferencePointer& listener);
     void unregisterListener(QObject* listener);
     
     void handleVerifiedPacket(std::unique_ptr<udt::Packet> packet);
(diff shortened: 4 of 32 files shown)


libraries/networking/src/ResourceCache.cpp
```diff
@@ -16,8 +16,9 @@
 #include <cmath>
 #include <assert.h>
 
-#include <QThread>
-#include <QTimer>
+#include <QtCore/QMetaMethod>
```
HifiExperiments, line 19: same here

libraries/networking/src/PacketReceiver.h
```diff
@@ -93,4 +135,42 @@ class PacketReceiver : public QObject {
     friend class OctreePacketProcessor;
 };
 
+template <class T>
+PacketReceiver::ListenerReferencePointer PacketReceiver::makeUnsourcedListenerReference(T* target, void (T::* slot)(QSharedPointer<ReceivedMessage>)) {
+    return QSharedPointer<UnsourcedListenerReference<T>>::create(target, slot);
```
HifiExperiments, line 140: when I first saw all these shared pointers, I was a little nervous, but under the hood, this code is still doing exactly what the old code was doing, right?  the old code was just doing it in a silly way with strings.  the number of shared pointers + basically all the logic is pretty much staying the same?

domain-server/src/DomainGatekeeper.cpp
```diff
@@ -17,7 +17,8 @@
 #include <openssl/x509.h>
 #include <random>
 
-#include <QDataStream>
+#include <QtCore/QDataStream>
+#include <QtCore/QMetaMethod>
```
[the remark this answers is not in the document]
odysseus654: QMetaMethod as an include was removed from PacketReceiver.h.  That broke some files that included PacketReceiver.h, most likely making references to QMetaEnum, so #includes were added "downstream".

interface/src/ui/DomainConnectionModel.cpp
```diff
@@ -9,7 +9,8 @@
 //
 
 #include "DomainConnectionModel.h"
-#include <QLoggingCategory>
+#include <QtCore/QLoggingCategory>
+#include <QtCore/QMetaMethod>
```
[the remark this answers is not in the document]
odysseus654: QMetaMethod as an include was removed from PacketReceiver.h.  That broke some files that included PacketReceiver.h, most likely making references to QMetaEnum, so #includes were added "downstream".

libraries/networking/src/ResourceCache.cpp
odysseus654: QMetaMethod as an include was removed from PacketReceiver.h.  That broke some files that included PacketReceiver.h, most likely making references to QMetaEnum, so #includes were added "downstream".

domain-server/src/DomainServerSettingsManager.h
```diff
@@ -211,6 +211,8 @@ private slots:
 
     /// guard read/write access from multiple threads to settings 
     QReadWriteLock _settingsLock { QReadWriteLock::Recursive };
+
+    friend class DomainServer;
```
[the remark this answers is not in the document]
odysseus654: DomainServer is registering three private slots on this class.  When the slots were being identified by string through Qt reflection this wasn't a problem, but now that we're making direct member references it needs to have access.  So I figure either declare it as a friend or make the slots no longer private.  Suggestions are welcome, I don't really like either approach.

libraries/networking/src/PacketReceiver.h
odysseus654: Not quite sure I understand here, I am likely creating one new shared pointer representing the functor that is being used to call the listener, templated in order to properly use member function references (I believe Qt gets around this with something inside Q_OBJECT), and with two subclassed alternatives to handle the two types of function call parameters.  I attempted to approximate the old code as much as I obviously could, the differences should be about the difference between the two QObject::connect() syntaxes.

odysseus654: As I am using subclassed functors they cannot be passed by value (tried that in an earlier version of this code) and pass-by-reference starts running into lifetime issues, thus using shared pointers.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-8fa9667_34e3d9dd.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-8fa9667.tar_34e3d9dd.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-8fa9667_34e3d9dd.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-8fa9667.tar_34e3d9dd.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-8fa9667_34e3d9dd.exe

JulianGro:
How would we test this? Just see if it builds and runs?

odysseus654:
At this point... probably.  Just make sure this code is at least as good as the code currently in the system.

kasenvr:
This appears to touch quite a few parts of the system, especially the server. So that would have to be tested.

odysseus654:
It touches every part of the system that receives and processes network packets... which is a good chunk of the system.  Hopefully it has zero meaningful effect.  A failure would be indicated by a system suddenly not being able to hear or react to a networking signal.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-e714de8_34e3d9dd.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-e714de8_34e3d9dd.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-e714de8.tar_34e3d9dd.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-e714de8.tar_34e3d9dd.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-e714de8_34e3d9dd.exe

kasenvr:
Test case: Do things networking related and confirm nothing really changes. :)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-d4cffb3.tar_34e3d9dd.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-d4cffb3.tar_34e3d9dd.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-d4cffb3_34e3d9dd.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-d4cffb3_34e3d9dd.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR613/Vircadia-Alpha-PR613-d4cffb3_34e3d9dd.exe

kasenvr:
Did entity updates, tried talking to myself from multiple clients, more or less everything seems fine.

kasenvr:
Also, there are missing copyright headers for the top of every file. Just need to append under the High Fidelity copyrights:

`Copyright 2020 Vircadia contributors.`

kasenvr (approved):
Upgrading my prior testing approval to full approval after hearing that this is a pretty safe PR. :)

PR #621, opened 2020-08-15
Support blendshapes on glTF models without needing FST files

ctrlaltdavid:
Blendshapes were already supported in glTF files provided you use an FST file to map them. This PR adds support for blendshapes in glTF files without needing an FST file, provided you use the standard blendshape names.

### DIFF
--- a/libraries/fbx/src/FBXSerializer.cpp
+++ b/libraries/fbx/src/FBXSerializer.cpp
@@ -1675,5 +1675,152 @@ HFMModel::Pointer FBXSerializer::read(const hifi::ByteArray& data, const hifi::V
 
     // FBXSerializer's mapping parameter supports the bool "deduplicateIndices," which is passed into FBXSerializer::extractMesh as "deduplicate"
 
-    return HFMModel::Pointer(extractHFMModel(mapping, url.toString()));
+    auto hfmModel = extractHFMModel(mapping, url.toString());
+
+    //hfmDebugDump(*hfmModel);
+
+    return HFMModel::Pointer(hfmModel);
+}
+
+void FBXSerializer::hfmDebugDump(const HFMModel& hfmModel) {
+    qCDebug(modelformat) << "---------------- hfmModel ----------------";
+    qCDebug(modelformat) << "  originalURL =" << hfmModel.originalURL;
+
+    qCDebug(modelformat) << "  hasSkeletonJoints =" << hfmModel.hasSkeletonJoints;
+    qCDebug(modelformat) << "  offset =" << hfmModel.offset;
+
+    qCDebug(modelformat) << "  neckPivot = " << hfmModel.neckPivot;
+
+    qCDebug(modelformat) << "  bindExtents.size() = " << hfmModel.bindExtents.size();
+    qCDebug(modelformat) << "  meshExtents.size() = " << hfmModel.meshExtents.size();
+
+    qCDebug(modelformat) << "  jointIndices.size() =" << hfmModel.jointIndices.size();
+    qCDebug(modelformat) << "  joints.count() =" << hfmModel.joints.count();
+    qCDebug(modelformat) << "---------------- Meshes ----------------";
+    qCDebug(modelformat) << "  meshes.count() =" << hfmModel.meshes.count();
+    qCDebug(modelformat) << "  blendshapeChannelNames = " << hfmModel.blendshapeChannelNames;
+    foreach(HFMMesh mesh, hfmModel.meshes) {
+        qCDebug(modelformat) << "\n";
+        qCDebug(modelformat) << "    meshpointer =" << mesh._mesh.get();
+        qCDebug(modelformat) << "    meshindex =" << mesh.meshIndex;
+        qCDebug(modelformat) << "    vertices.count() =" << mesh.vertices.size();
+        qCDebug(modelformat) << "    colors.count() =" << mesh.colors.count();
+        qCDebug(modelformat) << "    normals.count() =" << mesh.normals.size();
+        qCDebug(modelformat) << "    tangents.count() =" << mesh.tangents.size();
+        qCDebug(modelformat) << "    colors.count() =" << mesh.colors.count();
+        qCDebug(modelformat) << "    texCoords.count() =" << mesh.texCoords.count();
+        qCDebug(modelformat) << "    texCoords1.count() =" << mesh.texCoords1.count();
+        qCDebug(modelformat) << "    clusterIndices.count() =" << mesh.clusterIndices.count();
+        qCDebug(modelformat) << "    clusterWeights.count() =" << mesh.clusterWeights.count();
+        qCDebug(modelformat) << "    modelTransform =" << mesh.modelTransform;
+        qCDebug(modelformat) << "    parts.count() =" << mesh.parts.count();
+        qCDebug(modelformat) << "---------------- Meshes (blendshapes)--------";
+        foreach(HFMBlendshape bshape, mesh.blendshapes) {
+            qCDebug(modelformat) << "\n";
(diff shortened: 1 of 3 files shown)


ctrlaltdavid:
This PR needs testing. In particular, there is a potential issue regarding the scaling of blendshapes in glTF files. The original glTF blendshape code scaled blendshape values by `0.1` which made a test glTF avatar's in the original HiFi PR work. However, scaling by `0.1` made blendshapes not work as expected in the test cube models provided in the linked issue 495.

The scale used in the PR code is currently `1.0`. This can be modified by setting the value of a `DEFAULT_GLTF_WEIGHT` environment variable. For example, on Windows, to scale blendshapes by `0.1` per the original code, in a command prompt ...
```
set DEFAULT_GLTF_WEIGHT=0.1
```
... and run Interface from the command prompt.

Testing is needed for the following 8 scenarios:
- FBX entity - with and without FST
- glTF/GLB entity - ""
- FBX avatar - ""
- glTF/GLB avatar - ""

To check, for each:
- Does the model display as expected?
- Do the blendshapes work as expected?
- What blendshape scale makes the blendshapes work as expected?

Note: The user docs on blendshapes aren't up to date. See: https://github.com/kasenvr/vircadia-docs-sphinx/issues/58 for more info.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR621/Vircadia-Alpha-PR621-2ae7558.tar_073898b9.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR621/Vircadia-Alpha-PR621-2ae7558.tar_073898b9.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR621/Vircadia-Alpha-PR621-2ae7558_073898b9.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR621/Vircadia-Alpha-PR621-2ae7558_073898b9.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR621/Vircadia-Alpha-PR621-2ae7558_073898b9.exe

Colorfingers:
Sorry... seems I need to learn more about GIT.  Here's my experience with this PR...  Updated comments:  Here's my experience with this PR... The eyeblink and jawopen morphs are working but are extremely exaggerated. I tried loading it with the fst but that only crashes my interface. Does anyone have a script that can run to test out all of the existing blend shapes? Here are my source files if anyone wants to take a stab at playing with it in this PR. If someone can look at my fst file and tell me what I may be doing wrong that would be great. I tried it several ways with limiting it without the bone maps.... https://hapticmonkey.s3-us-west-1.amazonaws.com/avatar.fst https://hapticmonkey.s3-us-west-1.amazonaws.com/avatar.glb

I changed the max value to 0.001 in Blender for all three which is the lowest I can set the morphs. That did not work. You can see that in https://hapticmonkey.s3-us-west-1.amazonaws.com/avatar2.glb the amount of deformation has not changed. The original for JawOpen was 0.5 and both Eyeblinks were set to a max of 1. I don't think the problem is in the value but in something else...

ctrlaltdavid:
People have glTF avatars working with `set DEFAULT_GLTF_WEIGHT=0.1`. 
glTF non-avatar entities TBD.

SilverfishVR:
Testing PR621-2ae7558

I have used this for both model and avatar test:
testmodel
the vertical bar moves with a shapekey named `Slider`, the horizontal bar moves with a shapekey named `UserShapekey1`, both are 0-1 scale.

files are here https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/index.html?prefix=ShapekeyTest/

I did not change set `DEFAULT_GLTF_WEIGHT` so it should be 1.

Tested scenarios:
FBX model loaded directly
FBX model loaded by FST
FBX avatar loaded directly
FBX avatar loaded by FST
GLB model loaded directly
GLB model loaded by FST
GLB avatar loaded directly
GLB avatar loaded by FST

the FST maps `Slider` to `ÙserBlendshape0` like this: `bs = UserBlendshape0 = Slider = 1`
I also tested a scaling of 0.5 in all FST cases.

Overall, all cases worked as expected 👍, with a few hickups such as a few times, all blendshapes stopped working until I reloaded content, and if I visit my sandbox now in 2020.2.4 it will crash if I turn around and look at the FST loaded GLB model 🤔 

I have not really looked at the "engine driven" blendshapes (though they seem to work on this one) only driving user blendshapes with `MyAvatar.setBlendshape("UserBlendshape0", [VALUE]);` for avatars and `Entities.editEntity("[UUID]", {blendshapeCoefficients: JSON.stringify({"UserBlendshape0": [VALUE]})})` for models so if there is issues with scaling it may be in the part that drives mouth/eye movement

gLTF not tested yet

SilverfishVR:
if I test the "animatedMorphCube" from the Khronos samples the blendshapes don't show:
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/khronos_cube.fst

but if I open it in blender, and then export it, without changing anything it works 🤔 
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/khronos_cube_blenderExport.fst

the model does have keyframe animated blendshapes, which we don't support so maybe that plays a role

SilverfishVR:
testing the shapekeys on Mark_K (JawOpen, EyeBlink_L/R mapped to the sliders) as an avatar, this is what I get

all shapekeys are 0-1 range, scaling in FST's are 1.

directly from fbx: OK, jaw slider never goes above 50%

directly from GLB: new shapekeys for sliders seem ok, or maybe just too fast to see but EyeBlink_R does look to be overdriven, jaw slider never goes above 50%

directly frm GLTF: same as GLB.

FBX from FST: same as direct load.

GLB from FST: same as direct load.

GLTF from FST: same as direct load.

I also tested Hap's avatars aswell as some of own (known to work as fbx) avatars and the shapekeys are indeed all messed up, in different ways, but never right..

Going back to testing on models, I found something curious, not all verticies are affected equally:
Screenshot 2020-08-26 03 19 15 
Screenshot 2020-08-26 03 20 06

the plane in front of the mouth move forward on JawOpen to make the mouth flash when talking and all verticies are weighted 100% to the head bone, and nothing else, but on the gltf model it still gets warped, I don't think it's an issue with how much the blendshapes are driven, I think some verticies just don't end up where they are suposed to, almost like they have a different reference frame.

I also tried a version of Mark Avatar where I remade all the blendshapes, but with non-standard names and then remapped them in the fst. from what I can tell, the jaw and eyeblink shapes are wrong, but the UserBlendshape0 and 1 are not, neither are BrowsU_L BrowsU_R and BrowsU_C, this is the avatar used:
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/Mark_K_Duplicate_shapekeys.fst

I also tested a "minimal" avatar with only "JawOpen" blendshape", and just like the sliders on Mark avatar, it seems to work correctly, as avatar and as entity and JawOpen seems to max out at about 50%, this is also true for fbx and I think it "normal" behaviour (as in all avatars are like that) I'm confused.
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/cheeseHead/cheeseHead.fbx
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/cheeseHead/cheeseHead.glb
https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapekeyTest/cheeseHead/cheeseHead.gltf
max jawOpen (max set to 1) is top triangle tilted about 45 degrees

tl:dr I'm more confused now than before testing, I think there is some weirdness going on where some verts just don't end up where they are supposed to go

PR #631, opened 2020-08-20
Test moving to qt image error handling in avatar app

vegaslon:
An attempt to resolve avatar app slowdown some users are experiencing, going to first rule out xmlrequest.

### DIFF
--- a/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
+++ b/interface/resources/qml/hifi/avatarapp/AvatarsModel.qml
@@ -37,17 +37,6 @@ ListModel {
         return trimmedUrl;
     }
 	
-    function imageExists(imageUrl) {
-
-        var http = new XMLHttpRequest();
-
-        http.open('HEAD', imageUrl, false);
-        http.send();
-
-        return http.status !== 404;
-
-    }
-
     function makeThumbnailUrl(avatarUrl) {
         var marketId = extractMarketId(avatarUrl);
         if (marketId !== '') {
@@ -55,11 +44,6 @@ ListModel {
         }
         
         var avatarThumbnailFileUrl = trimFileExtension(avatarUrl) + ".jpg";
-        var thumbnailExist = imageExists(avatarThumbnailFileUrl);
-        
-        if (!thumbnailExist) {
-            return '';
-        }
         
         return avatarThumbnailFileUrl;
     }


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR631/Vircadia-Alpha-PR631-784429f_c6fbe08a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR631/Vircadia-Alpha-PR631-784429f_c6fbe08a.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR631/Vircadia-Alpha-PR631-784429f.tar_c6fbe08a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR631/Vircadia-Alpha-PR631-784429f.tar_c6fbe08a.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR631/Vircadia-Alpha-PR631-784429f_c6fbe08a.exe

languagelovr:
Tried the Windows build. No more lag for me. Had 2 custom avatars bookmarked. Also tried the current release for comparison and it lagged. This PR build seems to have fixed whatever that issue is/was.

ctrlaltdavid (approved):

ctrlaltdavid:
Tested the Windows build and it fixes the avatar app slow-down for me. Reviewed code and checked program log - just some "file not found"s logged.
```
[08/21 08:26:37] [WARNING] [default] qrc:/qml/hifi/avatarapp/RoundImage.qml:12:5: QML Image: Error transferring http://ctrlaltstudio.com/downloads/hifi/avatars/robot-02/robot-02.jpg - server replied: Not Found
```
Note though that this contains the avatar's URL. URLs have largely been scrubbed from the program log, so is this a problem?

Note also: If further speed-up is wanted, the `Image` control in \interface\resources\qml\hifi\avatarapp\RoundImage.qml could perhaps be set to asynchronous download - https://doc.qt.io/qt-5/qml-qtquick-image.html .

vegaslon:
For network requests that is asychronous by default. For the avatar app they can just grab the url by wearing the avatar no need to look at logs at all.

ctrlaltdavid:
Ah, yes, network requests for `Image` control are asynchronous by default - good!

daleglass:
Error message seems to come from Qt code, so to hide the URL would either need to filter the log or to avoid the error in the first place.

David says: "Check whether the image exists before setting the value - similar to how the original code functioned but without the synchronous blocking"

JulianGro:
> Note though that this contains the avatar's URL. URLs have largely been scrubbed from the program log, so is this a problem?

I feel like if we don't have avatar URLs in the logs otherwise, we shouldn't log this. This can make "stealing" avatars easier and there is a lot of people who are very paranoid about their avatar security. Of course we cannot offer much in terms of avatar security, but muting a log message like this, which doesn't seem very helpful, feels like a little thing we can do to make it just a little bit harder for people to just take someone else's avatar.

vegaslon:
The need for this url is kind of in a bad place to check for the image before setting it. Since there is issues with the warning that people do not have a thumbnail I would suggest just reverting the thumbnail code completely for now until we can build a new avatar app.

kasenvr:
Lovely, well for now though... I suppose it's a question of this for now:

Have a bit of log spam every time you open the Avatar app **or** remove this feature for the moment?

JulianGro:
I feel like the log spam is fine, as it only happens with the small number of avatars that fit into one page of the avatar page at a time. (at least that is my understanding)

We should at some point change the way thumbnails are handled, but I don't really know how we would do that without some external data like an fst file.

kasenvr:
Well we want to solve the freezing then we'll have to make a decision on this for the moment. :)

PR #645, opened 2020-08-24
Add LOD to graphic menu

vegaslon:
https://i.gyazo.com/6084b00b7cde1489148fc8b60e2cd80d.png

### DIFF
--- a/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
+++ b/interface/resources/qml/hifi/dialogs/graphics/GraphicsSettings.qml
@@ -21,6 +21,9 @@ import PerformanceEnums 1.0
 Item {
     HifiStylesUit.HifiConstants { id: hifi; }
 
+    readonly property real defaultMaxVisibilityDistance: 400.0
+    readonly property real unitElementMaxExtent: Math.sqrt(3.0) * 0.5
+
     id: root;
     anchors.fill: parent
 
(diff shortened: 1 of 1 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR645/Vircadia-Alpha-PR645-b78a58c_06b0545e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR645/Vircadia-Alpha-PR645-b78a58c.tar_06b0545e.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR645/Vircadia-Alpha-PR645-b78a58c.tar_06b0545e.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR645/Vircadia-Alpha-PR645-b78a58c_06b0545e.exe

daleglass:
Comments during meeting: the LOD setting may not be working correctly, and exposing it may not be desirable.

vegaslon:
Closing until such time properly talked about.

kasenvr:
@daleglass is correct since what we can do/remain for now is let the LOD be set to max manually to prevent it from affecting anyone. So I apologize for recommending this as an interesting thing to throw in, I was going off the idea that it was/would be already enabled for people and people would need to be able to disable it somehow. However, that is not the case, though I haven't installed a fresh build in a while to confirm.

One thing that this code can be converted for might be maximum texture memory, if not a radial button group. What do you think might be more ideal for something like that? A dropdown menu would also work, potentially.

Either way, many of the boxes in the Developer -> Render menu do need to be brought over such as: shadows, ambient occlusion, etc. as those are used features that would do well to be toggleable.

PR #657, opened 2020-08-25
Fix build on qt 5.14

daleglass:
This fixes the build on Qt 5.14 and above. Tested on 5.14.2 and 5.15.

Building under new Qt versions generates a lot of deprecation warnings, those will be addressed later.

So far 5.14.2 seems to work fine, and 5.15 seems to hang shortly after start for an unknown reason. The end goal here is to get 5.15 to work, since that should be the first version Vircadia can use without patching.

### DIFF
--- a/assignment-client/src/audio/AudioMixerClientData.h
+++ b/assignment-client/src/audio/AudioMixerClientData.h
@@ -14,7 +14,10 @@
 
 #include <queue>
 
+#if !defined(Q_MOC_RUN)
+// Work around https://bugreports.qt.io/browse/QTBUG-80990
 #include <tbb/concurrent_vector.h>
+#endif
 
 #include <QtCore/QJsonObject>
 
--- a/assignment-client/src/audio/AudioMixerSlave.h
+++ b/assignment-client/src/audio/AudioMixerSlave.h
@@ -12,7 +12,10 @@
 #ifndef hifi_AudioMixerSlave_h
 #define hifi_AudioMixerSlave_h
 
+#if !defined(Q_MOC_RUN)
+// Work around https://bugreports.qt.io/browse/QTBUG-80990
 #include <tbb/concurrent_vector.h>
+#endif
 
 #include <AABox.h>
 #include <AudioHRTF.h>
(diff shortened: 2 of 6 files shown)


daleglass:
@HifiExperiments That's a good point.

I've had some difficulty building Qt, including issues like build failures of Qt itself. 5.15 currently seems to have some sort of issue where it just hangs after a while. 

So while I very much want to move to the latest, so far this is potentially troublesome, and I'd rather not break compatibility before this gets some good testing, which could take a good deal of time and effort. In particular, we need to be sure everything keeps working, and that it can be built on all platforms.

So far the #ifdef stuff is very minor, so I think it's a reasonable compromise while we make sure nothing explodes.

So far it looks like at the very least 5.14.2 should work, so I think it could be adopted, if everyone can build it without problems.

HifiExperiments:
my worry here is that we make these changes, but then GHA for example is still using the old version, so most people aren’t seeing the benefits of the update.  if you made all the necessary changes here (for vcpkg, cmake, actual code, etc), then this PR would produce builds that we could have everyone test on more easily, and then once it’s merged, all devs have to do is merge master to reap the benefits

daleglass:
I think this is going to involve multiple PRs, or perhaps the PR will need expanding. This is the bare minimum needed for the build not to explode, and I think is quite harmless, and should allow people to start testing without disturbing anything.

Then there's a huge amount of deprecation warnings coming from Qt now, so those will also need looking at, and may be a fair amount of work, and may be more dangerous.

And then vcpkg and such would be changed after all that is dealt with.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR657/Vircadia-Alpha-PR657-f9fc41a.tar_5327ffee.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR657/Vircadia-Alpha-PR657-f9fc41a_5327ffee.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR657/Vircadia-Alpha-PR657-f9fc41a.tar_5327ffee.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR657/Vircadia-Alpha-PR657-f9fc41a_5327ffee.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR657/Vircadia-Alpha-PR657-f9fc41a_5327ffee.exe

PR #671, opened 2020-08-31
Fix and extend glTF blendshapes support

ctrlaltdavid:
This PR fixes up and extends the support for blendshapes on glTF models - both for entities and avatars, and both with and without using FST files.

Changes include:
- Fix application of glTF blendshapes (on model with one mesh).
- Fix (removes) erroneous weighting of glTF blendshapes by default blendshape values.
- Add support for glTF blendshapes on meshes comprising more than one submesh (glTF primitive).
- Add support for glTF blendshapes on models with more than one mesh.
- Add support for blendshapes in glTF models used directly without an FST. (The model must use standard blendshape names.)
- Blendshapes configured in an FST override blendshapes in the glTF model. (But any in the model that aren't overridden are still used.)
- Add HFM model debug dump for FBX models.

Note: Both .glTF and .glb file types are supported.

### DIFF
--- a/libraries/fbx/src/GLTFSerializer.cpp
+++ b/libraries/fbx/src/GLTFSerializer.cpp
@@ -1567,7 +1567,6 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
                     auto keys = blendshapeIndices.keys();
                     auto values = blendshapeIndices.values();
                     auto names = _file.meshes[node.mesh].extras.targetNames;
-                    QVector<double> weights = _file.meshes[node.mesh].weights;
 
                     for (int weightedIndex = 0; weightedIndex < keys.size(); ++weightedIndex) {
                         float weight = 1.0f;
@@ -1586,13 +1585,6 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
 
                         QVector<glm::vec3> normals;
                         QVector<glm::vec3> vertices;
-
-                        if (weights.size() == primitive.targets.size()) {
-                            int targetWeight = weights[targetIndex];
-                            if (targetWeight != 0) {
-                                weight = weight * targetWeight;
-                            }
-                        }
 
                         if (target.values.contains((QString) "NORMAL")) {
                             generateTargetData(target.values.value((QString) "NORMAL"), weight, normals);
--- a/libraries/fbx/src/GLTFSerializer.cpp
+++ b/libraries/fbx/src/GLTFSerializer.cpp
@@ -1593,8 +1593,9 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
                             generateTargetData(target.values.value((QString) "POSITION"), weight, vertices);
                         }
 
+                        auto blendshapeIndicesCount = blendshape.indices.count();
                         for (int i = 0; i < vertices.size(); i++) {
-                            blendshape.indices.push_back(i);
+                            blendshape.indices.push_back(blendshapeIndicesCount + i);
                             blendshape.vertices.push_back(vertices.value(i));
                             blendshape.normals.push_back(normals.value(i));
                         }
--- a/libraries/fbx/src/GLTFSerializer.cpp
+++ b/libraries/fbx/src/GLTFSerializer.cpp
@@ -1544,25 +1544,33 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
                 // Build morph targets (blend shapes)
                 if (!primitive.targets.isEmpty()) {
 
-                    // Build list of blendshapes from FST
+                    // Build list of blendshapes from FST and model.
                     typedef QPair<int, float> WeightedIndex;
                     hifi::VariantHash blendshapeMappings = mapping.value("bs").toHash();
                     QMultiHash<QString, WeightedIndex> blendshapeIndices;
 
                     for (int i = 0;; ++i) {
-                        hifi::ByteArray blendshapeName = BLENDSHAPE_NAMES[i];
+                        auto blendshapeName = QString(BLENDSHAPE_NAMES[i]);
                         if (blendshapeName.isEmpty()) {
                             break;
                         }
-                        QList<QVariant> mappings = blendshapeMappings.values(blendshapeName);
-                        foreach (const QVariant& mapping, mappings) {
-                            QVariantList blendshapeMapping = mapping.toList();
-                            blendshapeIndices.insert(blendshapeMapping.at(0).toByteArray(), WeightedIndex(i, blendshapeMapping.at(1).toFloat()));
+                        auto mappings = blendshapeMappings.values(blendshapeName);
+                        if (mappings.count() > 0) {
+                            // Use blendshape from mapping.
+                            foreach(const QVariant& mapping, mappings) {
+                                auto blendshapeMapping = mapping.toList();
+                                blendshapeIndices.insert(blendshapeMapping.at(0).toString(), 
+                                    WeightedIndex(i, blendshapeMapping.at(1).toFloat()));
+                            }
+                        } else {
+                            // Use blendshape from model.
+                            if (_file.meshes[node.mesh].extras.targetNames.contains(blendshapeName)) {
+                                blendshapeIndices.insert(blendshapeName, WeightedIndex(i, 1.0f));
+                            }
                         }
                     }
 
-                    // glTF morph targets may or may not have names. if they are labeled, add them based on
-                    // the corresponding names from the FST. otherwise, just add them in the order they are given
+                    // Create blendshapes.
                     mesh.blendshapes.resize((int)Blendshapes::BlendshapeCount);
                     auto keys = blendshapeIndices.keys();
                     auto values = blendshapeIndices.values();
--- a/libraries/fbx/src/GLTFSerializer.cpp
+++ b/libraries/fbx/src/GLTFSerializer.cpp
@@ -1571,7 +1571,9 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
                     }
 
                     // Create blendshapes.
-                    mesh.blendshapes.resize((int)Blendshapes::BlendshapeCount);
+                    if (!blendshapeIndices.isEmpty()) {
+                        mesh.blendshapes.resize((int)Blendshapes::BlendshapeCount);
+                    }
                     auto keys = blendshapeIndices.keys();
                     auto values = blendshapeIndices.values();
                     auto names = _file.meshes[node.mesh].extras.targetNames;
--- a/libraries/fbx/src/GLTFSerializer.cpp
+++ b/libraries/fbx/src/GLTFSerializer.cpp
@@ -1610,12 +1610,14 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash& 
                 }               
             }
 
-            // Add epsilon to mesh extents to compensate for planar meshes
-            mesh.meshExtents.minimum -= glm::vec3(EPSILON, EPSILON, EPSILON);
-            mesh.meshExtents.maximum += glm::vec3(EPSILON, EPSILON, EPSILON);
-            hfmModel.meshExtents.minimum -= glm::vec3(EPSILON, EPSILON, EPSILON);
-            hfmModel.meshExtents.maximum += glm::vec3(EPSILON, EPSILON, EPSILON);
-           
+            // Mesh extents must be at least minimum entity size, in particular for blendshapes to work on planar meshes.
+            const float ENTITY_ITEM_MIN_DIMENSION = 0.001f;
+            auto delta = glm::max(glm::vec3(ENTITY_ITEM_MIN_DIMENSION) - mesh.meshExtents.size(), glm::vec3(0.0f)) / 2.0f;
+            mesh.meshExtents.minimum -= delta;
+            mesh.meshExtents.maximum += delta;
+            hfmModel.meshExtents.minimum -= delta;
+            hfmModel.meshExtents.maximum += delta;
+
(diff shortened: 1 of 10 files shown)


ctrlaltdavid:
### Testing

#### Models

Here are a couple of test models. Other models shouid also be tested.

##### Morph cube

This is a single mesh comprising a single primitive (submesh) with two blendshapes.
The blendshapes are: `UserblendShape0` and `UserBlendshape1`.
One blendshape moves one side of the cube inwards, and the other makes that side of the cube slope.
You can see the expected rendering result of you load the glTF file into this Web page: https://play

Test URLs:
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes.glb
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes-glb.fst
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes.gltf
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes-gltf.fst
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes.fbx
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/MorphCubeUserBlendShapes-fbx.fst

Download zip:
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphCube/morphCube.zip

Example console commands (replace UUID with entity ID):
`
Entities.editEntity("{00000000-0000-0000-0000-000000000000}", { blendshapeCoefficients: JSON.stringify({ "UserBlendshape0": 0.5 })});
Entities.editEntity("{00000000-0000-0000-0000-000000000000}", { blendshapeCoefficients: JSON.stringify({ "UserBlendshape1": 0.5 })});
`

Derived from: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube


##### Morph plane

This is a single mesh comprising two primitives (submeshes), each using a single blendshape.
The blendshape is: `UserBlendshape5`.
In the FST, this blendshape is mapped to `UserBlendshape1`.
The blendshape raises a section of a plane upwards.
You can see the expected rendering result of you load the glTF file into this Web page: https://play

Test URLs:
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphPrimitivesTest/MorphPrimitivesTest-altered.gltf
- http://ctrlaltstudio.com/downloads/vircadia-dev/gltf/morphPrimitivesTest/MorphPrimitivesTest-altered-gltf.fst

Download zip:
- http://ctrlaltstudio.tst/downloads/vircadia-dev/gltf/morphPrimitivesTest/morphPrimitivesTest.zip

Example console command (replace UUID with entity ID):
`
Entities.editEntity("{00000000-0000-0000-0000-000000000000}", { blendshapeCoefficients: JSON.stringify({ "UserBlendshape1": 0.5 })});
Entities.editEntity("{00000000-0000-0000-0000-000000000000}", { blendshapeCoefficients: JSON.stringify({ "UserBlendshape5": 0.5 })});
`
Derived from: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MorphPrimitivesTest


#### Avatars

A variety of avatar models should be tested:
- Both with and without FSTs.
- Both using standard blendshapes in the model file, and using non-standard blendshapes in the model file with mappings specified in an FST.

Avatars should be tested both as entities and as avatars.

Note: One particular problem encountered with some avatar models to date is that when rezzed as entities, though they display at the correct size their entity dimensions are around 1/100th of what they should be. This appears to be a problem in generating the glTF: mesh values come through in metres whereas skeleton values come through in centimetres, and a global scale in the glTF file is `0.01,0.01,0.01` which multiplies the mesh and skeleton values.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-8385e64_22e6ae97.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-8385e64.tar_22e6ae97.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-8385e64.tar_22e6ae97.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-8385e64_22e6ae97.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-8385e64_22e6ae97.exe

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -1536,81 +1536,84 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
-                            if (isNewBlendshape) {
-                                blendshape.vertices.push_back(vertices[i]);
-                                blendshape.normals.push_back(normals[i]);
-                            } else {
-                                blendshape.vertices[count] = blendshape.vertices[count] + vertices[i];
-                                blendshape.normals[count] = blendshape.normals[count] + normals[i];
-                                ++count;
-                            }
+
+                        if (blendshape.indices.size() < prevMeshVerticesCount + vertices.size()) {
```
HifiExperiments, line 1606: why do we need to do this?  we resize it but then only set the things at the end?

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -1621,12 +1624,14 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
             }
 
-            // Add epsilon to mesh extents to compensate for planar meshes
-            mesh.meshExtents.minimum -= glm::vec3(EPSILON, EPSILON, EPSILON);
-            mesh.meshExtents.maximum += glm::vec3(EPSILON, EPSILON, EPSILON);
-            hfmModel.meshExtents.minimum -= glm::vec3(EPSILON, EPSILON, EPSILON);
-            hfmModel.meshExtents.maximum += glm::vec3(EPSILON, EPSILON, EPSILON);
-           
+            // Mesh extents must be at least minimum entity size, in particular for blendshapes to work on planar meshes.
+            const float ENTITY_ITEM_MIN_DIMENSION = 0.001f;
```
HifiExperiments, line 1628: "Entity" isn't really appropriate here since this code is agnostic of how the model is eventually used.

libraries/hfm/src/hfm/HFM.h
```diff
@@ -340,6 +340,8 @@ class Model {
     QMap<int, glm::quat> jointRotationOffsets;
     std::vector<ShapeVertices> shapeVertices;
     FlowData flowData;
+
+    void debugDump();
```
HifiExperiments, line 344: 👍

libraries/fbx/src/GLTFSerializer.cpp
```diff
@@ -1536,81 +1536,84 @@ bool GLTFSerializer::buildGeometry(HFMModel& hfmModel, const hifi::VariantHash&
...
-                    // Build list of blendshapes from FST
+                    // Build list of blendshapes from FST and model.
                     typedef QPair<int, float> WeightedIndex;
                     hifi::VariantHash blendshapeMappings = mapping.value("bs").toHash();
                     QMultiHash<QString, WeightedIndex> blendshapeIndices;
 
                     for (int i = 0;; ++i) {
-                        hifi::ByteArray blendshapeName = FACESHIFT_BLENDSHAPES[i];
+                        auto blendshapeName = QString(BLENDSHAPE_NAMES[i]);
                         if (blendshapeName.isEmpty()) {
```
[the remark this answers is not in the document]
ctrlaltdavid: This `if` statement ends the loop iterating over `BLENDSHAPE_NAMES`.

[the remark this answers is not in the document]
HifiExperiments: oh!  bizarre.

libraries/fbx/src/GLTFSerializer.cpp
ctrlaltdavid: This is handling the case that the glTF mesh comprises two glTF primitives. The primitives are iterated over in the loop...
```
foreach(auto &primitive, _file.meshes[node.mesh].primitives) {
```
Primitive vertices are pushed onto the the back of `mesh.vertices`. We need to set the blendshape data at the same indexes.

libraries/fbx/src/GLTFSerializer.cpp
ctrlaltdavid: Setting the minim dimension to the minimum entity dimension solves the problem of applying blendshapes to planar meshes. Using the original `EPSILON` value results in external code rescaling the mesh to the minimum entity dimension, i.e., multiplying the plane dimension by `ENTITY_ITEM_MIN_DIMENSION / EPSILON = 0.001 / 0.000001f = 1000.0`, wiht the result that the blendshape  distortions are also multiplied by `1000.0` in this direction.
This is the "morph plane" test case.

libraries/fbx/src/GLTFSerializer.cpp
HifiExperiments: right, I guess my question is, can we add prevMeshVerticesCount to the blendshape index without enlarging the blendshape itself?

e.g. remove the resizes + the other uses of prevMeshVerticesCount, except when we set the index to prevMeshVerticesCount + i

libraries/fbx/src/GLTFSerializer.cpp
HifiExperiments: I see.  Can we just rename ENTITY_ITEM_MIN_DIMENSION to something like MODEL_MIN_DIMENSION?

I also wonder how FBX handles this case.  are planar blendshapes similarly broken there?

ctrlaltdavid: I will rename as suggested.
A planar FBX model may very well suffer the same problem. I can't see in the FBX code where planar mesh is handled. I've added an issue to test/fix the FBX case: https://github.com/kasenvr/project-athena/issues/675

libraries/fbx/src/GLTFSerializer.cpp
ctrlaltdavid: Say, a mesh has a blendshape. (Blendshapes are specified at the mesh level.)
Say, the mesh has two primitives (submeshes). Both must target the blendshape.
Primitive 0 is processed. It has 10 vertices, say. So 10 vertices are added to `mesh.vertices` at indexes `0 .. 9` and 10 blendshape target values are added to each of the relevant `blendshape.indices`, `blendshape.vertices`, and `blendshape.normals`, also at indices `0 - 9`.
Primitive 1 is processed. It has 5 vertices, say. So 5 further items are added to each of `mesh.vertices` and the relevant `blendshape.indices` etc., at indices `10 - 14`.

It's not possible to get a primitive (submesh) with a blendshape without preceding primitives also having the same blendshape. And, besides, the `hfm::Model` class doesn't have any facility for specifying the start index for blendshape target data.

We could do away with explicit `resize()` statements and use either `push_back()` to add a new blendshape target value or use `[ ] ` to update an existing blendshape target value, depending on whether or not the blendshape already has a value at the target index. But this seems more complex logic, to me.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-08d20c4_9fc9b203.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-08d20c4_9fc9b203.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-08d20c4.tar_9fc9b203.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-08d20c4.tar_9fc9b203.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR671/Vircadia-Alpha-PR671-08d20c4_9fc9b203.exe

libraries/fbx/src/GLTFSerializer.cpp
HifiExperiments: ok, I *think* I understand.  I'm a little confused because I don't see similar code for FBX, and the old code here didn't seem to do this, but it sounds like you're effectively adding support for "blendshapes on meshes with multiple submeshes" here, and since blendshapes are specified per mesh, it's not like we'll ever unnecessarily resize these to say 15 and then only set the last 5 values; the first 10 values will have already been set from the first submesh pass

Colorfingers:
I have tested this PR and GLTF avatar blendshapes work perfectly. Skeletal animations work as expected.  Great job by ctrlAltDavid!

kasenvr:
If we can get this QA Approved quite soon and merged, I'll see what I can do to put it in for RC2, have a day or two of testing from that, and then it'll be part of this next release! :)

SilverfishVR:
I have not yet had a chance to fully test all scenarios but I have tried a bunch of models and avatars, both directly and loaded from fst, and all except for the morph cube from Khronos works as expected, no more messed up geometry 👍 
Like before, if I open and resave the morph cube in blender it works, 🤷

libraries/fbx/src/GLTFSerializer.cpp
ctrlaltdavid: This is correct. Except that in glTF a mesh is defined as one or more "primitives" (which is where the vertices, for example, reside). The old code only supported blendshapes on a mesh having one primitive (and even then it didn't work properly). It's not really "submeshes" in the FBX sense.

PR #677, opened 2020-09-02
Add option to capture mouse

HifiExperiments:
(Currently not hooked up to any scripts or keys)

To enable, in console: `Camera.captureMouse = true;`

Set to false to disable.

Needs to be tested on other OSes, and not sure how it interacts with the mouse logic in HMD

### DIFF
--- a/interface/resources/controllers/keyboardMouse.json
+++ b/interface/resources/controllers/keyboardMouse.json
@@ -213,12 +213,30 @@
             ]
         },
 
+        { "from": { "makeAxis" : ["Keyboard.MouseMoveLeft", "Keyboard.MouseMoveRight"] },
+          "to": "Actions.DeltaYaw",
+          "when": "Application.CaptureMouse",
+          "filters":
+            [
+                { "type": "scale", "scale": 0.2 }
+            ]
+        },
+
         { "from": { "makeAxis" : ["Keyboard.MouseMoveUp", "Keyboard.MouseMoveDown"] },
           "when": ["!Application.CameraSelfie", "!Application.CameraLookAt", "Keyboard.RightMouseButton"],
           "to": "Actions.DeltaPitch",
           "filters":
             [
                 { "type": "scale", "scale": 0.6 }
+            ]
+        },
+
+        { "from": { "makeAxis" : ["Keyboard.MouseMoveUp", "Keyboard.MouseMoveDown"] },
+          "to": "Actions.DeltaPitch",
+          "when": "Application.CaptureMouse",
+          "filters":
+            [
+                { "type": "scale", "scale": 0.2 }
             ]
         },
 
--- a/interface/resources/controllers/keyboardMouse.json
+++ b/interface/resources/controllers/keyboardMouse.json
@@ -3,9 +3,9 @@
     "channels": [
         { "from": "Keyboard.A", "when": ["Keyboard.RightMouseButton", "!Keyboard.Control"], "to": "Actions.LATERAL_LEFT" },
         { "from": "Keyboard.D", "when": ["Keyboard.RightMouseButton", "!Keyboard.Control"], "to": "Actions.LATERAL_RIGHT" },
-        { "from": "Keyboard.E", "when": ["!Application.CameraSelfie", "!Keyboard.Control"], "to": "Actions.LATERAL_RIGHT" },
-        { "from": "Keyboard.Q", "when": ["!Application.CameraSelfie", "!Keyboard.Control"], "to": "Actions.LATERAL_LEFT" },
+        { "from": "Keyboard.Q", "when": ["!Application.CameraSelfie", "!Keyboard.Control", "!Application.CaptureMouse"], "to": "Actions.LATERAL_LEFT" },
         { "from": "Keyboard.Q", "when": ["Application.CameraSelfie", "!Keyboard.Control"], "to": "Actions.LATERAL_RIGHT" },
+        { "from": "Keyboard.E", "when": ["!Application.CameraSelfie", "!Keyboard.Control", "!Application.CaptureMouse"], "to": "Actions.LATERAL_RIGHT" },
         { "from": "Keyboard.E", "when": ["Application.CameraSelfie", "!Keyboard.Control"], "to": "Actions.LATERAL_LEFT" },
         { "from": "Keyboard.T", "when": "!Keyboard.Control", "to": "Actions.TogglePushToTalk" },
 
@@ -72,46 +72,20 @@
         { "from": { "makeAxis" : [
                 ["Keyboard.Left"],
                 ["Keyboard.Right"]
-             ]
+            ]
           },
-          "when": ["Application.CameraFirstPerson", "!Keyboard.Shift"],
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "!Application.CaptureMouse", "!Keyboard.Shift"],
           "to": "Actions.Yaw"
         },
 
-        { "from": { "makeAxis" : [
-                ["Keyboard.Left"],
-                ["Keyboard.Right"]
-             ]
-          },
-          "when": ["Application.CameraFirstPersonLookat", "!Keyboard.Shift"],
-          "to": "Actions.Yaw"
+        { "from": "Keyboard.Left",
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "Application.CaptureMouse", "!Keyboard.Shift"],
+          "to": "Actions.LATERAL_LEFT"
         },
 
-        { "from": { "makeAxis" : [
-                ["Keyboard.Left"],
-                ["Keyboard.Right"]
-             ]
-          },
-          "when": ["Application.CameraThirdPerson", "!Keyboard.Shift"],
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.Left"],
-                ["Keyboard.Right"]
-             ]
-          },
-          "when": ["Application.CameraLookAt", "!Keyboard.Shift"],
-          "to": "Actions.Yaw"
-        },
-
-        { "from": { "makeAxis" : [
-                ["Keyboard.Left"],
-                ["Keyboard.Right"]
-             ]
-          },
-          "when": ["Application.CameraSelfie", "!Keyboard.Shift"],
-          "to": "Actions.Yaw"
+        { "from": "Keyboard.Right",
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "Application.CaptureMouse", "!Keyboard.Shift"],
+          "to": "Actions.LATERAL_RIGHT"
         },
 
         { "from": { "makeAxis" : [
@@ -119,53 +93,18 @@
                 ["Keyboard.D"]
             ]
           },
-          "when": ["Application.CameraFirstPerson", "!Keyboard.Control"],
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "!Application.CaptureMouse", "!Keyboard.Control"],
           "to": "Actions.Yaw"
         },
 
-        { "from": { "makeAxis" : [
-                ["Keyboard.A"],
-                ["Keyboard.D"]
-            ]
-          },
-          "when": ["Application.CameraFirstPersonLookat", "!Keyboard.Control"],
-          "to": "Actions.Yaw"
+        { "from": "Keyboard.A",
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "Application.CaptureMouse", "!Keyboard.Control"],
+          "to": "Actions.LATERAL_LEFT"
         },
 
-        { "from": { "makeAxis" : [
-                ["Keyboard.A"],
-                ["Keyboard.D"]
-            ]
-          },
-          "when": ["Application.CameraThirdPerson", "!Keyboard.Control"],
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.A"],
-                ["Keyboard.D"]
-             ]
-          },
-          "when": ["Application.CameraLookAt", "!Keyboard.Control"],
-          "to": "Actions.Yaw"
-        },
-
-        { "from": { "makeAxis" : [
-                ["Keyboard.A"],
-                ["Keyboard.D"]
-             ]
-          },
-          "when": ["Application.CameraSelfie", "!Keyboard.Control"],
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.TouchpadLeft"],
-                ["Keyboard.TouchpadRight"]
-            ]
-          },
-          "when": "Application.CameraFirstPerson",
-          "to": "Actions.Yaw"
+        { "from": "Keyboard.D",
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity", "Application.CaptureMouse", "!Keyboard.Control"],
+          "to": "Actions.LATERAL_RIGHT"
         },
 
         { "from": { "makeAxis" : [
@@ -173,39 +112,12 @@
                 ["Keyboard.TouchpadRight"]
             ]
           },
-          "when": "Application.CameraFirstPersonLookat",
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.TouchpadLeft"],
-                ["Keyboard.TouchpadRight"]
-            ]
-          },
-          "when": "Application.CameraThirdPerson",
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.TouchpadLeft"],
-                ["Keyboard.TouchpadRight"]
-            ]
-          },
-          "when": "Application.CameraLookAt",
-          "to": "Actions.Yaw"
-        },
-        
-        { "from": { "makeAxis" : [
-                ["Keyboard.TouchpadLeft"],
-                ["Keyboard.TouchpadRight"]
-            ]
-          },
-          "when": "Application.CameraSelfie",
+          "when": ["!Application.CameraFSM", "!Application.CameraIndependent", "!Application.CameraEntity"],
           "to": "Actions.Yaw"
         },
 
         { "from": { "makeAxis" : ["Keyboard.MouseMoveLeft", "Keyboard.MouseMoveRight"] },
-          "when": "Keyboard.RightMouseButton",
+          "when": ["Keyboard.RightMouseButton", "!Application.CaptureMouse"],
           "to": "Actions.DeltaYaw",
           "filters":
             [
@@ -223,7 +135,7 @@
         },
 
         { "from": { "makeAxis" : ["Keyboard.MouseMoveUp", "Keyboard.MouseMoveDown"] },
-          "when": ["!Application.CameraSelfie", "!Application.CameraLookAt", "Keyboard.RightMouseButton"],
+          "when": ["!Application.CameraSelfie", "!Application.CameraLookAt", "!Application.CaptureMouse", "Keyboard.RightMouseButton"],
           "to": "Actions.DeltaPitch",
           "filters":
             [
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -4699,6 +4699,11 @@ void Application::maybeToggleMenuVisible(QMouseEvent* event) const {
 void Application::mouseMoveEvent(QMouseEvent* event) {
     PROFILE_RANGE(app_input_mouse, __FUNCTION__);
 
+    if (_ignoreMouseMove) {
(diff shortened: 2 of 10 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-b60ae9f_1403f890.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-b60ae9f.tar_1403f890.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-b60ae9f.tar_1403f890.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-b60ae9f_1403f890.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-b60ae9f_1403f890.exe

vegaslon:
Here is a quck script made to test it, uses the "m" key to enable and disable.
https://gist.githubusercontent.com/vegaslon/0bf901ed5cf2fdcdddded61dbcdd5ded/raw/1e2cbbc41fbd3b846ebc699f22e00a2f87a44a61/Mousecapturewithm.js

vegaslon:
Did not see it cause any trouble when enabled and disable in VR mode with a Occulus Rift. The was functional again when disabled.

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +111,20 @@ public slots:
...
+     * Gets the current mouse capture state.
+     * @function Camera.getMouseCapture
+     * @returns {boolean} The current mouse capture state.
+     */
+    bool getCaptureMouse() const { return _captureMouse; }
+
+    /**jsdoc
+     * Sets mouse capture state.
+     * @function Camera.setCaptureMouse
+     * @param {boolean} captureMouse - Whether or not to capture the mouse.
```
ctrlaltdavid, line 124:
```suggestion
     * @param {boolean} captureMouse - <code>true</code> to capture the mouse, <code>false</code> to release the mouse.
```

ctrlaltdavid (changes requested):

ctrlaltdavid:
Also, update copyrights in:
- CompositorHelper.cpp
- KeyboardMouseDevice.cpp
- KeyboardMouseDevice.h
- Camera.h

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-e19a419.tar_932f5dbf.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-e19a419_932f5dbf.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-e19a419.tar_932f5dbf.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-e19a419_932f5dbf.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-e19a419_932f5dbf.exe

ctrlaltdavid (approved):

JulianGro:
Mouse look works on Linux Mint 19.3, but its performance is pretty bad (same as when holding right click). It is very sensitive and kind of makes jumps, as if it took the mouse moving a pixel and multiplied it with 10 to move 10 pixel at once. Even on the lowest sensitivity setting of my mouse, the movement is not smooth at all.
This jumpy hypersensitivity constantly makes me lose orientation.

HifiExperiments:
@JulianGro interesting!  what’s your framerate like?  it sounds like it’s not a new problem so I wouldn’t block this PR, but we can add mouse sensitivity settings or make the camera framerate independent if it’s not already.  it’s weird because all the calculations are just based on pixel differences, so I’m not sure why linux would be different or why framerate would affect it

JulianGro:
After you asking, I tried it in the serverless rather than the hub, and it does feel smooth in there. The sensitivity is still stupidly high though. I measured it and a 360° spin is just 3mm of mouse movement. For comparison, moving the cursor from one side of the screen to the other is 1cm of mouse movement.
Also I just noticed that the losing orientation in the hub isn't actually because of the high sensitivity, but because the speed is not consistent there. It seems like the movement speeds up when the "gamerate" dips down.

interface/src/Application.cpp
```diff
@@ -680,6 +680,7 @@ class ApplicationMeshProvider : public scriptable::ModelProviderFactory  {
  *     <tr><td><code>CameraIndependent</code></td><td>number</td><td>number</td><td>The camera is in independent mode.</td></tr>
  *     <tr><td><code>CameraEntity</code></td><td>number</td><td>number</td><td>The camera is in entity mode.</td></tr>
  *     <tr><td><code>InHMD</code></td><td>number</td><td>number</td><td>The user is in HMD mode.</td></tr>
+ *     <tr><td><code>CaptureMouse</code></td><td>number</td><td>number</td><td>The mouse is captured.</td></tr>
```
ctrlaltdavid, line 683: Expand on the description - what does it mean for the mouse to be "captured"?

libraries/input-plugins/src/input-plugins/KeyboardMouseDevice.cpp
```diff
@@ -113,9 +113,16 @@ void KeyboardMouseDevice::eraseMouseClicked() {
...
 
-void KeyboardMouseDevice::mouseMoveEvent(QMouseEvent* event) {
+void KeyboardMouseDevice::mouseMoveEvent(QMouseEvent* event, bool capture, QPointF captureTarget) {
     QPoint currentPos = event->pos();
 
+    if (!capture) {
+        _accumulatedMove += currentPos - _lastCursor;
+    } else if (!isNaN(captureTarget.x())) {
+        QPointF change = event->globalPos() - captureTarget;
+        _accumulatedMove += QPoint(change.x(), change.y());
```
ctrlaltdavid, line 123: If `mouseMoveEvent()` calls happen more often than calls to `pluginUpdate()` then the amount of movement will be multiplied?

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +112,20 @@ public slots:
      */
     void setOrientation(const glm::quat& orientation);
 
+    /**jsdoc
+     * Gets the current mouse capture state.
```
ctrlaltdavid, line 116: What does it mean for the mouse to be "captured"?

ctrlaltdavid (changes requested):

interface/src/Application.cpp
```diff
@@ -715,6 +716,7 @@ static const QString STATE_PLATFORM_ANDROID = "PlatformAndroid";
 static const QString STATE_LEFT_HAND_DOMINANT = "LeftHandDominant";
 static const QString STATE_RIGHT_HAND_DOMINANT = "RightHandDominant";
 static const QString STATE_STRAFE_ENABLED = "StrafeEnabled";
+static const QString STATE_CAMERA_CAPTURE_MOUSE = "CaptureMouse";
```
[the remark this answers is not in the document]
HifiExperiments: It’s not quite a camera mode (in that you can be in first person or third person or otherwise and have the mouse captured or not) but it does affect the camera...I will change to STATE_CAPTURE_MOUSE

libraries/input-plugins/src/input-plugins/KeyboardMouseDevice.cpp
HifiExperiments: not sure what you mean.  it’s additive, so this _accumulatedMove will get larger if multiple mouse moves happen at once.  perhaps this explains the framerate dependent behavior that Julian is seeing, but it’s equivalent to how it’s done in master now (the difference is just calculated here instead of above)

JulianGro:
If this can apparently enabled by just any script, it should definitely be overwritten by something like the escape key.
Also I don't know if we have much in terms of security for scripts, but even with an override, I don't think there is anything that would stop someone from putting a script on their avatar that constantly captures and frees the cursor to potentially make it impossible to open the menu to block/mute/kick/ban them.

HifiExperiments:
@JulianGro "capturing" the mouse in this way is actually already possible via Reticle.visible and Reticle.position, which similarly hide the mouse and can take control of it.  this new mode has the advantage that it relinquishes control when Interface is not active or when a menu item is selected, which allows you to get free (like the mouse capture in HMD).  it is true that we'll need a key to get in/out of this mode by default (the issue suggested Alt) and escape (probably via away.js) should always exit this mode and then restore it when you come back, but I've left those to be designed for another PR (and maybe someone who scripts more than me).

and then of course there's a larger discussion about script security + protecting against malicious scripts, but that affects pretty much every aspect of the scripting

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-894791e.tar_9546cebe.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-894791e.tar_9546cebe.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-894791e_9546cebe.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-894791e_9546cebe.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-894791e_9546cebe.dmg

libraries/input-plugins/src/input-plugins/KeyboardMouseDevice.cpp
ctrlaltdavid: OK

ctrlaltdavid (approved):

HifiExperiments:
@kasenvr @ctrlaltdavid @JulianGro I've added an additional `Camera.sensitivity` setting so that people have some more control.  This doesn't fix the framerate issue Julian mentioned, it just allows people to dial the speed of the camera rotation (whether the mouse is captured or not).  I added a slider for this value to Settings -> Controls.

I'm noticing that these preferences might not be saved, which is annoying...but instead of handling it for just this value, I think we should fix that for everything (and do a general push to clean up the settings and expose more helpful things)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-895d013.tar_e4f32f1c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-895d013_e4f32f1c.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-895d013.tar_e4f32f1c.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-895d013_e4f32f1c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-895d013_e4f32f1c.exe

kasenvr:
I thought the Settings API was pretty safe in terms of saving and retrieving. Is it no longer working as expected or which part of things is giving trouble, exactly?

HifiExperiments:
the settings API works, but the Preferences dialogs don’t seem to be all backed by settings, which is why they aren’t saved

interface/resources/controllers/keyboardMouse.json
```diff
@@ -72,156 +72,86 @@
...
         },
 
         { "from": { "makeAxis" : ["Keyboard.MouseMoveLeft", "Keyboard.MouseMoveRight"] },
-          "when": "Keyboard.RightMouseButton",
           "to": "Actions.DeltaYaw",
+          "when": "Application.CaptureMouse",
           "filters":
             [
-                { "type": "scale", "scale": 0.6 }
+                { "type": "scale", "scale": 0.2 }
```
ctrlaltdavid, line 133: Is this change of scale value intentional? Does it adversely affect current sensitivity / speed of movement?

interface/resources/controllers/keyboardMouse.json
```diff
@@ -72,156 +72,86 @@
...
                 { "type": "scale", "scale": 0.6 }
             ]
         },
 
+        { "from": { "makeAxis" : ["Keyboard.MouseMoveUp", "Keyboard.MouseMoveDown"] },
+          "to": "Actions.DeltaPitch",
+          "when": "Application.CaptureMouse",
+          "filters":
+            [
+                { "type": "scale", "scale": 0.2 }
```
ctrlaltdavid, line 151: Ditto.

libraries/shared/src/shared/Camera.h
```diff
@@ -43,6 +44,8 @@ class Camera : public QObject {
     Q_PROPERTY(glm::quat orientation READ getOrientation WRITE setOrientation)
     Q_PROPERTY(QString mode READ getModeString WRITE setModeString NOTIFY modeUpdated)
     Q_PROPERTY(QVariantMap frustum READ getViewFrustum CONSTANT)
+    Q_PROPERTY(bool captureMouse READ getCaptureMouse WRITE setCaptureMouse NOTIFY captureMouseUpdated)
+    Q_PROPERTY(float sensitivity READ getSensitivity WRITE setSensitivity)
```
ctrlaltdavid, line 48: JSDoc for these two new properties needs to be added to FancyCamera.h.

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +113,36 @@ public slots:
...
+     * Gets the current mouse capture state.
+     * @function Camera.getMouseCapture
+     * @returns {boolean} The current mouse capture state.
+     */
+    bool getCaptureMouse() const { return _captureMouse; }
+
+    /**jsdoc
+     * Sets the mouse capture state.  When <code>true</code>, the mouse will be invisible and cannot leave the bounds of
+     * Interface, as long as Interface is the active window and no menu item is selected.  When <code>false</code>, the mouse
+     * will behave normally.
```
ctrlaltdavid, line 126:
```suggestion
     * Sets the mouse capture state.  When <code>true</code>, the mouse is invisible and cannot leave the bounds of
     * Interface, as long as Interface is the active window and no menu item is selected.  When <code>false</code>, the mouse
     * behaves normally.
```

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +113,36 @@ public slots:
      */
     void setOrientation(const glm::quat& orientation);
 
+    /**jsdoc
+     * Gets the current mouse capture state.
+     * @function Camera.getMouseCapture
+     * @returns {boolean} The current mouse capture state.
```
ctrlaltdavid, line 119:
```suggestion
     * @function Camera.getCaptureMouse
     * @returns {boolean} <code>true</code> if the mouse is captured (is invisible and cannot leave the bounds of Interface, if Interface is the active window and no menu item is selected), <code>false</code> if the mouse is behaving normally.
```

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +113,36 @@ public slots:
...
+     * will behave normally.
+     * @function Camera.setCaptureMouse
+     * @param {boolean} captureMouse - <code>true</code> to capture the mouse, <code>false</code> to release the mouse.
+     */
+    void setCaptureMouse(bool captureMouse) { _captureMouse = captureMouse; emit captureMouseUpdated(captureMouse); }
+
+    /**jsdoc
+     * Gets the current camera sensitivity.
+     * @function Camera.getSensitivity
+     * @returns {boolean} The current camera sensitivity.
```
ctrlaltdavid, line 135:
```suggestion
     * @returns {number} The current camera sensitivity. TODO: Describe the permissible range of values.
```

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +113,36 @@ public slots:
...
+     */
+    float getSensitivity() const { return _sensitivity; }
+
+    /**jsdoc
+     * Sets the camera sensitivity.  Higher values mean that the camera will be more sensitive to mouse movements.
+     * @function Camera.setSensitivity
+     * @param {boolean} sensitivity - The desired camera sensitivity.
```
ctrlaltdavid, line 142:
```suggestion
     * @param {number} sensitivity - The desired camera sensitivity. TODO: Describe the permissible range of values.
```

libraries/shared/src/shared/Camera.h
```diff
@@ -110,6 +113,36 @@ public slots:
...
+     */
+    void setSensitivity(float sensitivity) { _sensitivity = glm::max(0.0f, sensitivity); }
```
ctrlaltdavid, line 144: Should there be an upper bound to the sensitivity value?

libraries/shared/src/shared/Camera.h
```diff
@@ -181,6 +214,20 @@ public slots:
...
+     * @param {boolean} newCaptureMouse - The new mouse capture state.
+     * @returns {Signal}
+     * @example <caption>Report mouse capture state changes.</caption>
+     * function onCaptureMouseUpdated(newCaptureMouse) {
+     *     print("The mouse capture has changed to " + newCaptureMouse);
+     * }
+     *
+     * Camera.captureMouseUpdated.connect(onCaptureMouseUpdated);
+     */
+    void captureMouseUpdated(bool newCaptureMouse);
```
ctrlaltdavid, line 229:
```suggestion
    void captureMouseChanged(bool newCaptureMouse);
```
`captureMouseChanged()` would be a better name for this signal. (Most such signals use "Changed" rather than "Updated".)

ctrlaltdavid (changes requested):
Other items in the PreferencesDialog.cpp have their settings saved - in a round about manner. For example, "analog plus walk speed" setting is backed by `Setting::Handle<float> _analogPlusWalkSpeedSetting;` and the setting is saved in `MyAvatar::saveData()`.

kasenvr:
Will give this a QA pass after the CR concludes, then I'll have Nova also test it (Windows) to confirm that all's well for the first-pass implementation then we'll be done. Quick and easy. :)

interface/resources/controllers/keyboardMouse.json
HifiExperiments: This is a matter of the diff being confusing.  The block above this handles the old cases when !Application.CaptureMouse, and still has scale 0.6.  because of how the mouse capture mode works, we want this to be less sensitive.  then on top of that, you can now control the sensitivity further with Camera.sensitivity (with the caveat of framerate-dependent behavior as discussed previously)

interface/resources/controllers/keyboardMouse.json
HifiExperiments: same here, this is the new capture case where we want lower sensitivity

libraries/shared/src/shared/Camera.h
HifiExperiments: I think not needed for now.  In the Preferences dialog it maxes out at 5, but no need to restrict people

ctrlaltdavid (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-1a11d6c.tar_e0289436.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-1a11d6c.tar_e0289436.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-1a11d6c_e0289436.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-1a11d6c_e0289436.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR677/Vircadia-Alpha-PR677-1a11d6c_e0289436.exe

kasenvr (approved):
Appears to work as advertised. One note is that the sensitivity setting seems to be in an odd place, it should get its own area instead of being under "VR Movement", perhaps a "Desktop" section at the top?

Also a very lovely surprise is that this works great for TPS! Maybe we'll want horizontal and vertical sensitivity later + acceleration options, but for now this is a perfect start. Thanks so much for your help on this! I know it was tough but I think it's well worth it. :)

vegaslon:
Tested and seems to work as planned, although at really slow sensitivity the mouse cursor is seen flickering on the screen. That is no reason to not merge this PR just an observation.

PR #668, opened 2020-08-31
New Metaverse PR

kasenvr:
OK, so this is semi-tested...

Based on PR #691 .

PR #691, opened 2020-09-06
Create a system to manage usage of external resources

daleglass:
This is very much a prototype. The idea of this is rather than hardcoding full URLs everywhere, or even having a constant somewhere, it'd be better to actually go through a function. This would have a number of advantages:

* The paths can be set in a config file rather than the source
* Check URLs for sanity
* Work around S3 idiosyncrasies (eg, '/foo/bar' and '/foo//bar' are not the same to S3)
* Useful for debugging
* Spreading the load among multiple mirrors

Possibly it would make sense to use this for other URLs, such as the metaverse API?

It seems to work OK, but the main issue I ran into was making the enum work in QML. The changes in Application.cpp are obviously crap and there has to be some better way of doing it.
 
This was inspired by and will conflict with #668


interface/src/Application.cpp
```diff
@@ -7530,6 +7532,15 @@ void Application::registerScriptEngineWithApplicationServices(const ScriptEngine
     scriptEngine->registerGlobalObject("HifiAbout", AboutUtil::getInstance());  // Deprecated.
     scriptEngine->registerGlobalObject("ResourceRequestObserver", DependencyManager::get<ResourceRequestObserver>().data());
 
+    // This is obviously wrong -- FIXME!
+    scriptEngine->registerValue("Bucket.Public", static_cast<int>(ExternalResource::Bucket::Public));
+    scriptEngine->registerValue("Bucket.Content", static_cast<int>(ExternalResource::Bucket::Content));
+    scriptEngine->registerValue("Bucket.MPAssets", static_cast<int>(ExternalResource::Bucket::MPAssets));
+    scriptEngine->registerValue("Bucket.Assets", static_cast<int>(ExternalResource::Bucket::Assets));
```
daleglass, line 7539: This is bad. How do I do it better?

kasenvr:
I'm thinking the Assets API might be a good place to put this kind of functionality. Currently it deals mainly with ATP stuff, but it should include a great many things that are specifically... well, assets. So when we for example do something like IPFS, commands for that would also go in the Assets API.

daleglass:
That could make sense to move it somewhere else, yeah. I think a shorter name would be a good thing, too. But I'm not sure if it fits in with the ATP stuff at all

kasenvr:
I agree it should not be referenced with anything ATP related, so we'd have to update the API docs to be clear that the Assets API has to do with all Assets perhaps?

Alternatively, then I'd wonder if there's anything that matches the category of... this. Then we could make a namespace for that. But it seems like Asset retrieval all the same.

vegaslon:
Resources might be a good namespace to put this under

kasenvr:
That sounds reasonable. Though, I think we should define what a "Resource" vs an "Asset" is. What makes one or the other different? Maybe assets are to be used in world...? Resources meanwhile are passed internally for the system and for scripts to use? Hm.

daleglass:
Kinda the reason why I went with "ExternalResource" -- clearest thing that came to mind.

I would say that a clear difference would be that an Asset is necessarily something used in-world (and therefore something like a 3D object, sound, texture, etc), while this is concerned with just URLs and could be used for any other place where an external URL is referenced, such as the crash logger. The Bucket enum would need to be renamed for clarity though.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-ed53883_f2ef3cfb.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-ed53883.tar_f2ef3cfb.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-ed53883.tar_f2ef3cfb.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-ed53883_f2ef3cfb.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-ed53883_f2ef3cfb.exe

interface/src/Application.cpp
HifiExperiments: what if you exposed these as QPROPERTIES on the ExternalResource class so that you can do things like ExternalResource.Public?  other scripting interfaces have examples, like this: https://github.com/kasenvr/project-athena/blob/master/interface/src/raypick/PickScriptingInterface.h

daleglass: I poked at the problem a bit more and came up with what's in the commit I just added -- how does that look? It seems fairly similar to what you linked.

HifiExperiments: that’s definitely cleaner!  I still think that since the Bucket enum is only used by the ExternalResource interface, the enum should be a property of ExternalResource.  that way we don’t clutter up the scripting API with two different namespaces

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-9a452eb.tar_54b8a86c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-9a452eb_54b8a86c.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-9a452eb.tar_54b8a86c.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-9a452eb_54b8a86c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-9a452eb_54b8a86c.exe

daleglass:
Moved enum into ExternalResources namespace as per @HifiExperiments's suggestion.

An interesting thing I noticed is that this requires a particular order to work right. The class needs to be registered first, and the enum second. Otherwise the functions inside ExternalResources won't be available to JS.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-643b652_f9aeae3c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-643b652_f9aeae3c.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-643b652.tar_f9aeae3c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-643b652.tar_f9aeae3c.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-643b652_f9aeae3c.exe

interface/src/Application.cpp
```diff
@@ -7530,6 +7532,10 @@ void Application::registerScriptEngineWithApplicationServices(const ScriptEngine
     scriptEngine->registerGlobalObject("HifiAbout", AboutUtil::getInstance());  // Deprecated.
     scriptEngine->registerGlobalObject("ResourceRequestObserver", DependencyManager::get<ResourceRequestObserver>().data());
 
+    
+    scriptEngine->registerGlobalObject("ExternalResource", ExternalResource::getInstance());
```
HifiExperiments, line 7536: if we want QML to have access to this, I think it also needs to be added in `onDesktopRootContextCreated` and `setupQmlSurface`

interface/src/Application.cpp
```diff
@@ -7530,6 +7532,10 @@ void Application::registerScriptEngineWithApplicationServices(const ScriptEngine
...
+    scriptEngine->registerEnum("ExternalResource.Bucket", QMetaEnum::fromType<ExternalResource::Bucket>());
```
HifiExperiments, line 7537: if we do it this way, you need to type "ExternalResource.Bucket.Blah" every time, vs. just "ExternalResource.Blah".  I don't have an opinion either way, but some people might vote for the shorter one, while some people might prefer the explicitness of the former.

libraries/networking/src/ExternalResource.h
```diff
@@ -0,0 +1,139 @@
...
+     * Returns the location of the resource \p relative_path in bucket \p bucket
+     *
+     * @note The resulting path will be sanitized by condensing multiple instances of '/' to one.
+     * This is done for easier usage with Amazon S3 and compatible systems.
+     *
+     * @param bucket The bucket in which the resource is found
+     * @param relative_path The path of the resource within the bucket
+     * @returns The resulting URL as a QUrl
+     */
+    Q_INVOKABLE QUrl getQUrl(Bucket bucket, QString path) {
```
HifiExperiments, line 92: do these all need to be Q_INVOKABLE? I think the public interface should just be getUrl, and we shouldn't expose methods that reference Qt types like QUrl

libraries/networking/src/ExternalResource.h
```diff
@@ -0,0 +1,139 @@
...
+     * @returns The resulting URL as a QString
+     */
+    Q_INVOKABLE QString getUrl(Bucket bucket, const QString &relative_path) {
+        return ExternalResource::getQUrl(bucket, QUrl(relative_path)).toString();
+    };
+
+    private:
+    ExternalResource(QObject* parent = nullptr);
+
+    QMap<Bucket, QUrl> _bucketBases {
```
HifiExperiments, line 131: nitpicking (it really doesn't matter for this small map), but what you want here is really more like just a list, especially since the values are hardcoded.  a QMap is a red-black tree with non-constant lookup, and a QHash is a proper hashmap with constant lookup but extra memory footprint.  also this could be const and static

libraries/networking/src/ExternalResource.h
```diff
@@ -0,0 +1,139 @@
...
+        { Bucket::Public, QUrl("https://public.vircadia.com")},
+        { Bucket::Content, QUrl("https://content.vircadia.com")},
+        { Bucket::Assets, QUrl("https://assets.vircadia.com")},
+        { Bucket::MPAssets, QUrl("https://mpassets.vircadia.com")},
+    };
+};
+
+#endif
```
HifiExperiments, line 139: some general code style things (beating ctrlaltdavid to the punch):

- extra blank lines in the file above
- remove space before/after parens in if statements
- unindent public/private
- newline at the end of this file
- spaces after `;` below

HifiExperiments (changes requested):

daleglass:
Okay, improvements. A bunch of review fixes, now the base paths are modifiable from the settings.

The load/save settings is a bit dubious perhaps. I'm not sure if it's appropriate to handle settings in the networking library? If it is, I can just move it into the class.

interface/src/Application.cpp
daleglass: Ok, tried that. How should it be tested to make sure it went right?

interface/src/Application.cpp
daleglass: Ok, changed

libraries/networking/src/ExternalResource.h
daleglass: I wasn't entirely sure there. Removed. It looks like QUrl just behaves like a string in the console anyway.

libraries/networking/src/ExternalResource.h
daleglass: For now I've left this as-is, as I found a benchmark suggesting QMap is the faster choice for small numbers of items. Also hopefully the latest updates show better what I had in mind. Now the map can actually be modified.

libraries/networking/src/ExternalResource.h
daleglass: Ran the class through the formatter

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-b75ad29.tar_4096c744.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-b75ad29.tar_4096c744.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-b75ad29_4096c744.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-b75ad29_4096c744.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-b75ad29_4096c744.dmg

ctrlaltdavid:
The `Paths` API could perhaps be an alternative place to put this stuff : un-deprecate it and add in this new stuff.

daleglass:
@ctrlaltdavid Where's that in the code? That's quite the generic word

ctrlaltdavid:
\libraries\shared\src\PathUtils.h, .cpp
https://apidocs.vircadia.dev/Paths.html
Though it's only available to Interface, client entity, and avatar scripts.

kasenvr:
> 
> 
> \libraries\shared\src\PathUtils.h, .cpp
> https://apidocs.vircadia.dev/Paths.html
> Though it's only available to Interface, client entity, and avatar scripts.

How hard would it be to make that available to the server? That seems like a sensible namespace.

ctrlaltdavid:
JavaScript API currently implemented by the PR...
```
ExternalResource.Assets number
ExternalResource.Content number
ExternalResource.MPAssets number
ExternalResource.Public number
ExternalResource.getBase(Bucket) function
ExternalResource.getUrl(Bucket,QString) function
ExternalResource.setBase(Bucket,QString) function
```

Possibly better JavaScript API...
```
Script.getExternalPath(number, QString) function
Script.ExternalPaths.Assets
Script.ExternalPaths.HF_Content
Script.ExternalPaths.HF_Marketplace
Script.ExternalPaths.HF_Public
```

Notes:
- `getExternalPath()` is possible new name for `getUrl()`.
- `getBase()` and `setBase()` presumably shouldn't be in the JavaScript API. (Don't annotate them as `Q_INVOKABLE`.)

Similar properties that might be useful as source of code inspiration:
- Desktop.DockArea.BOTTOM number
- Resource.State.FAILED number

interface/src/Application.cpp
```diff
@@ -5534,6 +5539,21 @@ void Application::loadSettings() {
     }
 
     getMyAvatar()->loadData();
+
+
+    auto bucketEnum = QMetaEnum::fromType<ExternalResource::Bucket>();
+    auto er = ExternalResource::getInstance();
```
ctrlaltdavid, line 5545:
```suggestion
    auto externalResource = ExternalResource::getInstance();
```
Coding Standard 1.2.12

interface/src/Application.cpp
```diff
@@ -5534,6 +5539,21 @@ void Application::loadSettings() {
...
+
+    for(int i=0;i<bucketEnum.keyCount();i++) {
+        const char *keyName = bucketEnum.key(i);
+        QString setting("ExternalResource/");
+        setting += keyName;
+        auto bucket = static_cast<ExternalResource::Bucket>(bucketEnum.keyToValue(keyName));
+        Setting::Handle<QString> url(setting, er->getBase(bucket));
+        er->setBase( bucket, url.get() );
+    }
```
ctrlaltdavid, line 5554:
```suggestion
    for (int i = 0; i < bucketEnum.keyCount(); i++) {
        const char* keyName = bucketEnum.key(i);
        QString setting("ExternalResource/");
        setting += keyName;
        auto bucket = static_cast<ExternalResource::Bucket>(bucketEnum.keyToValue(keyName));
        Setting::Handle<QString> url(setting, er->getBase(bucket));
        er->setBase(bucket, url.get());
    }
```
And similar elsewhere.

interface/src/Application.cpp
```diff
@@ -5534,6 +5539,21 @@ void Application::loadSettings() {
...
+
+
```
ctrlaltdavid, line 5556:
```suggestion

```

libraries/networking/src/ExternalResource.h
```diff
@@ -0,0 +1,144 @@
...
+
+    QMap<Bucket, QUrl> _bucketBases{
+        { Bucket::Public, QUrl("https://public.vircadia.com") },
+        { Bucket::Content, QUrl("https://content.vircadia.com") },
+        { Bucket::Assets, QUrl("https://assets.vircadia.com") },
+        { Bucket::MPAssets, QUrl("https://mpassets.vircadia.com") },
+    };
+};
+
+#endif
```
ctrlaltdavid, line 144: Add new line at end of file.

libraries/script-engine/src/ScriptEngine.cpp
```diff
@@ -827,6 +827,19 @@ void ScriptEngine::init() {
...
+    if (!newEnum.isValid()) {
+        qCCritical(scriptengine) << "registerEnum called on invalid enum with name " << enumName;
+        return;
+    }
+
+    for(int i=0;i<newEnum.keyCount();i++) {
+        const char *keyName = newEnum.key(i);
+        QString fullName = enumName + "." + keyName;
+        registerValue(fullName, newEnum.keyToValue(keyName));
+    }
```
ctrlaltdavid, line 840:
```suggestion
    for (int i = 0; i < newEnum.keyCount(); i++) {
        const char* keyName = newEnum.key(i);
        QString fullName = enumName + "." + keyName;
        registerValue(fullName, newEnum.keyToValue(keyName));
    }
```

ctrlaltdavid (changes requested):

daleglass:
@ctrlaltdavid Hmm, why move it over to Script? Looking at it, it doesn't seem entirely fitting.

ctrlaltdavid:
The `Script` API already has two functions related to paths:
- `resolvePath(string)` : Resolves a relative path to an absolute path. The relative path is relative to the script's location.
- `resourcesPath()` : Gets the path to the resources directory for QML files.
Adding a new `getExternalPath(string)` function seems to fit in?
Though, upon reflection, perhaps a more consistent name would be `externalPath(string)`.

An alternative would be to un-deprecate the `Paths` API and add the `externalPath(string)` function there. (Possibly along with new `resolvePath(string)` and `resourcesPath()` functions to make things more complete.) Though then the two existing functions would be in two places and the new function in only one place, which isn't so nice.)

daleglass:
Pushed a non-working implementation of @ctrlaltdavid 's requests. It seems to be mostly there, but I ran into this:

    Script.getExternalPath(Script.ExternalPaths.Assets, "hi")
    [UncaughtException evaluate] Error: cannot call getExternalPath(): argument 1 has unknown type `ExternalResource::Bucket' (register the type with qScriptRegisterMetaType()) in about:console:1 [Backtrace] <global>() at about:console:1
    TypeError: cannot call getExternalPath(): argument 1 has unknown type `ExternalResource::Bucket' (register the type with qScriptRegisterMetaType())


Apparently the trouble is that I register the enum with `ENUM(Bucket)` but `getExternalPath` references it as `ExternalResource::Bucket`, and it doesn't like the difference, though they're the exact same enum.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-12b733b.tar_9da39c79.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-12b733b.tar_9da39c79.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-12b733b_9da39c79.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-12b733b_9da39c79.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-12b733b_9da39c79.exe

ctrlaltdavid:
I've provided a proposed fix to Dale - registers the type with qScriptRegisterMetaType().

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-0a8e4bc_77afe1f4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-0a8e4bc.tar_77afe1f4.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-0a8e4bc.tar_77afe1f4.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-0a8e4bc_77afe1f4.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-0a8e4bc_77afe1f4.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-1d394da_929f3955.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-1d394da.tar_929f3955.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-1d394da_929f3955.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-1d394da.tar_929f3955.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR691/Vircadia-Alpha-PR691-1d394da_929f3955.exe

daleglass:
Since we've got a chain of PRs going, I'm going to close this one, and we'll use Kalila's #668 as the base PR for further development.

PR #713, opened 2020-09-14
Fix audio output switching crash

kasenvr:
Fix double locking attempt if waitForFinished causes prepareLocalAudioInjectors to be called on this thread.

Co-Authored-By: Heather Anderson <1115056+odysseus654@users.noreply.github.com>

### DIFF
--- a/libraries/audio-client/src/AudioClient.cpp
+++ b/libraries/audio-client/src/AudioClient.cpp
@@ -2072,13 +2072,14 @@ bool AudioClient::switchOutputToAudioDevice(const HifiAudioDeviceInfo outputDevi
     // NOTE: device start() uses the Qt internal device list
     Lock lock(_deviceMutex);
 
-    Lock localAudioLock(_localAudioMutex);
     _localSamplesAvailable.exchange(0, std::memory_order_release);
 
     //wait on local injectors prep to finish running
     if ( !_localPrepInjectorFuture.isFinished()) {
         _localPrepInjectorFuture.waitForFinished();
     }
+
+    Lock localAudioLock(_localAudioMutex);
 
     // cleanup any previously initialized device
     if (_audioOutput) {


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR713/Vircadia-Alpha-PR713-aef3968_932a3b9e.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR713/Vircadia-Alpha-PR713-aef3968.tar_932a3b9e.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR713/Vircadia-Alpha-PR713-aef3968_932a3b9e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR713/Vircadia-Alpha-PR713-aef3968.tar_932a3b9e.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR713/Vircadia-Alpha-PR713-aef3968_932a3b9e.exe

HifiExperiments:
I'm not sure I understand this fix.  what is this mutex supposed to be guarding?  _localSamplesAvailable is atomic, so it doesn't really seem to need a lock, but don't we still have the problem that _localPrepInjectorFuture could be modified by multiple threads?  or is the problem the weird way that this lock is used?

kasenvr:
I'm not 100% sure myself, but maybe the lock is not necessary? The lock being before it however was a troublesome case because the system was trying to lock again on the same thread. 

https://github.com/kasenvr/project-athena/blob/f6ee158731a0ba819a3488892d0cc7b0234ce05f/libraries/audio-client/src/AudioClient.cpp#L1494

The alternative might have been to correctly check to see if it's already locked here. But if waitForFinished() spins up another of these threads and it hits that code while this is locked, then it won't be happy.

So that seems to be the problem, but as to why... we're locking like that? https://github.com/kasenvr/project-athena/blame/e6f3ba4458d8f9b5e4dfd13eb30d456f12a4c927/libraries/audio-client/src/AudioClient.cpp#L2075 This might offer insight into the thought process behind all of it.

All I know is that before this change I was crashing quite a bit, and after I couldn't exactly get it to crash.

odysseus654:
I'm not sure what _localAudioMutex is guarding but i'm really hoping it's not guarding the waitForFinished, as _localSamplesAvailable is an atomic I considered it relatively safe to take out from under the umbrella of the mutex.  The fix here is that _localAudioMutex *cannot* be held while calling waitForFinished (due to the calling thread being part of the same threadpool)

The crash was explicitly because lock() was throwing std::system_error (due to attempted lock reentry), which was being translated by Qt into QUnhandledException.

HifiExperiments:
what about switching the mutex to http://www.cplusplus.com/reference/mutex/recursive_mutex/ just to be safe since we don't really understand this code?

odysseus654:
I'm hoping this fix is more appropriate after taking the time to understand the code.  Recursive mutexes would resolve the issue but this feels more like a coding gap than something to work around.

odysseus654:
Rephrase: I generally avoid recursive mutexes unless there is a specific design decision to do so, otherwise they just feel like lazy coding and prone to difficult-to-debug problems that they're trying to cover up.

HifiExperiments (approved):
ok!  I'm still skeptical because I don't understand all the audio threading behavior but if this fixes the issue it seems safe (or at least as safe as the current code)

kasenvr:
Preliminary testing shows it as less crashy than before, I view that as a win! However, in QA we'll put that to the test. :)

SilverfishVR:
Generally, I't seems to solve the crashing when changing audio output 👍 

there are still issues with the logic where the actual output does not correspond to what is selected on the settings page and other weirdness, a few examples:
* switching from VR mod to desktop (Valve index to wireless headset) the output is sometimes in a wierd state where it is on for 2 seconds, and off for 2 seconds, this is in the headset, Idex audio is silent.
* setting it to "computer's default" I would expect to hear the audio on whatever device that currently is, but is seems to be whatever device it was at some arbitrary point, so changing default while Vircadia is running seems to nothing.
* For me generally, switching from one output device to another (in desktop mode) works, switching back again leaves both devices silent, sometimes I can convince one to work again by changing a bunch of times, sometimes not.
* My "best" method now is to quickly switch between a random device, and the one I want to use 👾 it's like a minigame and the prize is getting sound 😄 

note that the logs shows that the correct device is being selected, just no sound.

kasenvr:
All good points and information! Those problems should be resolved, but at least we got the crashing out of the way... The audio code logic is quite complex so to fix those problems might take a little thought.

Issue #508, opened 2020-07-05
Vircadia is still being build with hifiAC


JulianGro:
Since we already broke compatibility with High Fidelity, there is little reason to keep hifiAC if we are worried about permission or even just including proprietary software in our open source project.

daleglass:
Agreed with this during the dev meeting.

The codec is proprietary, and while Kalila has an agreement with HiFi the company, that probably doesn't apply to any forks.

We'll look into adding a parameter to cmake that's disabled by default, so that it can still be built if somebody specifically wants it.

JulianGro:
Yes :angry:

PR #762, opened 2020-09-27
Remove the HiFiAC codec

daleglass:
This is a proprietary codec and it's uncertain who can use it, and
under what conditions. At this point, Opus is stable and a suitable
replacement.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR762/Vircadia-Alpha-PR762-2f2019d_125841af.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR762/Vircadia-Alpha-PR762-2f2019d.tar_125841af.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR762/Vircadia-Alpha-PR762-2f2019d.tar_125841af.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR762/Vircadia-Alpha-PR762-2f2019d_125841af.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR762/Vircadia-Alpha-PR762-2f2019d_125841af.exe

kasenvr:
I agree, though generally I sense we should probably improve the Opus controls a bit before doing this so that people at least have some tuning in how they want to handle things?

HifiExperiments:
if we have permissions to use it, it seems like we could keep this?  if opus becomes the default and is configurable, that’s great, but we might have situations where the hifi codec is necessary?

JulianGro:
#508

JulianGro:
To me this is a no-brainer. We are an open-source project and if someone builds the interface, they essentially infringe the copyright of that codec.
There is no situation where HifiAC *really* is necessary . It is supposed to use a lot less CPU, but you can pretty much always get a bigger or a second server for that. There is also the consideration that HifiAC uses a lot higher bitrate than Opus, which might break Vircadia for people with low bandwidth internet connections.

domain-server/resources/describe-settings.json
```diff
@@ -1343,8 +1343,8 @@
           "name": "codec_preference_order",
           "label": "Audio Codec Preference Order",
           "help": "List of codec names in order of preferred usage",
-          "placeholder": "opus, hifiAC, zlib, pcm",
-          "default": "opus,hifiAC,zlib,pcm",
+          "placeholder": "opus,  zlib, pcm",
```
JulianGro, line 1346:
```suggestion
          "placeholder": "opus, zlib, pcm",
```

JulianGro (changes requested):

daleglass:
Dev meeting: this is waiting on opus performance improvements

PR #804, opened 2020-10-12
Improve logging: color, ms timestamps, PID, thread id via environment variable

daleglass:
This patch exposes some features of the logging system that were only available by tweaking the code, and adds the ability to use ANSI color, which seems to be available on all systems (Windows appears to have also added it).

This should make interesting messages much more distinctive. Example with color and milliseconds enabled:

Screenshot_20201012_174601

### Usage

Set `VIRCADIA_LOG_OPTIONS` to a string containing these keywords:

* color
* milliseconds
* process_id
* thread_id

Multiple options are comma separated. Example:

    $ export VIRCADIA_LOG_OPTIONS="color,milliseconds"

### DIFF
--- a/libraries/shared/src/LogHandler.cpp
+++ b/libraries/shared/src/LogHandler.cpp
@@ -17,6 +17,10 @@
 #ifdef Q_OS_WIN
 #include <windows.h>
 #endif
+#ifdef Q_OS_UNIX
+#include <stdio.h>
+#include <unistd.h>
+#endif
 
 #include <QtCore/QCoreApplication>
 #include <QtCore/QDateTime>
@@ -33,22 +37,36 @@ LogHandler& LogHandler::getInstance() {
 }
 
 LogHandler::LogHandler() {
-    QString log_options = qgetenv("VIRCADIA_LOG_OPTIONS").toLower();
+    QString logOptions = qgetenv("VIRCADIA_LOG_OPTIONS").toLower();
 
-    if (log_options.contains("color")) {
-        _useColor = 1;
+#ifdef Q_OS_UNIX
+    // Enable color by default if we're on Unix, and output is a tty (so we're not being piped into something)
+    //
+    // On Windows the situation is more complex, and color is supported or not depending on version and
+    // registry settings, so for now it's off by default and it's up to the user to do what's required.
+    if (isatty(fileno(stdout))) {
+        _useColor = true;
     }
+#endif
 
-    if (log_options.contains("process_id")) {
-        _shouldOutputProcessID = true;
-    }
+    auto optionList = logOptions.split(",");
 
-    if (log_options.contains("thread_id")) {
-        _shouldOutputThreadID = true;
-    }
+    for(auto option : optionList) {
+        option = option.trimmed();
 
-    if (log_options.contains("milliseconds")) {
-        _shouldDisplayMilliseconds = true;
+        if (option == "color") {
+            _useColor = true;
+        } else if (option == "nocolor") {
+            _useColor = false;
+        } else if (option == "process_id") {
+            _shouldOutputProcessID = true;
+        } else if (option == "thread_id") {
+            _shouldOutputThreadID = true;
+        } else if (option == "milliseconds") {
+            _shouldDisplayMilliseconds = true;
+        } else if (option != "") {
+            fprintf(stdout, "Unrecognized option in VIRCADIA_LOG_OPTIONS: '%s'\n", option.toUtf8().constData());
+        }
     }
 }
 
--- a/libraries/shared/src/LogHandler.cpp
+++ b/libraries/shared/src/LogHandler.cpp
@@ -32,6 +32,26 @@ LogHandler& LogHandler::getInstance() {
     return staticInstance;
 }
 
+LogHandler::LogHandler() {
+    QString log_options = qgetenv("VIRCADIA_LOG_OPTIONS").toLower();
+
+    if (log_options.contains("color")) {
+        _useColor = 1;
+    }
+
+    if (log_options.contains("process_id")) {
+        _shouldOutputProcessID = true;
+    }
+
+    if (log_options.contains("thread_id")) {
+        _shouldOutputThreadID = true;
+    }
+
+    if (log_options.contains("milliseconds")) {
+        _shouldDisplayMilliseconds = true;
+    }
+}
+
 const char* stringForLogType(LogMsgType msgType) {
     switch (msgType) {
         case LogInfo:
@@ -49,6 +69,29 @@ const char* stringForLogType(LogMsgType msgType) {
         default:
             return "UNKNOWN";
     }
+}
+
+const char* colorForLogType(LogMsgType msgType) {
+    switch (msgType) {
+        case LogInfo:
+            return "\u001b[37;1m";  // Bold white
+        case LogDebug:
+            return "";
+        case LogWarning:
+            return "\u001b[35;1m";  // Bright magenta
+        case LogCritical:
+            return "\u001b[31;1m";  // Bright red
+        case LogFatal:
+            return "\u001b[31;1m";  // Bright red
+        case LogSuppressed:
+            return "";
+        default:
+            return "";
+    }
+}
+
+const char* colorReset() {
+    return "\u001b[0m";
 }
 
 // the following will produce 11/18 13:55:36
@@ -133,7 +176,15 @@ QString LogHandler::printMessage(LogMsgType type, const QMessageLogContext& cont
 
     QString logMessage = QString("%1 %2\n").arg(prefixString, message.split('\n').join('\n' + prefixString + " "));
 
-    fprintf(stdout, "%s", qPrintable(logMessage));
+    const char* color = "";
+    const char* resetColor = "";
+
+    if (_useColor) {
+        color = colorForLogType(type);
+        resetColor = colorReset();
+    }
+
+    fprintf(stdout, "%s%s%s", color, qPrintable(logMessage), resetColor);
 #ifdef Q_OS_WIN
     // On windows, this will output log lines into the Visual Studio "output" tab
     OutputDebugStringA(qPrintable(logMessage));
--- a/libraries/shared/src/LogHandler.cpp
+++ b/libraries/shared/src/LogHandler.cpp
@@ -51,7 +51,7 @@ LogHandler::LogHandler() {
 
     auto optionList = logOptions.split(",");
 
-    for(auto option : optionList) {
+    for (auto option : optionList) {
         option = option.trimmed();
 
         if (option == "color") {
--- a/libraries/shared/src/LogHandler.h
+++ b/libraries/shared/src/LogHandler.h
@@ -57,7 +57,7 @@ public:
     void setupRepeatedMessageFlusher();
 
 private:
-    LogHandler() = default;
+    LogHandler();
     ~LogHandler() = default;
 
     void flushRepeatedMessages();
@@ -66,6 +66,7 @@ private:
     bool _shouldOutputProcessID { false };
     bool _shouldOutputThreadID { false };
     bool _shouldDisplayMilliseconds { false };
+    bool _useColor { false };
 
     int _currentMessageID { 0 };
     struct RepeatedMessageRecord {


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-387afa0_389cc5c4.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-387afa0.tar_389cc5c4.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-387afa0.tar_389cc5c4.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-387afa0_389cc5c4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-387afa0_389cc5c4.exe

libraries/shared/src/LogHandler.h
```diff
@@ -66,6 +66,7 @@ class LogHandler : public QObject {
     bool _shouldOutputProcessID { false };
     bool _shouldOutputThreadID { false };
     bool _shouldDisplayMilliseconds { false };
+    bool _useColor{ false };
```
ctrlaltdavid, line 69:
```suggestion
    bool _useColor { false };
```

libraries/shared/src/LogHandler.cpp
```diff
@@ -133,7 +176,15 @@ QString LogHandler::printMessage(LogMsgType type, const QMessageLogContext& cont
 
     QString logMessage = QString("%1 %2\n").arg(prefixString, message.split('\n').join('\n' + prefixString + " "));
 
-    fprintf(stdout, "%s", qPrintable(logMessage));
+    const char *color = "";
+    const char *resetColor = "";
+
+    if ( _useColor ) {
```
ctrlaltdavid, line 182:
```suggestion
    if (_useColor) {
```

libraries/shared/src/LogHandler.cpp
```diff
@@ -133,7 +176,15 @@ QString LogHandler::printMessage(LogMsgType type, const QMessageLogContext& cont
 
     QString logMessage = QString("%1 %2\n").arg(prefixString, message.split('\n').join('\n' + prefixString + " "));
 
-    fprintf(stdout, "%s", qPrintable(logMessage));
+    const char *color = "";
+    const char *resetColor = "";
```
ctrlaltdavid, line 180:
```suggestion
    const char* color = "";
    const char* resetColor = "";
```

libraries/shared/src/LogHandler.cpp
```diff
@@ -32,6 +32,26 @@ LogHandler& LogHandler::getInstance() {
...
+        _shouldOutputProcessID = true;
+    }
+
+    if ( log_options.contains("thread_id")) {
+        _shouldOutputThreadID = true;
+    }
+
+    if ( log_options.contains("milliseconds")) {
+        _shouldDisplayMilliseconds = true;
+    }
```
ctrlaltdavid, line 52:
```suggestion
    if (log_options.contains("color")) {
        _useColor = 1;
    }

    if (log_options.contains("process_id")) {
        _shouldOutputProcessID = true;
    }

    if (log_options.contains("thread_id")) {
        _shouldOutputThreadID = true;
    }

    if (log_options.contains("milliseconds")) {
        _shouldDisplayMilliseconds = true;
    }
```

ctrlaltdavid (changes requested):

kasenvr:
I'll add a checkbox for you like I did for Heather's PR now or later mayhaps? :)

libraries/shared/src/LogHandler.cpp
```diff
@@ -51,6 +71,29 @@ const char* stringForLogType(LogMsgType msgType) {
...
+            return "\u001b[31;1m";  // Bright red
+        case LogSuppressed:
+            return "";
+        default:
+            return "";
+    }
+}
+
+const char* colorReset() {
+    return "\u001b[0m";
```
ctrlaltdavid, line 94: Add a comment saying what colour this is?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-96e20d8.tar_895d4d41.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-96e20d8.tar_895d4d41.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-96e20d8_895d4d41.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-96e20d8_895d4d41.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-96e20d8_895d4d41.exe

libraries/shared/src/LogHandler.cpp
daleglass: It's not a color, it resets the color to the default.

daleglass: https://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html#16-colors

ctrlaltdavid (approved):

JulianGro:
Seems to work good.
Any reason colors are not on by default?

daleglass:
> Seems to work good.
> Any reason colors are not on by default?

I looked into it and apparently while Windows does support ANSI color these days, it's a fairly recent addition, and I'm not sure what happens on earlier versions. I'm not sure if there's a way of determining whether it's supported.

On Unixes, it should check if stderr is a tty or not.

So it could be done, it'd just need more work to figure out all the details to make sure it doesn't break anything, and this was a random thing I did because I wanted a more readable output for debugging. But I could look into it if there's interest.

JulianGro:
> > Seems to work good.
> > Any reason colors are not on by default?
> 
> I looked into it and apparently while Windows does support ANSI color these days, it's a fairly recent addition, and I'm not sure what happens on earlier versions. I'm not sure if there's a way of determining whether it's supported.
> 
> On Unixes, it should check if stderr is a tty or not.
> 
> So it could be done, it'd just need more work to figure out all the details to make sure it doesn't break anything, and this was a random thing I did because I wanted a more readable output for debugging. But I could look into it if there's interest.

At least on Unix I would like that.
I feel like Windows users are pretty unlikely to run Vircaida in the command line.

daleglass:
> At least on Unix I would like that.
> I feel like Windows users are pretty unlikely to run Vircaida in the command line.

Okay, done. Color on Unix by default if used on a terminal. Added "nocolor" option, and made the options comma separated.

Testing:

    # color enabled
    $ interface/interface

    # No color, not being sent to a terminal directly
    $ interface/interface | tee test.log

    # Force color on, logs escape sequences to test.log
    $ VIRCADIA_LOG_OPTIONS="color" interface/interface | tee test.log

    # Comma separated options (whitespace is not significant)
    $ VIRCADIA_LOG_OPTIONS="color, milliseconds" interface/interface

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-56536fb_11245904.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-56536fb_11245904.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-56536fb.tar_11245904.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-56536fb.tar_11245904.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-56536fb_11245904.exe

JulianGro:
Just built it and it doesn't seem to work as expected.
The color works if it is manually enabled, but it does not just turn on.
I ran Vircadia using the `run_interface` script created by vircadia-builder inside gnome-terminal 3.28.1

daleglass:
> Just built it and it doesn't seem to work as expected.
> The color works if it is manually enabled, but it does not just turn on.
> I ran Vircadia using the `run_interface` script created by vircadia-builder inside gnome-terminal 3.28.1

That's because this is in the script:

    "$SCRIPT_DIR/interface/interface"  "$@" 2>&1 | tee -a "$SCRIPT_DIR/interface.log"

It sends the output through `tee`, and that disables color, since most people don't want to see ANSI sequences in `less` or `vim`. Really, I should just get rid of the tee bit, since `~/.local/share/Vircadia - dev/Interface/Logs` exists.

Anyway, change it to this and it'll work properly:

    "$SCRIPT_DIR/interface/interface"  "$@"

JulianGro:
That makes a lot of sense.
It does work fine now.
I guess the only thing that is a bit weird to me is that you can set `color` and `nocolor` at the same time (it will use the last one specified). Might make sense to throw a warning for that.
Also I guess that something like `Unrecognized option in VIRCADIA_LOG_OPTIONS: 'process_id thread_id'` should have a warning color as well.

Seems good to merge to me anyways though.

daleglass:
@JulianGro "Use the last option" is a common policy, actually. And sometimes helpful in that it allows overriding options that come from somewhere else. Not that it'd happen here, but it's a pretty normal Linux pattern.

There's no color because it's done during the log system's initialization, when we're deciding whether to have color. It's also just printing it directly to the console, since we can't have the log system using itself to log stuff.

daleglass:
Updated vircadia-builder to remove the tee stuff, now color will work from `run_interface`

kasenvr:
Needs testin' on Windows if anyone wants to give it a shot. :)

kasenvr:
Running this in VS2019 with the following options...
image
Does not give me millisecond precision logging, how come?

daleglass:
@kasenvr Ah, I see, you have to remove the quotes.

On Linux, if you do this, the second word is interpreted as a command name

    $ A=hello world
    zsh: world: command not found...

Eg:

    $ VIRCADIA_LOG_OPTIONS=milliseconds interface/interface

To have spaces in a variable you have to use quotes:

    $ A="hello world"
    $ echo $A
    hello world

But the quotes aren't literally part of the value, they're there for the shell.

kasenvr:
Ahh, I am aware. I use quotes though because you're typically allowed to use them even if you don't need them. So I'm confused on why it wasn't working when I did add them. Are we parsing the env variable differently there?

daleglass:
The quotes just aren't recognized as a valid part of the value. It should actually output an error message

libraries/shared/src/LogHandler.cpp
```diff
@@ -32,6 +36,40 @@ LogHandler& LogHandler::getInstance() {
...
+    // On Windows the situation is more complex, and color is supported or not depending on version and
+    // registry settings, so for now it's off by default and it's up to the user to do what's required.
+    if (isatty(fileno(stdout))) {
+        _useColor = true;
+    }
+#endif
+
+    auto optionList = logOptions.split(",");
+
+    for(auto option : optionList) {
```
kasenvr, line 54:
```suggestion
    for (auto option : optionList) {
```

kasenvr (changes requested):
It appears that when combining milliseconds with another item such as "thread_id" or "process_id" then the milliseconds logging would not trigger...

daleglass:
It works, they just have to be comma separated. In my first commit it didn't matter what separator you used, but with the addition of things like "nocolor" I had to be a bit pickier.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-3ae3bd4.tar_6c25b947.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-3ae3bd4.tar_6c25b947.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-3ae3bd4_6c25b947.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-3ae3bd4_6c25b947.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR804/Vircadia-Alpha-PR804-3ae3bd4_6c25b947.dmg

kasenvr:
I did comma separate them. Though I am inputting these in the environment section of Visual Studio debugger, perhaps I will try adding them to my env directly and restarting Visual Studio?

kasenvr (approved):
Works fine for me on Windows without quotations. :)

PR #847, opened 2020-10-29
Example "modkit" JS Plugin renovation

humbletim:
This PR revamps the modkit example plugin we added last year to hopefully work as an easier starting point for anyone wanting to experiment with adding new native JS APIs.

These changes are also meant to address several questions to do with returning different value types from the C++ side, as well as to document a few best (or simplest) practices when it comes to futureproofing Qt/C++/JS integration code.

Note: To use this plugin (or any scripting plugin), the checkbox in Security->Settings has to be ticked. From the standpoint of a typical user, this PR therefore does not affect any default application functionality.

### Summary of changes:

* Eliminated the need for `ExampleScriptPlugin.h` (in favor of keeping everything as a single, easy-to-copy .cpp file).
* Eliminated unnecessary 'package' fields from `plugin.json` (which were just ideas we had back then and were never actually wired up).
* Replaced previous Blendshape example with simpler examples that cover different return types:
  - key-value maps (JS `Object`)
  - value lists (JS `Array`)
  - raw binary data (JS `ArrayBuffer`);
  - native C++ instances ("managed" JS `Object`)

Most of these JS methods intentionally avoid using `QScriptEngine`/`QScriptValue` so as to remain compatible with future scripting engines -- for example, this same QObject plugin class could be registered as-is into QML `QJSEngine` contexts too.

#### Notes on managed native instances:

The method `JSAPIExample.getScopedSettings(scope)` uses a QtScript-specific feature that allows C++-allocated classes to be lifecycle managed by the scripting engine itself.

In this case, when there are no more references to a "scoped settings" instance, the script engine automatically garbage collects it and invokes the native destructor.

### Test Plan:

1) Enable `Settings -> Security` checkbox to allow script plugins to be used (note: you will need to be logged in to see the checkbox)
2) Restart interface.exe
3) Open the JS Console...
4) Type: `JSAPIExample` [and hit enter]

... if you see `[JSAPIExample v0.0.1]`, then it worked!

#### Misc

While none of the JS methods added by this plugin are really meant for production use, I needed something "instanced" to demonstrate how QtScript managed object support works. The resulting "scoped settings" concept might be useful to consider some day on its own.

Basically, scripts request access to a sanitized, prefixed .json file that lives alongside -- but is otherwise unrelated to -- the main app Interface.json file.

```js
var settings = JSAPIExample.getScopedSettings("example");

print(settings.fileName());
// for me above prints /home/tim/.config/Vircadia - dev/jsapi_example.json

settings.setValue('foo', 'bar');
print(settings.getValue('foo'));
print(settings.allKeys()); // prints ['foo']
```

//


plugins/JSAPIExample/src/JSAPIExample.cpp
```diff
@@ -0,0 +1,250 @@
...
+QVariant JSSettingsHelper::getValue(const QString& key, const QVariant& defaultValue) {
+    return _settings ? _settings->value(key, defaultValue) : defaultValue;
+}
+bool JSSettingsHelper::setValue(const QString& key, const QVariant& value) {
+    if (_settings) {
+        if (value.isValid()) {
+          _settings->setValue(key, value);
+        } else {
+          _settings->remove(key);
+        }
```
ctrlaltdavid, line 233: Indenting of 2 x `_settings->...` lines.

plugins/JSAPIExample/src/JSAPIExample.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include <QtCore/QTimer>
+#include <QtScript/QScriptEngine>
+#include <QtScript/QScriptable>
+
+#include <SettingHelpers.h>  // for ::settingsFilename()
+#include <SharedUtil.h>      // for ::usecTimestampNow()
+#include <shared/ScriptInitializerMixin.h>
+
+// NOTE: replace this with your own namespace when starting a new plugin (to avoid .so/.dll symbol clashes)
+namespace REPLACE_ME_WITH_UNIQUE_NAME {
```
ctrlaltdavid, line 29: Not indenting the namespace content makes it a bit hard to find its extent.

plugins/JSAPIExample/src/JSAPIExample.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+static constexpr auto JSAPI_SEMANTIC_VERSION = "0.0.1";
+static constexpr auto JSAPI_EXPORT_NAME = "JSAPIExample";
+
+QLoggingCategory logger{ "jsapiexample" };
```
ctrlaltdavid, line 34:
```suggestion
QLoggingCategory logger { "jsapiexample" };
```
Existing code / coding standard prefers a space.

plugins/JSAPIExample/src/JSAPIExample.cpp
```diff
@@ -0,0 +1,250 @@
...
+      * print("(before) example::timestamp", settings.getValue("timestamp"));
+      * settings.setValue("timestamp", Date.now());
+      * print("(after) example::timestamp", settings.getValue("timestamp"));
+      * print("all example::* keys", settings.allKeys());
+      * settings = null; // optional best pratice; allows the object to be reclaimed ASAP by the JS garbage collector
+      */
+    QScriptValue getScopedSettings(const QString& scope) {
+        auto engine = QScriptable::engine();
+        if (!engine)
+            return QScriptValue::NullValue;
```
ctrlaltdavid, line 162:
```suggestion
        if (!engine) {
            return QScriptValue::NullValue;
        }
```

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR847/Vircadia-Alpha-PR847-cde86db_5c2a8bd4.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR847/Vircadia-Alpha-PR847-cde86db.tar_5c2a8bd4.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR847/Vircadia-Alpha-PR847-cde86db.tar_5c2a8bd4.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR847/Vircadia-Alpha-PR847-cde86db_5c2a8bd4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR847/Vircadia-Alpha-PR847-cde86db_5c2a8bd4.exe

ctrlaltdavid (approved):

kasenvr:
Could not get the QML checkbox setting to persist, very weird.... I'll have to manually set this in Interface.json if only to test the PR.

kasenvr (approved):
Works as expected.

Issue #845, opened 2020-10-27
enterEntity is broken.


AleziaKurdis:
enterEntity and leaveEntity seems broken. Those event fails to trigger (It was working in the paste)

STEPS TO REPRODUCE:
1- Display the Debug Window (**Developer > Scripting > Script Log (HMD friendly...)**)

2- Create a **Shape (Box**) entity

3- Resize it to be enough tall for your avatar to be able to fit inside of it. 
(but keep the box in front of you, you must not be inside at this moment.)

4- Set it collisionless

5- Add this script (Client Entity Script): 
_http://metaverse.bashora.com/scripts/curling/test_enterEntity.js_
This a very simple script:
################################
(function(){ 
    this.preload = function(entityID) {
        print("ACTIVATED!");
    }
    this.enterEntity = function(entityID) {
        print("ENTERED!");
    }; 
    this.leaveEntity = function(entityID) {
        print("LEFT!");
    };
})
#############################

5- Note that you can saw "**ACTIVATED!**" displayed in the Debug Window as the script loaded.

6- Enter the box entity...
Nothing happens. Normally , **enterEntity** would have printed "**ENTERED!**" in the Debug Window.  
(It works sometimes, if you have already visited the domain before, it might decide to work. 
Sadly, it's too bad for the newcomer. 

7- Exit the box... Nothing happens there too. 

I get this situation in different domains and serverless domain.

ctrlaltdavid:
Just tested 2020.3.1 on localhost and it's working OK.
I entered the URL given in the "Script" entity property.
"LEFT!" and "ENTERED!" are being reported to the program log and script debug window, OK.
Also tested in the Community Village and the Help > Tutorial serverless domain.

It doesn't work if the script is loaded as a server entity script, though.
Should it?

AleziaKurdis:
I reproduce it here:
http://metaverse.bashora.com/explore/serverlessEnterEntityTester.json

The first time I went on this serverless, the script worked.
I could see "ENTERED!" and "LEFT!" logged in the debug window when I crossed the yellow cube.
Then I did Reload Content, and the scripts get loaded (I could see "ACTIVATED!" printed in the debug window)
but nothing happens when I crosse the yellow cube.
So I re teleport to this location (using explore app, by pasting the URL)
the script didn't work more after.
Then I quit Interface, and relaunched it, then went to that location
then the script was working fine.

So to recap, after have done "Reload Content", enterEntity stops to work until we restart interface.

[...]

PR #853, opened 2020-11-03
Fix enterEntity event not firing in entity script after content reload

ctrlaltdavid:
Fixes `enterEntity` and `leaveEntity` not firing in client entity script after doing Developer > Network > Reload Content.

### DIFF
--- a/libraries/entities-renderer/src/EntityTreeRenderer.cpp
+++ b/libraries/entities-renderer/src/EntityTreeRenderer.cpp
@@ -167,6 +167,13 @@ void EntityTreeRenderer::resetEntitiesScriptEngine() {
     auto entityScriptingInterface = DependencyManager::get<EntityScriptingInterface>();
     entityScriptingInterface->setEntitiesScriptEngine(entitiesScriptEngineProvider);
 
+    connect(_entitiesScriptEngine.data(), &ScriptEngine::entityScriptPreloadFinished, [&](const EntityItemID& entityID) {
+        EntityItemPointer entity = getTree()->findEntityByID(entityID);
+        if (entity) {
+            entity->setScriptHasFinishedPreload(true);
+        }
+    });
+
     // Connect mouse events to entity script callbacks
     if (!_mouseAndPreloadSignalHandlersConnected) {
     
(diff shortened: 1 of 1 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR853/Vircadia-Alpha-PR853-b58bd0a.tar_c6728be4.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR853/Vircadia-Alpha-PR853-b58bd0a_c6728be4.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR853/Vircadia-Alpha-PR853-b58bd0a.tar_c6728be4.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR853/Vircadia-Alpha-PR853-b58bd0a_c6728be4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR853/Vircadia-Alpha-PR853-b58bd0a_c6728be4.exe

libraries/entities-renderer/src/EntityTreeRenderer.cpp
```diff
@@ -167,6 +167,13 @@ void EntityTreeRenderer::resetEntitiesScriptEngine() {
...
 
+    connect(_entitiesScriptEngine.data(), &ScriptEngine::entityScriptPreloadFinished, [&](const EntityItemID& entityID) {
+        EntityItemPointer entity = getTree()->findEntityByID(entityID);
+        if (entity) {
+            entity->setScriptHasFinishedPreload(true);
+        }
+    });
+
     // Connect mouse events to entity script callbacks
     if (!_mouseAndPreloadSignalHandlersConnected) {
```
HifiExperiments, line 178: I've been suspicious of this if check for a while.  are we sure all the other mouse events that are connected below work after a reload?  I think if we just removed the third argument, I would expect them to, but with the third argument being `_entitiesScriptEngine.data()`, I would expect the signal to be disconnected in the same way when the script engine is reset

ctrlaltdavid: Yes, the mouse events below work after reload - even though `_mouseAndPreloadSignalHandlersConnected == true`  still after reload. (Which is why the code calling `entity->setSciptHasFinishedPreload(true);` had to be moved - to connect the new `_entitiesScriptEngine`'s signal `entityScriptPreloadFinished`, and so set things up such that `enterEntty` and `leaveEntity` can fire.

The following test script demonstrates this...

[enterEntity.zip](https://github.com/kasenvr/project-athena/files/5503762/enterEntity.zip)

HifiExperiments (approved):

AleziaKurdis:
This passed the test successfully 
for the Issue #845 enterEntity is broken

daleglass (approved):

kasenvr (approved):
Works fine to me. :)

Issue #869, opened 2020-11-17
New Zones appear as parent of a ghost child. Issue with Entities.getChildrenIDs()


AleziaKurdis:
New Zones appear as parent of a ghost child. Issue with Entities.getChildrenIDs()

1- Create a new zone, (note that it is displayed as a Parent, which doesn't make sense.)
2- Select the zone, then do "Selection... > Add Children to Selection" (the selection in the list become display as no more the last entity, the selection in-world is now grey, as for a multiple selection)
3- from the list, keep pressed CTRL, and deselect the zone. Now you see the mysterious ghost child in cyan selected in-world.

![image](https://user-images.githubusercontent.com/60075796/99343508-0a4f3f00-285c-11eb-98de-e6509e956252.png)

This happens only with zones. The ghost disappears if we add another entity that is not a zone.

The thing is that it doesn't seem to be the editor.
As you can see here, this is the piece of script that is responsible to figure what is the parent state of the entity:

function getParentState(id) {
    var state = "NONE";
    var properties = Entities.getEntityProperties(id, ["parentID"]);
    var children = Entities.getChildrenIDs(id);
    if (properties.parentID !== Uuid.NULL) {
        if (children.length > 0) {
            state = "PARENT_CHILDREN";
        } else {
            state = "CHILDREN";
        }
    } else {
        if (children.length > 0) {
            state = "PARENT";
        }
    }
    return state;
}

Entities.getChildrenIDs() seems to return something when it's a new zone entity for any obscure reason.
For sure, that ghost child has an id, since it's able to determine that it has a parent. What's hidden behind this?
Severity: **Medium**
Priority: **Medium**

HifiExperiments:
is it possible that this is something that edit.js is making?  the “zone outline” thing where it shows a grid when you select a zone is accomplished with some local entities + a material entity, maybe they’re not being initialized properly?

AleziaKurdis:
It does make sense. I will check how this has been done.

This leads to some questions...
Is a local entity can be parented to an a domain entity ? (I guess this would be possible)
is a domain entity can be parented to a local entity?  (This might be a bit dirty)

What would be the most performant?
- looping (script side) on the results returned by Entities.getChildrenIDs() to exclude the local entities
- or have this done (code side) with an eventual optional parameter like Entities.getChildrenIDs(onlyDomainEntities)

AleziaKurdis:
I added some code to exclude the non-domain entities
[...]

PR #874, opened 2020-11-19
Fix new Zones appearing as parent of a ghost child

AleziaKurdis:
Fix for Issue #869

This excludes the local and avatar entities from what is returns by Entities.getChildrenIDs
to avoid the selection tools of the Zone entities to be considered as Children of it.
This was necessary for the display of the hierarchy status, but also for the "Add Children to Selection".

**For the QA:**
1- Open the "Create" Application
2- Make sure to have the "Hierarchy" column displayed in the entity list.
3- Create a new Zone entity
**Before this fix:**
The new zone entity was displayed as being a Top Parent. (because there was a local entity parented to it. (to display the grid box)
In In-world view, the bounding box was also indicating that it was a Top Parent (it was displayed in Orange instead of light gray)
If you selected that new zone, and do "**Selection... > Add Children to Selection**", the bounding box turned light gray as a multiple selection. If you were keeping CTRL pressed and deselect the zone only, then you could see the invisible local entity selected with a bounding box in cyan (as a child).

**After this fix:**
No more trace of any children, since local entities are now ignored when we get the list of the children.
So new zones are no more displayed as a Parent of a child that we didn't expect. And the  "**Selection... > Add Children to Selection**" won't consider this false child too.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-43f3ba5_22abf1c2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-43f3ba5_22abf1c2.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-43f3ba5_22abf1c2.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-43f3ba5.tar_22abf1c2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-43f3ba5.tar_22abf1c2.gz

ctrlaltdavid (approved):

kasenvr:
According to QA it worked as expected: I created the zone entity and it doesn't show any children under it on the icon. Though, I didn't test any parenting as I was unsure what causes what in that respect.

AleziaKurdis:
Hmm... Can we use the create application to edit avatar entities?

Because we can't for "Local" entities. They are not visible in the list and can't be selected.
To be able to create a local entity or set it a local entities as child of another entity, the only way is to do it using a script.

kasenvr:
We are able to adjust wearables with the Create App as far as I know.

AleziaKurdis:
We will put this PR on HOLD.
I will change the strategy.
and figure to exclude those specific "tools" entities
instead of excluding local entities (and avatar entities)
Seems that eventually, Local entities could be visible like all the others.

AleziaKurdis:
Finally, I change the approach.
I exclude now only the Entity Shape Visualizer 
instead of systematically all the local entities (and the avatar entities by the same time)

So this will continue to work for all types of entities. (domain, local, avatar)

I still don't think it will be viable to have the local entities visible in the create Application
it might seriously complexify everything to be able to protect the selection tool to self-edit

If we do this, I suggest that we implement a property to have specific local entities to never be returned by the findEntities### methods, getChildrens, and certainly some others.

AleziaKurdis:
You can now remove the "**Do not Merge**" flag and gives this to QA.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-4b9bf94_b633bc4a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-4b9bf94.tar_b633bc4a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-4b9bf94_b633bc4a.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-4b9bf94.tar_b633bc4a.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-4b9bf94_b633bc4a.exe

scripts/system/create/edit.js
```diff
@@ -2923,4 +2925,17 @@ function getParentState(id) {
     return state;
 }
 
+function getDomainOnlyChildrenIDs(id) {
+    var allChildren = Entities.getChildrenIDs(id);
+    var realChildren = [];
+    var properties;
+    for (var i = 0; i < allChildren.length; i++) {
+        properties = Entities.getEntityProperties(allChildren[i], ["name"]);
+        if (properties.name !== entityShapeVisualizerSessionName && properties.name !== undefined) {
```
ctrlaltdavid, line 2934:
```suggestion
        if (properties.name !== undefined && properties.name !== entityShapeVisualizerSessionName) {
```
Thought it doesn't matter in this code, I suggest testing against `undefined` first, just as a matter of good habit.

ctrlaltdavid (changes requested):

ctrlaltdavid (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-6b2ea04_343ded8c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-6b2ea04.tar_343ded8c.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-6b2ea04.tar_343ded8c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-6b2ea04_343ded8c.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/kasenvr_project-athena/PR874/Vircadia-Alpha-PR874-6b2ea04_343ded8c.exe

Aitolda:
Seems to be working as expected now.

kasenvr (approved):
Works fine for me as well.

PR #918, opened 2020-12-15
Add initial data export API.

digisomni:
This is an initial working version of a data export API for Interface scripts to use. It's a bit dangerous since it can of course write (or overwrite) files on the system of a specified name and path... so the way I set it up is that the actual path and filename can only be set by the user using a Qt/QML dialog. Once set, all scripts can write to that path.

I think what's missing still is a reset function to stop scripts from mucking with your export after you've completed it. 

Generation 2 should consist of: a per-app export API that allows you to set paths per-app that persist restarts. (this would of course have to use the private context in order to prevent scripts from altering the paths.)

Generation 3 should consist of: a per-app limited storage based on path+filename/URL+filename. Maybe a 512KB data limit would be good with mayhaps a permission dialog opening to say the app wants to have its own space to write to that sits in the Vircadia appData folder.

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -30,6 +30,7 @@
 #include <QtCore/QMimeData>
 #include <QtCore/QThreadPool>
 #include <QtCore/QFileSelector>
+#include <QtCore/QSaveFile>
 #include <QtConcurrent/QtConcurrentRun>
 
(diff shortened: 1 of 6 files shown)


daleglass:
Dev meeting: it'd be nice to have a test script for this one, or at least scripter-friendly documentation on how to use it

Issue #851, opened 2020-11-02
Friction does not seem to scale with avatar size.


Aitolda:
Shrink down very small and try to move.  It's like wading through molasses.

JulianGro:
Are you sure this isn't just the movement speed scaling with the avatar size?

Issue #825, opened 2020-10-19
VR head position constantly being "corrected"


JulianGro:
When you move you head in VR vertically, Interface will move it back into the position it would have if you were standing.
Not only is this extremely disorienting even for veteran VR users, but you cannot move your head down to pick stuff up, look at a small person, sit down, etc.
Interestingly this also happens when trackers are enabled, so even if you are laying on the floor, Interface will move your laying avatar into the air.

Oculus should have a way to give Interface the height of the user. (At least this was part of the calibration process when I still had an Oculus)
SteamVR does not have data on the user height, so we would need to have people either set their real life height or have them calibrate their height. We could add a wizard that asks this when we are missing the info in VR. Recalibration or changing the height could be done in the controls menu.

ctrlaltdavid:
This behaviour happens with Oculus, too.

HifiExperiments:
I might be wrong but isn’t there a setting for sitting vs room scale VR or something like that?

JulianGro:
> I might be wrong but isn’t there a setting for sitting vs room scale VR or something like that?

There is a setting that makes your head position absolute, but that leaves your body anchored to the middle of the playspace.
So if you use that setting and move in real life, your avatar will just bend unrealistically.

ctrlaltdavid:
Settings ? Controls... has some options that should already be able to control this as required by user for different situations: Auto / Seated / Standing. C.f. behaviour with documentation and what is wanted.

Issue #117, opened 2020-01-23
User-real world height in Settings/Controls... can only be edited with steps of 1 meter


Basinsky:
The field user-real world height in Settings/Controls.. can not be edited to the correct number. 1 meter is minimum and 2.2 is maximum. Pushing on the up and down buttons increases the height with 1 meter instead of 0.001 meter. Multiple input fields dont allow typing of decimal numbers.

daleglass:
This is partially fixed by PR #1161, but there are still localization issues.

PR #928, opened 2020-12-22
VR fixes for: couldn't sit on the floor, wrong walk directions.

Phil-Palmer:
### Fixes

1. Fix being unable to sit in the real world in VR without the avatar and viewpoint being popped up to a standing position (fixes https://github.com/vircadia/vircadia/issues/825).  With the new option "Allow my avatar to stand", we can choose to have the avatar stand only when the user is standing.  The default behaviour is the same as before (always allowed to stand; allowed to lean only when user is standing), except that now, when foot trackers are used, the avatar is never made to stand artificially.
2. Fix very wrong walk/fly directions when using foot tracking or when using the mode that allows the avatar to sit when the user sits (previously "Avatar lean behavior: disabled").
3. Fix VR walk speeds being too slow at small avatar scale and too fast at large avatar scale (fixes https://github.com/vircadia/vircadia/issues/851).  The avatar scale was being factored-in twice in the velocity calculation.
4. The "Avatar lean behavior" setting was unclear and had muddled logic.  It's now split into two settings that work more usefully: "Allow my avatar to stand" and "Allow my avatar to lean".
5. The previous experimental mode that allowed the avatar to sit on the floor ("Avatar leaning behavior: disabled") is improved by now having the avatar follow the user's orientation ("Allow my avatar to lean: Always, no recenter").
6. Fix unreliable hand-relative walk/fly directions, due to avatar-frame vectors being wrongly transformed by the head rotation.  UpdateMotors was transforming avatar-space directions by the head rotation (even if hand-relative walking was chosen).  It could give the appearance of working, but only when the avatar rotation followed the head rotation.
7. Fix for the recentring not behaving as requested by the user after getting up from click-to-sit (always popping up to standing position).
8. Fix for incorrect vertical position of the avatar and viewpoint after changing recentre mode while not standing in the real world.
9. The logic controlling the direction of MyAvatar's action motor was hard to follow and maintain, as it was split between two functions.  Now it's clearer and all done in calculateScaledDirection.
10. Leaning and lean recentring now work as intended at any avatar scale.

### Code changes

- Divided the option "Avatar leaning behavior" into two options that work more usefully: "Allow my avatar to stand" and "Allow my avatar to lean" (PreferencesDialog.cpp).  Made the necessary fixes so that the avatar can be set to stand only when the user is standing (more details below).
- The logic controlling the direction of MyAvatar's action motor is now centralised in calculateScaledDirection (was previously split between there and updateMotors).  calculateScaledDirection now returns a velocity in world space.
- CharacterController::FollowHelper now uses separate follow timers for rotation, horizontal and vertical (previously followed all three based on the longest of their follow times).  Where appropriate, FollowHelper can now snap immediately to the desired rotation/horizontal/vertical independently (see FOLLOW_TIME_IMMEDIATE_SNAP).
- FollowHelper::FollowType has therefore moved to CharacterController::FollowType.
- (point 1 above) MyAvatar::FollowHelper::postPhysicsUpdate: If MyAvatar is not allowed to stand when the user is sitting, this now avoids recentring the body based on the head height.
- Deprecated Q_PROPERTY(MyAvatar::SitStandModelType, as the sitting/standing/leaning model uses different enums now (see setAllowAvatarStandingPreference, setAllowAvatarLeaningPreference).
- Deprecated Q_PROPERTY(bool isSitStandStateLocked which is no longer used, because we now always track the user's real-world sit/stand state, regardless of what we're doing with it.
- MyAvatar::FollowHelper::shouldActivateHorizontal: If MyAvatar is not allowed to lean, this now returns true to recentre the footing if the head is outside the base of support.
- MyAvatar::FollowHelper::shouldActivateHorizontalCG: If MyAvatar is not allowed to lean, this now always returns true to recentre the footing.  Rearranged to avoid computing values that weren't used depending on the conditions.  Resolved some duplicated code.
- MyAvatar::setUserRecenterModel previously set HMDLeanRecenterEnabled based on the chosen mode, but it got reset when getting out of a sit.  Now HMDLeanRecenterEnabled is only controlled by the scripts.
- Added Rig::getUnscaledHipsHeight (like getUnscaledEyeHeight).  Refactored a little to avoid duplicated code.  Added DEFAULT_AVATAR_HIPS_HEIGHT which is the value that Rig::getUnscaledHipsHeight returns when using the default avatar.
- (point 7 above) MyAvatar::endSit now passes false to centerBodyInternal for 'forceFollowYPos'.
- (point 8 above) MyAvatar::setAllowAvatarStandingPreference now calls centerBodyInternal with false for 'forceFollowYPos'.
- computeHipsInSensorFrame: The code now matches the comments in that it only skips the dampening of the hips rotation if the centre-of-gravity model is being used.
- (point 3 above) MyAvatar::scaleMotorSpeed: The action motor velocity was being scaled once too many by the sensor-to-world scale.
- MyAvatar's body is now recentred when foot-tracking starts or ends (in MyAvatar::update).  More info: 2be87fe7dcf47ff33560349fd36d2be6012cf30b
- Optimisation: MyAvatar::update now stores bools indicating which body parts are tracked (_isBodyPartTracked).  This avoids the unnecessary computations of using getControllerPoseInAvatarFrame to transform controller::Pose from sensor space to world space and then to avatar space, only to check if the pose was valid.
- Decoupled MyAvatar.centerOfGravityModelEnabled from the user standing state, because if artificial standing is disabled, we use centre-of-gravity leaning regardless of the user's sit/stand state (c5fe49bc3074b7d2ee02db04af34fdf1a241c357).
- Removed a "squat" recentre (different from the sitting recentre), that would happen after 30 seconds.  More info: 8d3e2ae9eecc456cabc46757c6b64e379427859f
- (point 10 above) See 3cc08022fcb8ed6b4aa7b9b6182df5a0af80e144

### Tests done

On Vive, with Developer > Avatar > Debug Draw Default Pose:

- Tested all combinations of "Allow my avatar to stand" and "Allow my avatar to lean" with real-world sitting and standing.
- Tested with foot trackers and hips tracker (by design, the avatar leaning/standing options now have no effect when feet are tracked).
- Tested head-relative and hand-relative walking, with and without foot trackers.  For point 5 above: using hand-relative walking, turn head sharply while walking in a straight line; previously you would swerve off-course momentarily.
- Tested flying.
- Tested click-to-sit and getting back up again, in all the stand/lean modes, with and without feet & hips tracking.
- Tested third-person and mirror-mode walking and flying.
- Tested with minimum and maximum avatar scaling.
- Tested user heights ranging from 1m to 2.1m.  In each case, with the HMD at the corresponding head height, the avatar was standing straight with its feet touching the floor, in all modes of 'Allow my avatar to stand'.
- Tested upside-down and radial gravity (merged the branch with Tivoli code and stood on the acorn at hifi://d72e6cae-d37b-1000-8000-679d336fa274/0.106048,3.49667,-45.2924/0.544102,0.108512,0.826886,0.0918574).

### Tests to do

- Test with other HMDs (other than Vive).
- Test a scripted motor on MyAvatar; check its velocity applies correctly.
- Riding on vehicles.

### Screenshots

image

### Click-to-sit

1. Some older domains used a click-to-sit script that wasn't VR-compatible.  All the seats in The Hub use a VR-compatible click-to-sit script, so that's where I'm testing.
2. If foot tracking is not enabled, all click-to-sit behaviour should be the same as before.
3. If foot tracking is enabled, we assume the user is already in the position in which they want to virtually sit when they click the seat.  It works nicely if they're sitting on the floor or on a real seat.  The hips are pinned to the seat as usual (can't move or rotate them).  If the user is standing when they click the seat, it will look wrong because the avatar will be standing inside the seat (we _could_ fix that to have them initially standing on the seat, but it would surely complicate the script and I don't particularly think it's worth it).
4. If hip tracking is enabled, the hips are not pinned to the seat.  The user can adjust their position and rotate on the seat.
5. In all cases, the virtual sitting ends when the desired hips position moves too far (on any translation axis) from the seat.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee.tar_6fc40385.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee_6fc40385.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee.tar_6fc40385.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee_6fc40385.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee_6fc40385.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-2aea3ee_6fc40385.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee_6fc40385.apk

Phil-Palmer:
BUG: With this Windows PR build, the 'Away' overlay is not disappearing when leaving the away state (eg. pressing W to walk).
https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2aea3ee_6fc40385.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124.tar_287f710e.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124_287f710e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124.tar_287f710e.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124_287f710e.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124_287f710e.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-9b20124_287f710e.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9b20124_287f710e.apk

Phil-Palmer:
BUG: The user's real-world height isn't being taken into account properly.  I'll work on this now.

JulianGro:
> Should the PR be marked 'do not merge' until that's fixed?
It's fine, it is missing CR and QA anyways. We only use that once a PR is stuck for the foreseeable future.

SilverfishVR:
> BUG: The user's real-world height isn't being taken into account properly. I'll work on that now. Should the PR be marked 'do not merge' until that's fixed?

Issue related to user height: https://github.com/vircadia/vircadia/issues/117

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974_8b839fe7.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974_8b839fe7.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974.tar_8b839fe7.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974.tar_8b839fe7.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-3ba0974_8b839fe7.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974_8b839fe7.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba0974_8b839fe7.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871.tar_a489e9dd.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871.tar_a489e9dd.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871_a489e9dd.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871_a489e9dd.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-3ba5871_a489e9dd.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871_a489e9dd.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-3ba5871_a489e9dd.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da_3e25e32f.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da.tar_3e25e32f.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da.tar_3e25e32f.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da_3e25e32f.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-d45c5da_3e25e32f.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da_3e25e32f.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-d45c5da_3e25e32f.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9_2be87fe7.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9.tar_2be87fe7.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9.tar_2be87fe7.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9_2be87fe7.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9_2be87fe7.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-9175db9_2be87fe7.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-9175db9_2be87fe7.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d.tar_1e54d41f.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d.tar_1e54d41f.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d_1e54d41f.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d_1e54d41f.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-439882d_1e54d41f.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d_1e54d41f.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-439882d_1e54d41f.exe

interface/src/avatar/MyAvatar.cpp
```diff
@@ -5499,65 +5584,83 @@ MyAvatar::FollowHelper::FollowHelper() {
...
+bool MyAvatar::FollowHelper::shouldActivateRotation(const MyAvatar& myAvatar,
+                                                    const glm::mat4& desiredBodyMatrix,
+                                                    const glm::mat4& currentBodyMatrix,
+                                                    bool& shouldSnapOut) const {
+    // If hips are under direct control (tracked), they give our desired body rotation and we snap to it every frame.
+    if (myAvatar.areHipsTracked()) {
+        shouldSnapOut = true;
+        return true;
+    }
+    else {
```
ctrlaltdavid, line 5642:
```suggestion
    } else {
```

interface/src/avatar/MyAvatar.cpp
```diff
@@ -3793,54 +3815,115 @@ glm::vec3 MyAvatar::scaleMotorSpeed(const glm::vec3 forward, const glm::vec3 rig
...
-                    right = glm::normalize(transform);
-                } else {
-                    right = Vectors::ZERO;
-                }
-                break;
-            case LocomotionRelativeMovementMode::MOVEMENT_HMD_RELATIVE:
-            default:
-                forward = IDENTITY_FORWARD;
-                right = IDENTITY_RIGHT;
+    const int movementReference = getMovementReference();
```
ctrlaltdavid, line 3827: Unnecessary `const` variables, here and elsewhere.

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07_7e5680b2.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07.tar_7e5680b2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07.tar_7e5680b2.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07_7e5680b2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07_7e5680b2.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-dd5fa07_7e5680b2.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-dd5fa07_7e5680b2.exe

libraries/animation/src/Rig.cpp
```diff
@@ -2783,19 +2789,36 @@ float Rig::getUnscaledEyeHeight() const {
...
         return DEFAULT_AVATAR_EYE_HEIGHT;
     }
 }
 
+// Get the vertical position of the hips joint, in the rig coordinate frame, ignoring the avatar scale.
+float Rig::getUnscaledHipsHeight() const {
+    // This factor can be used to scale distances in the geometry frame into the unscaled rig frame.
+    const float scaleFactor = GetScaleFactorGeometryToUnscaledRig();
+
+    const int hipsJoint = indexOfJoint("Hips");
```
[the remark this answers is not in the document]
Phil-Palmer: Thanks, fixed in 8c7c91ed6fad981f098953329e4d0d8bf4809812

interface/src/avatar/MyAvatar.cpp
Phil-Palmer: Thanks, fixed in 8c7c91ed6fad981f098953329e4d0d8bf4809812

interface/src/avatar/MyAvatar.h
```diff
@@ -413,8 +408,6 @@ class MyAvatar : public Avatar {
     Q_PROPERTY(float walkBackwardSpeed READ getWalkBackwardSpeed WRITE setWalkBackwardSpeed NOTIFY walkBackwardSpeedChanged);
     Q_PROPERTY(float sprintSpeed READ getSprintSpeed WRITE setSprintSpeed NOTIFY sprintSpeedChanged);
     Q_PROPERTY(bool isInSittingState READ getIsInSittingState WRITE setIsInSittingState);
-    Q_PROPERTY(MyAvatar::SitStandModelType userRecenterModel READ getUserRecenterModel WRITE setUserRecenterModel);
-    Q_PROPERTY(bool isSitStandStateLocked READ getIsSitStandStateLocked WRITE setIsSitStandStateLocked);
```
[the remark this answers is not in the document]
Phil-Palmer: Thanks, fixed in cba79c72f51338f80674c27649ef83801c1a004d

interface/src/avatar/MyAvatar.h
```diff
@@ -2841,23 +2873,15 @@ private slots:
...
-        void activate(FollowType type);
+        void deactivate(CharacterController::FollowType type);
+        void activate(CharacterController::FollowType type, const bool snapFollow);
         bool isActive() const;
-        bool isActive(FollowType followType) const;
-        float getMaxTimeRemaining() const;
+        bool isActive(CharacterController::FollowType followType) const;
         void decrementTimeRemaining(float dt);
-        bool shouldActivateRotation(const MyAvatar& myAvatar, const glm::mat4& desiredBodyMatrix, const glm::mat4& currentBodyMatrix) const;
+        bool shouldActivateRotation(const MyAvatar& myAvatar, const glm::mat4& desiredBodyMatrix, const glm::mat4& currentBodyMatrix, bool& shouldSnapOut) const;
```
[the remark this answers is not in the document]
Phil-Palmer: This one is an output parameter.  Do we have a convention for them that would make it clearer?

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb.tar_8cf76609.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb_8cf76609.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb.tar_8cf76609.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb_8cf76609.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb_8cf76609.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-95f51eb_8cf76609.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-95f51eb_8cf76609.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f_976ed2b1.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f.tar_976ed2b1.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f.tar_976ed2b1.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f_976ed2b1.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f_976ed2b1.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-f0d693f_976ed2b1.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f0d693f_976ed2b1.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf.tar_44324584.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf.tar_44324584.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf_44324584.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf_44324584.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf_44324584.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-f8014cf_44324584.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f8014cf_44324584.apk

interface/src/avatar/MyAvatar.h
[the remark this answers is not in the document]
ctrlaltdavid: Ah, I missed seeing that. : )

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c_ba5bea15.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c.tar_ba5bea15.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c_ba5bea15.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c.tar_ba5bea15.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-02dde7c_ba5bea15.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c_ba5bea15.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-02dde7c_ba5bea15.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8.tar_3cc08022.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8_3cc08022.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8.tar_3cc08022.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8_3cc08022.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8_3cc08022.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-2217db8_3cc08022.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-2217db8_3cc08022.apk

interface/src/avatar/MyAvatar.h
```diff
@@ -283,15 +283,17 @@ class MyAvatar : public Avatar {
...
      * @property {MyAvatar.SitStandModelType} userRecenterModel - Controls avatar leaning and recentering behavior.
-     * @property {number} isInSittingState - <code>true</code> if the user wearing the HMD is determined to be sitting
-     *     (avatar leaning is disabled, recentering is enabled), <code>false</code> if the user wearing the HMD is
-     *     determined to be standing (avatar leaning is enabled, and avatar recenters if it leans too far).
-     *     If <code>userRecenterModel == 2</code> (i.e., "auto") the property value automatically updates as the user sits
-     *     or stands, unless <code>isSitStandStateLocked == true</code>. Setting the property value overrides the current
-     *     sitting / standing state, which is updated when the user next sits or stands unless
-     *     <code>isSitStandStateLocked == true</code>.
+     *     <p class="important">Deprecated: This property is deprecated and will be removed. If you need it, please contact 
+     *     the developers.</p>
```
ctrlaltdavid, line 287:
```suggestion
     *     <p class="important">Deprecated: This property is deprecated and will be removed.</p>
```
Don't need to say "contact the developers", here an elsewhere.

interface/src/avatar/MyAvatar.h
```diff
@@ -413,8 +415,19 @@ class MyAvatar : public Avatar {
     Q_PROPERTY(float walkBackwardSpeed READ getWalkBackwardSpeed WRITE setWalkBackwardSpeed NOTIFY walkBackwardSpeedChanged);
     Q_PROPERTY(float sprintSpeed READ getSprintSpeed WRITE setSprintSpeed NOTIFY sprintSpeedChanged);
     Q_PROPERTY(bool isInSittingState READ getIsInSittingState WRITE setIsInSittingState);
+
+    /**jsdoc
+    * @deprecated This property is deprecated and will be removed. If you need it, please contact the developers.
+    */	
     Q_PROPERTY(MyAvatar::SitStandModelType userRecenterModel READ getUserRecenterModel WRITE setUserRecenterModel);
```
ctrlaltdavid, line 422:
```suggestion
    Q_PROPERTY(MyAvatar::SitStandModelType userRecenterModel READ getUserRecenterModel WRITE setUserRecenterModel);  // Deprecated
```
This JSDoc won't generate any documentation. The deprecation is documented adequately above. A simple comment after the property declaration suffices.
Similarly, elsewhere.

interface/src/avatar/MyAvatar.h
```diff
@@ -519,6 +532,8 @@ class MyAvatar : public Avatar {
 
     /**jsdoc
      * <p>Specifies different avatar leaning and recentering behaviors.</p>
+     * <p class="important">Deprecated: This enumeration is deprecated and will be removed. If you need it, please contact 
+     *     the developers</p>
```
ctrlaltdavid, line 536:
```suggestion
     * <p class="important">Deprecated: This type is deprecated and will be removed.</p>
```
There's no such thing as an "enumeration" in JavaScript. This JSDoc is listed in the "Types" section of the MyAvatar API docs page.

interface/src/avatar/MyAvatar.h
```diff
@@ -549,6 +564,30 @@ class MyAvatar : public Avatar {
...
+        WhenUserIsStanding,
+        Always,
+        Count,
+        Default = Always
+    };
+    Q_ENUM(AllowAvatarStandingPreference)
+
+    // Note: The option strings in setupPreferences (PreferencesDialog.cpp) must match this order.
+    enum class AllowAvatarLeaningPreference : uint
+    {
```
ctrlaltdavid, line 578:
```suggestion
    enum class AllowAvatarLeaningPreference : uint {
```

interface/src/avatar/MyAvatar.h
```diff
@@ -549,6 +564,30 @@ class MyAvatar : public Avatar {
...
+        WhenUserIsStanding,
+        Always,
+        Never,
+        AlwaysNoRecenter,  // experimental
+        Count,
+        Default = WhenUserIsStanding,
```
ctrlaltdavid, line 584:
```suggestion
        Default = WhenUserIsStanding
```

libraries/physics/src/CharacterController.h
```diff
@@ -53,7 +53,21 @@ const btScalar MIN_CHARACTER_MOTOR_TIMESCALE = 0.05f;
 class CharacterController : public btCharacterControllerInterface {
 
 public:
-    CharacterController();
+    enum class FollowType : uint8_t
+    {
```
ctrlaltdavid, line 57:
```suggestion
    enum class FollowType : uint8_t {
```

interface/src/avatar/MyAvatar.h
```diff
@@ -2922,6 +2981,9 @@ private slots:
 
     bool _centerOfGravityModelEnabled { true };
     bool _hmdLeanRecenterEnabled { true };
+    bool _hmdCrouchRecenterEnabled{
```
ctrlaltdavid, line 2984:
```suggestion
    bool _hmdCrouchRecenterEnabled {
```

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a_5d1bd5cc.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a.tar_5d1bd5cc.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a.tar_5d1bd5cc.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a_5d1bd5cc.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a_5d1bd5cc.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-6db421a_5d1bd5cc.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-6db421a_5d1bd5cc.apk

interface/src/avatar/MyAvatar.cpp
```diff
@@ -5159,16 +5234,46 @@ bool MyAvatar::getIsInWalkingState() const {
...
+    };
+
+    return legacySitStandModels[static_cast<uint>(_allowAvatarLeaningPreference.get())];
 }
 
+// Deprecated, will be removed.
 bool MyAvatar::getIsSitStandStateLocked() const {
-    return _lockSitStandState.get();
+    qCDebug(interfaceapp) << "MyAvatar.getIsSitStandStateLocked is deprecated and will be removed.  If you need it, please "
+                             "contact the developers.  See also: MyAvatar.getUserRecenterModel.";
```
ctrlaltdavid, line 5261: A remaining occurrence of "contact the developers".

ctrlaltdavid (changes requested):

interface/src/avatar/MyAvatar.cpp
Phil-Palmer: Thanks, fixed in a54a40dc2cb1555b4973489d83e57d449bc19c9c

ctrlaltdavid (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f94b3fd_265aeb76.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f94b3fd.tar_265aeb76.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f94b3fd.tar_265aeb76.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f94b3fd_265aeb76.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Quest-Alpha-PR928-f94b3fd_265aeb76.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR928/Vircadia-Alpha-PR928-f94b3fd_265aeb76.exe

interface/src/avatar/MyAvatar.cpp
Phil-Palmer: It's maybe a shame that we don't use `const` more though.  I like it because it tells the reader "this is the only line that sets this variable" :)

digisomni:
So I did a general set of tests and combinations on a Valve Index with Vive Trackers and all seems pretty well. Things work... intuitively. I didn't do an exhaustive test though, I simply tried out generally the things fixed in the PR and then tried doing things that a normal person would want. With my full body tracking, things worked... well.

SilverfishVR:
testing PR928-f94b3fd on Valve Index, no full body

seems to work overall 👍  but the different leaning/sitting is still a bit convoluted, I'm not entirely sure what is supposed to do what, that could do with some simplification in the future, for example, I think a easily accessible sit/stand toggle would be nice

When using hand relative movement the HMD direction is still used, would point 6 not change that? btw, I think it's super weird that when flying, head yaw controls strafe (translate L/R) and head roll controls turn L/R, is that really what we want?

speed when small/big does seem a lot more sensible now 👍 

I have not tested scripted movement.

Phil-Palmer:
Thanks @SilverfishVR , I know what you mean.  I was really careful to preserve the existing behaviour wherever possible, which is why I didn't do more of a redesign.  Instead, this PR takes the functionality that was already there, and makes it clearer in the interface and in the code.
Here's a function that illustrates that quite well; it translates the old "avatar leaning behaviors" into the simpler standing & leaning behaviours that each one implied (but which were not clear from their descriptions) :

```
// Deprecated, will be removed.
void MyAvatar::setUserRecenterModel(MyAvatar::SitStandModelType modelName) {
    qCDebug(interfaceapp)
        << "MyAvatar.setUserRecenterModel is deprecated and will be removed.";

    switch (modelName) {
        case SitStandModelType::ForceSit:
            setAllowAvatarStandingPreference(AllowAvatarStandingPreference::Always);
            setAllowAvatarLeaningPreference(AllowAvatarLeaningPreference::Never);
            break;
        case SitStandModelType::ForceStand:
            setAllowAvatarStandingPreference(AllowAvatarStandingPreference::Always);
            setAllowAvatarLeaningPreference(AllowAvatarLeaningPreference::Always);
            break;
        case SitStandModelType::Auto:
        default:
            setAllowAvatarStandingPreference(AllowAvatarStandingPreference::Always);
            setAllowAvatarLeaningPreference(AllowAvatarLeaningPreference::WhenUserIsStanding);
            break;
        case SitStandModelType::DisableHMDLean:
            setAllowAvatarStandingPreference(AllowAvatarStandingPreference::WhenUserIsStanding);
            setAllowAvatarLeaningPreference(AllowAvatarLeaningPreference::AlwaysNoRecenter);
            break;
    }
}
```

For the flying, I agree it would benefit from a redesign, but it would be outside the scope of this PR :)

You said that HMD direction should have an effect on hand-relative movement?  I don't think I'm aware of that - could you explain it please?
Thanks!

SilverfishVR:
yes no reason to rip it all out at once, and on a side not, I'm a VR geek at heart, so whenever someone gives VR support some love I'm ever grateful, so thank you for that.

I found that, when using _hand relative_ or _hand relative (leveled)_ my head orientation is still mixed in so my actual movement direction seems to be an average of the two, and well that is unexpected I think.

Phil-Palmer:
Thanks @SilverfishVR .  I just ran the PR build and tested hand-relative & levelled hand-relative walking, and it's working correctly for me: no influence from the HMD direction.  I'm trying to understand what you're seeing.

1. Were you talking about walking, or flying?  Hand-relative flying does have influence from the head (that's existing behaviour that I preserved).

2. Independent of the walk direction, the body angle follows the head in hand-relative walking (existing behaviour that I preserved).

3. Without this PR, hand-relative walking would swerve briefly when turning your head, but that's fixed in this PR (point 6).

Any info you could give to help me understand or reproduce the problem you're seeing would be greatly appreciated.  I'm only able to test on Vive.
Thanks!

JulianGro:
> 1. Were you talking about walking, or flying?  Hand-relative flying does have influence from the head (that's existing behaviour that I preserved).

While I personally would never use hand-relative directions, having it influenced by the head when flying while not when walking just seems wrong. It probably makes sense to have the head not influence the direction in that case, so the controls do not suddenly become alien when taking off to fly.

> 2. Independent of the walk direction, the body angle follows the head in hand-relative walking (existing behaviour that I preserved).

For the body angle, it makes sense to me that the head at least influences it even when using hand-relative walking. Maybe use the hand rotation up for the body angle up to a difference of something like 85° between hand and head rotation and then have anything more than that influence the body angle.
That way you can still look around while walking forwards without rotating your body, but won't have an unnatural head rotation in more extreme cases. Obviously it might still look a little weird if you are turning your hand rather than your head, so maybe it might make more sense to have a curve for the body angle influence rather than a hard limit.

Phil-Palmer:
Thanks @JulianGro .  There are plenty of things to improve in hifi's VR locomotion :)  Those are best tackled in separate PRs after this one, and they'll be somewhat easier now that this PR has made the code clearer.

SilverfishVR:
yes flying, actually I did not notice it did not do it when walking, I turned it off quickly since I find it quite annoying 😄 though with some modification, it could be neat for that superman flying experience.

So it seems that the PR behaves as expected, we just have more work to do.


> 2\. Independent of the walk direction, the body angle follows the head in hand-relative walking (existing behavior that I preserved).

🤔 I would have sworn this was addressed following user request to be able to walk alongside someone and turn your head to look at them like irl (for cinematic shots)

Issue #793, opened 2020-10-09
Enhancement/FeatureRequest: Optional preserve point of origin from models.


Aitolda:
At the moment, the interface uses volume to determine the center point of a model.  If we were to simply preserve the original origin point from the fbx or gltf, most current assets would break.  However, if we leave it at volume by default, and add an optional checkbox to the model properties for preserve origin, this would allow the existing assets to continue functioning.

PR #951, opened 2021-01-02
Model Entities: useOriginalPivot Property

HifiExperiments:
implements #284 resolves https://github.com/vircadia/vircadia/issues/793

adds a new property to Model Entities: useOriginalPivot

defaults to false (current behavior), but adds option to toggle in new model dialog

caveat: in create, the bounding box of a selected entity is centered around the actual entity position, which is now offset from the model's position when useOriginalPivot is true.  I actually like this, as it reflects that offset, but maybe others have different opinions

Test plan:
- run this script: https://gist.github.com/HifiExperiments/7b3faa7919eed7a308c762ffe03d893f/raw/615f87b8315e433ac35b23e47a52ab6954e27f6b/PivotTest.js
- first and third models have useOriginalPivot false, second and fourth are true.  First and second models use natural dimensions, third and fourth use { 1, 1, 2 }
- enable Developer -> Physics -> Show Bullet Collisions.  green collision lines should match the 4 models
- hover over the entities with your mouse.  a small white sphere will appear when your mouse is over the entities.  make sure it lines up with the models
- press "o" to toggle through different registrationPoints for all of the models.  note that they show the correct behavior, and that the physics and ray picking continue to work on all 4 models
- verify the workflow of adding a model with/without "Use Original Pivot" via create, and that you can toggle it on a created model via the property window
- run luci.js and use the Bounding Box section to confirm that the render bounds of each model is in the correct place, encompassing it regardless of pivot


interface/src/ui/Stats.h
```diff
@@ -248,9 +248,9 @@ private: \
  *     <em>Read-only.</em>
  * @property {string} lodStatus - Description of the current LOD.
  *     <em>Read-only.</em>
- * @property {string} numEntityUpdates - The number of entity updates that happened last frame.
+ * @property {number} numEntityUpdates - The number of entity updates that happened last frame.
```
HifiExperiments, line 251: this (and some below) is unrelated cleanup to silence some warnings and fixup this JSDoc

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -162,9 +164,10 @@ void RenderableModelEntityItem::updateModelBounds() {
 
         // now recalculate the bounds and registration
         model->setScaleToFit(true, scaledDimensions);
-        model->setSnapModelToRegistrationPoint(true, registrationPoint);
+        model->setSnapModelToRegistrationPoint(!_useOriginalPivot, registrationPoint);
         updateRenderItems = true;
-        model->scaleToFit();
```
HifiExperiments, line 167: I don't think we need to call scaleToFit here because we call simulate below which calls scaleToFit as necessary

libraries/entities-renderer/src/RenderableModelEntityItem.cpp
```diff
@@ -452,7 +465,7 @@ void RenderableModelEntityItem::computeShapeInfo(ShapeInfo& shapeInfo) {
                 pointCollection[i][j] = scaleToFit * (pointCollection[i][j] + model->getOffset()) - registrationOffset;
             }
         }
-        shapeInfo.setParams(type, 0.5f * extents, getCompoundShapeURL());
+        shapeInfo.setParams(type, 0.5f * extents, getCompoundShapeURL() + model->getSnapModelToRegistrationPoint());
```
HifiExperiments, line 468: this is a hack to make sure that the same model with/without useOriginalPivot know that they have slightly different physics shapes (the difference being the offset)

libraries/entities-renderer/src/RenderablePolyVoxEntityItem.cpp
```diff
@@ -723,8 +723,10 @@ ShapeType RenderablePolyVoxEntityItem::getShapeType() const {
 }
 
 void RenderablePolyVoxEntityItem::setRegistrationPoint(const glm::vec3& value) {
-    if (value != _registrationPoint) {
```
HifiExperiments, line 726: there's some cleanup below to make sure this stuff is all threadsafe and takes into account the pivot

libraries/entities/src/EntityItemProperties.h
```diff
@@ -302,6 +302,7 @@ class EntityItemProperties {
     DEFINE_PROPERTY(PROP_RELAY_PARENT_JOINTS, RelayParentJoints, relayParentJoints, bool, ENTITY_ITEM_DEFAULT_RELAY_PARENT_JOINTS);
     DEFINE_PROPERTY_REF(PROP_GROUP_CULLED, GroupCulled, groupCulled, bool, false);
     DEFINE_PROPERTY_REF(PROP_BLENDSHAPE_COEFFICIENTS, BlendshapeCoefficients, blendshapeCoefficients, QString, "");
+    DEFINE_PROPERTY_REF(PROP_USE_ORIGINAL_PIVOT, UseOriginalPivot, useOriginalPivot, bool, false);
```
digisomni, line 305:
```suggestion
    DEFINE_PROPERTY_REF(PROP_USE_ORIGINAL_PIVOT, UseOriginalPivot, useOriginalPivot, bool, true);
```
If using the original pivot is the standard, we should probably respect that and encourage doing the same going forward. What you export from Blender should generally show up in the same fashion in-world if at all possible.

Though, is it set to false because all existing entities without this property will get populated with false (in order to prevent breakage?)

In that case, I guess it'll have to stay false, the Create app will have to set it to "true" by default, and then we'll have to plan to set the default value to true at some point further down the line...

HifiExperiments: yeah, the default needs to be false because of old content.  I can change the default in the new model dialog to be true, but there’s no easy way to change the default in the future (without breaking lots of stuff).

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-a55be64.tar_20e0c241.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-a55be64_20e0c241.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-a55be64.tar_20e0c241.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-a55be64_20e0c241.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-a55be64_20e0c241.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-a55be64_20e0c241.apk

libraries/entities/src/EntityItemProperties.h
digisomni: Does the domain server not have a way to parse the entity tree and upgrade them to latest? Therefore if an old entity is parsed, set it to false, and then we can set it to true by default in all places for new entities.

HifiExperiments: we have a way to do this for domain content/backups, so a domain that’s upgrading to a new version could properly detect it, but there’s no way to do it for things like scripts that could be doing anything

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97_5dc45e89.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97_5dc45e89.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97.tar_5dc45e89.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97.tar_5dc45e89.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-100ee97_5dc45e89.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97_5dc45e89.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-100ee97_5dc45e89.exe

libraries/render-utils/src/Model.cpp
```diff
@@ -1182,17 +1198,8 @@ glm::vec3 Model::getNaturalDimensions() const {
...
-    const Extents& extents = getHFMModel().meshExtents;
-
-    // even though our caller asked for "unscaled" we need to include any fst scaling, translation, and rotation, which
-    // is captured in the offset matrix
-    glm::vec3 minimum = glm::vec3(getHFMModel().offset * glm::vec4(extents.minimum, 1.0f));
-    glm::vec3 maximum = glm::vec3(getHFMModel().offset * glm::vec4(extents.maximum, 1.0f));
-    Extents scaledExtents = { minimum * _scale, maximum * _scale };
-    return scaledExtents;
+    Extents extents = getUnscaledMeshExtents();
+    return { extents.minimum * _scale, extents.maximum * _scale};
```
ctrlaltdavid, line 1202:
```suggestion
    return { extents.minimum * _scale, extents.maximum * _scale };
```

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -1733,6 +1734,7 @@ QScriptValue EntityItemProperties::copyToScriptValue(QScriptEngine* engine, bool
         COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_RELAY_PARENT_JOINTS, relayParentJoints);
         COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_GROUP_CULLED, groupCulled);
         COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_BLENDSHAPE_COEFFICIENTS, blendshapeCoefficients);
+        COPY_PROPERTY_TO_QSCRIPTVALUE(PROP_USE_ORIGINAL_PIVOT, useOriginalPivot);
```
ctrlaltdavid, line 1737: JSDoc for this new property is needed. See the JSDoc above `ScriptValue EntityItemProperties::copyToScriptValue()` in this file.

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-1467eb9_030524c3.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.exe

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-1467eb9_030524c3.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9.tar_030524c3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9.tar_030524c3.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-1467eb9_030524c3.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1467eb9_030524c3.apk

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -1002,6 +1003,8 @@ EntityPropertyFlags EntityItemProperties::getChangedProperties() const {
  * @property {string} blendshapeCoefficients - A JSON string of a map of blendshape names to values.  Only stores set values.
  *     When editing this property, only coefficients that you are editing will change; it will not explicitly reset other
  *     coefficients.
+ * @property {boolean} useOriginalPivot=false - If false, the model will be centered based on its content, ignoring any offset
+ *     in the model itself. If true, the model will respect its original offset.
```
ctrlaltdavid, line 1007:
```suggestion
 * @property {boolean} useOriginalPivot=false - If <code>false</code>, the model will be centered based on its content,
 *     ignoring any offset in the model itself. If <code>true</code>, the model will respect its original offset.
```

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2.tar_1ccdfe37.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2.tar_1ccdfe37.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-322c9d2_1ccdfe37.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2.tar_1ccdfe37.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2.tar_1ccdfe37.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-322c9d2_1ccdfe37.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-322c9d2_1ccdfe37.apk

ctrlaltdavid (approved):

digisomni:
For testing purposes, let's get the Create app changes to set this property to true by default merged into this branch (from someone, anyone) so that it all gets shipped with the PR and isn't forgotten.

HifiExperiments:
@digisomni that’s all already included!

digisomni:
Nevermind!

ctrlaltdavid (approved):

Aitolda:
Works as intended if you use when of the generated collisions types like basic, exact, etc.. BUT if you use something like Box or Sphere collision center will be the original position.  Not necessarily a bug, but this might lead to some confusion.  Not sure how we should address this.

HifiExperiments:
@Aitolda thanks for catching that, I think it should be fixed in the latest commit, but of course windows builds are broken here too

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.sh

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-01de369_7bed9b07.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.apk

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369.tar_7bed9b07.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-01de369_7bed9b07.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-01de369_7bed9b07.apk

Aitolda:
Pivot point of compound shapes still doesn't seem to be working, at least on the offset cube model I've been using.

[Offset Cube](https://aitolda.s3.amazonaws.com/test/Offset.glb)

HifiExperiments:
@Aitolda @digisomni ok hopefully NOW it's working 😛

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322.tar_ace6bf6c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322.tar_ace6bf6c.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322_ace6bf6c.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322_ace6bf6c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322_ace6bf6c.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-1510322_ace6bf6c.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Alpha-PR951-1510322_ace6bf6c.apk

Aitolda:
Still some major issues with using compound collisions, I see two problems potentially working together.

1. on an fbx model (only) the compound collision hull is rotated on the wrong axis.
2. on both fbx and glb,  if the model pivot point is moved away from 0,0,0 within a 3d application before exporting, the compound collision may be ignoring this offset while the main mesh respects it

I've attached several models which should help demonstrate the issues.  Unfortunately html doesn't like my naming scheme so I will label them here.  Note, in blender Z is up, and the coordinates in my naming scheme reflects this.

[ArrowUp(0,0,0)Blender](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%280%2C0%2C0%29Blender.png)
[ArrowUp(0,0,0).fbx](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%280%2C0%2C0%29.fbx)
[ArrowUp(0,0,0).glb](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%280%2C0%2C0%29.glb)

[ArrowUp(-5,-5,0)Blender](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%285%2C-5%2C0%29Blender.png)
[ArrowUp(-5,-5,0).fbx](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%28-5%2C-5%2C0%29.fbx)
[ArrowUp(-5,-5,0).glb](https://aitolda.s3.amazonaws.com/test/PivotProblems/ArrowUp%28-5%2C-5%2C0%29.glb)

Aitolda:
Thinking about it more, I'm not sure respecting the 0,0,0 (as mentioned in the post above) is even desirable.  it would make sense only if you were making complex environments most of the time, and I think a json exported is a better approach for this.  Just want to clarify we're talking about the distance of the pivot point from zero in blender or 3dmax,   not the pivot point itself, which is the whole point of this pr.

Aitolda:
> Thinking about it more, I'm not sure respecting the 0,0,0 (as mentioned in the post above) is even desirable. it would make sense only if you were making complex environments most of the time, and I think a json exported is a better approach for this. Just want to clarify we're talking about the distance of the pivot point from zero in blender or 3dmax, not the pivot point itself, which is the whole point of this pr. In other words, I think with the exception of the wrong rotation on fbx, the compound shape is doing what it's supposed to, and it's the mesh that's incorrectly respecting the distance from 0

Regarding treating 0,0,0 in blender/Maya/3dMax export as the pivot point, after doing some digging around, it unfortunately looks like even Unreal and Unity have adopted this.  I assume their is some reason, but I think it's one of those things in the grand scheme are actually more difficult on artists who want to bring in elaborate scenes, but perhaps we can tackle that issue by updating the blender plugin.  Either way I think once the rotation issue is solved on fbx we should move forward as is for the time being.

HifiExperiments:
after a conversation with @Aitolda, he found that other engines also have problems with non-zero pivot points (https://forum.unity.com/threads/wrong-pivot-point-from-blender-to-unity.423176/).  for now, we think it's ok to move forward with this, and if we fix it in the future, we can try to do so in a backwards-compatible way.  after this is merged, I'll open an issue tracking that

digisomni:
Okeydokeseys!

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330_766b91f1.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330.tar_766b91f1.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330_766b91f1.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330.tar_766b91f1.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330_766b91f1.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-Quest-Alpha-PR951-6750330_766b91f1.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR951/Vircadia-PR951-6750330_766b91f1.apk

AleziaKurdis:
Just tested this. (I have not read any of the comments in this thread yet)
Here are my impressions:

The feature works as expected. I tested it with fbx and glb

But where it hurts is about having this behavior as default value in the UI.
It looks like this with an ex-centric pivot:
image

I don't know how other tools deal with this but is the bounding box is clearly not design to support that.
This is also ruins a bit the "snap to surface" feature. Is there a way to figure where is the edge of the model anymore?
A lot of stuff are going to behave weirdly if we have this by default.

I think we should try to figure if honoring the original pivot is the mainstream usecase before decide if this needs to be the default option. What are the usecase where you need more the original pivot than a consistent central point?

In my example the pivot was largely outside. BI suspect that in many cases, people will just have it approximatively in the middle, but without being mathematically in the center. So they might get the bounding box offset in a significant number of cases.
I would have let this option for the case where it is intentional (not as default). This is my opinion.

In the create app (Create Model), the option is not persisted using a setting like what we have for "grabbable".
You can be sure that I will minimally plug such a setting in the future to not have this defaulted (minimally for my-self).
I see advantage for this feature for doors, or anything that we need an ex-centric pivot. Which is not the main usecase for me.

I tested the object as dynamic... here too, excentricity doesn't appears to be a main desired behavior. 

**To recap:**
- Feature is good
- Set by Default in UI: Not sure about this. _(I assume it is not by the api, because that could ruin scripts)_
- Set by Default in UI: create app behave weirdly which is not going to reflect stability for new comers.
- A setting in create app would be a must. (Consider to log an issue for this)

HifiExperiments:
@AleziaKurdis like I said in the description, I actually think the bounding box thing in create is somewhat helpful, as it reflects the ACTUAL position of the entity, but I agree it’s another layer of complexity.  I’m open to suggestions on how to fix this...we could expose the offset to create and draw two different boxes maybe?

snap to surface should still work, since it uses findRayIntersection, right?  that should still work, even on pivoted models.

I agree the setting should be persistent!

Kalila wanted to get this in for the next release but I can sync with you to fix any bugs you find or make any improvements in a follow up

in general, I think this feature request is less about eccentricity for rotations and more about making building complex scenes easier for modelers

PR #966, opened 2021-01-17
Feature/qt macos update

ArcadeFever:
Update MacOSX build to QT 5.15.2 along with updated build instructions

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,6 +1,6 @@
 # General Build Information
 
-*Last Updated on August 26, 2020*
+*Last Updated on January 16, 2021*
 
 ### OS Specific Build Guides
 
--- a/BUILD.md
+++ b/BUILD.md
@@ -11,7 +11,7 @@
 
 ### Dependencies
 - [git](https://git-scm.com/downloads): >= 1.6  
-- [cmake](https://cmake.org/download/):  3.9
+- [cmake](https://cmake.org/download/):  3.9 (or greater up to 3.18.x)
 - [Python](https://www.python.org/downloads/): 3.6 or higher
 - [Node.JS](https://nodejs.org/en/): >= 12.13.1 LTS
     - Used to build the Screen Sharing executable.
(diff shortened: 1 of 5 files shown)


BUILD.md
```diff
@@ -21,7 +21,7 @@
 These dependencies need not be installed manually. They are automatically downloaded on the platforms where they are required.
 - [Bullet Physics Engine](https://github.com/bulletphysics/bullet3/releases):  2.83  
 - [glm](https://glm.g-truc.net/0.9.8/index.html):  0.9.8  
-- [Oculus SDK](https://developer.oculus.com/downloads/):   1.11 (Windows) / 0.5 (Mac)  
+- [Oculus SDK](https://developer.oculus.com/downloads/):   23.0 (Android)
```
JulianGro, line 24: Oculus SDK is used on Windows, unless it got removed in the last couple of months which is doubt.

JulianGro (changes requested):

BUILD.md
ctrlaltdavid: Yes, I think this PR shouldn't be changing anything to do w.r.t. Oculus?

screenshare/package-lock.json
```diff
@@ -1,8 +1,2732 @@
 {
   "name": "highfidelity_screenshare",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "highfidelity_screenshare",
+      "version": "1.0.0",
```
ctrlaltdavid, line 9: I don't think Screenshare's package-lock.json file needs any changes in this PR?

ctrlaltdavid (changes requested):

BUILD.md
ArcadeFever: Good call out, I wasn't aware of any specific Oculus support in Windows and Mac, and thought this from the pre-Oculus (Rift) days.  As this is the general build doc, it could also apply to Android so I put in the Oculus SDK we updated on Android.  I can revert this change if Windows/Mac support was for something like Oculus controllers only.

screenshare/package-lock.json
ArcadeFever: I have no idea how this got in there, I thought I reverted it before submitting.

ctrlaltdavid:
Closed in favour of https://github.com/vircadia/vircadia/pull/968

PR #967, opened 2021-01-17
Update BUILD_WIN.md

liberza:
Add MSVC v142 to Visual Studio 2019 install, as it was also required for me.


digisomni:
@ctrlaltdavid was it VS 2015 or 2017 that we just ditched entirely? I could have sworn there was some level of weird issues and we talked about only really supporting the latest version (2019)?

Issue #92, opened 2020-01-14
Cache busting for scripts (and other assets) needs to be revisited.


MarcusLlewellyn:
One of the things we've gotten used to but is really a pain for both experienced and new user alike is that scripts and other objects are aggressively cached, and not updated when new version are present.

daleglass:
Can we have a reproduction for testing and experimenting with this?

daleglass:
Suggestion by Odysseus: have an internal timeout so that if something is requested multiple times really quickly the asset is only requested once

daleglass:
#1012 may address part of this

daleglass:
yes

Issue #990, opened 2021-01-30
Script caching isn't working properly


ctrlaltdavid:
Checking whether script file has changed and needs to be redownloaded (instead of just using whatever's in cache).
Respecting script file expiry dates (i.e., redownload if expired).
Cache-busting.

digisomni:
This happens with models too! Generally requests and caching aren't really checked I think according to the HTTP spec.

HifiExperiments:
I’ve looked into this before and I think the cachebusting etc is necessary because we force our HTTP requests to use qnetworkrequest::PreferCache in https://github.com/vircadia/vircadia/blob/ec1c8f2b87187c5be8f5d208b539b6848ebd97b4/libraries/networking/src/HTTPResourceRequest.cpp

the default value, PreferNetwork, might just solve a lot of these problems: https://doc.qt.io/qt-5/qnetworkrequest.html#CacheLoadControl-enum

HifiExperiments:
https://github.com/vircadia/vircadia/issues/92

daleglass:
Dev meeting: Need to check what exactly does PreferNetwork do: does it go to the server every time to check the timestamp?

daleglass:
Also, do we store expiration times in our cache?

HifiExperiments:
it *sounds* like even PreferCache respects expiry times, but I guess the only way to find out is to try it

I would assume hifi originally chose PreferCache because it means fewer network requests to check timestamps, but the trade off seems like it’s not worth it.  or perhaps if the extra requests are too much, we have a “developer mode” that lets you temporarily switch to PreferNetwork while testing models/scripts, but most people still use PreferCache.  there’s already a switch in the code, but the alternative is AlwaysNetwork, which doesn’t seem like what we want either

PR #1012, opened 2021-02-09
Experiment with Qt HTTP caching options

HifiExperiments:
potentially closes https://github.com/vircadia/vircadia/issues/92 and https://github.com/vircadia/vircadia/issues/990

should theoretically
1. remove the need for cachebusting
2. improve loading times

Test plan/open questions:
- confirm that cachebusting scripts/models is no longer necessary (links won't update automatically, you'll still need to refresh the script or modelURL).  but...we might need to add a refresh button in Create next to modelURL, because otherwise it will probably ignore model URL changes before even checking the network?
- compare loading times (initial/loading from cache) between this PR and master.  is this PR faster?
- does it respect HTTP expiry times?  can someone smarter than me make a test for this

### DIFF
--- a/libraries/networking/src/AssetClient.cpp
+++ b/libraries/networking/src/AssetClient.cpp
@@ -19,7 +19,6 @@
 #include <QtScript/QScriptEngine>
 #include <QtNetwork/QNetworkDiskCache>
 
-#include <shared/GlobalAppProperties.h>
 #include <shared/MiniPromises.h>
 
 #include "AssetRequest.h"
@@ -35,7 +34,6 @@
 MessageID AssetClient::_currentID = 0;
 
 AssetClient::AssetClient() {
-    _cacheDir = qApp->property(hifi::properties::APP_LOCAL_DATA_PATH).toString();
     setCustomDeleter([](Dependency* dependency){
         static_cast<AssetClient*>(dependency)->deleteLater();
     });
@@ -55,34 +53,6 @@ AssetClient::AssetClient() {
     connect(nodeList.data(), &LimitedNodeList::nodeKilled, this, &AssetClient::handleNodeKilled);
     connect(nodeList.data(), &LimitedNodeList::clientConnectionToNodeReset,
             this, &AssetClient::handleNodeClientConnectionReset);
-}
-
-void AssetClient::initCaching() {
-    Q_ASSERT(QThread::currentThread() == thread());
-
-    // Setup disk cache if not already
-    auto& networkAccessManager = NetworkAccessManager::getInstance();
-    if (!networkAccessManager.cache()) {
-        if (_cacheDir.isEmpty()) {
-#ifdef Q_OS_ANDROID
-            QString cachePath = QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
-#else
-            QString cachePath = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
-#endif
-            _cacheDir = !cachePath.isEmpty() ? cachePath : "interfaceCache";
-        }
-        QNetworkDiskCache* cache = new QNetworkDiskCache();
-        cache->setMaximumCacheSize(MAXIMUM_CACHE_SIZE);
-        cache->setCacheDirectory(_cacheDir);
-        networkAccessManager.setCache(cache);
-        qInfo() << "ResourceManager disk cache setup at" << _cacheDir
-                 << "(size:" << MAXIMUM_CACHE_SIZE / BYTES_PER_GIGABYTES << "GB)";
-    } else {
-        auto cache = qobject_cast<QNetworkDiskCache*>(networkAccessManager.cache());
-        qInfo() << "ResourceManager disk cache already setup at" << cache->cacheDirectory()
-                << "(size:" << cache->maximumCacheSize() / BYTES_PER_GIGABYTES << "GB)";
-    }
-
 }
 
 namespace {
--- a/libraries/networking/src/HTTPResourceRequest.cpp
+++ b/libraries/networking/src/HTTPResourceRequest.cpp
@@ -58,7 +58,7 @@ void HTTPResourceRequest::doSend() {
     networkRequest.setHeader(QNetworkRequest::UserAgentHeader, NetworkingConstants::VIRCADIA_USER_AGENT);
 
     if (_cacheEnabled) {
-        networkRequest.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferCache);
+        networkRequest.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferNetwork);
     } else {
         networkRequest.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::AlwaysNetwork);
     }
--- a/libraries/networking/src/NetworkAccessManager.cpp
+++ b/libraries/networking/src/NetworkAccessManager.cpp
@@ -11,18 +11,42 @@
 
 #include "NetworkAccessManager.h"
 
-#include <QThreadStorage>
-
 #include "AtpReply.h"
+#include "ResourceCache.h"
+
+#include <QStandardPaths>
+#include <QThreadStorage>
 #include <QtNetwork/QNetworkProxy>
+#include <QtNetwork/QNetworkDiskCache>
+
+#include <shared/GlobalAppProperties.h>
 
 QThreadStorage<QNetworkAccessManager*> networkAccessManagers;
 
 QNetworkAccessManager& NetworkAccessManager::getInstance() {
     if (!networkAccessManagers.hasLocalData()) {
-        networkAccessManagers.setLocalData(new QNetworkAccessManager());
+        auto networkAccessManager = new QNetworkAccessManager();
+        // Setup disk cache if not already
+        if (!networkAccessManager->cache()) {
+            QString cacheDir = qApp->property(hifi::properties::APP_LOCAL_DATA_PATH).toString();
+            if (cacheDir.isEmpty()) {
+#ifdef Q_OS_ANDROID
+                QString cachePath = QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
+#else
+                QString cachePath = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
+#endif
+                cacheDir = !cachePath.isEmpty() ? cachePath : "interfaceCache";
+            }
+            QNetworkDiskCache* cache = new QNetworkDiskCache();
+            cache->setMaximumCacheSize(MAXIMUM_CACHE_SIZE);
+            cache->setCacheDirectory(cacheDir);
+            networkAccessManager->setCache(cache);
+            qInfo() << "NetworkAccessManager disk cache set up at" << cacheDir
+                     << "(size:" << MAXIMUM_CACHE_SIZE / BYTES_PER_GIGABYTES << "GB)";
+        }
+        networkAccessManagers.setLocalData(networkAccessManager);
     }
-    
+
     return *networkAccessManagers.localData();
 }
 
--- a/libraries/networking/src/AssetClient.h
+++ b/libraries/networking/src/AssetClient.h
@@ -64,8 +64,6 @@ public:
     Q_INVOKABLE AssetUpload* createUpload(const QByteArray& data);
 
 public slots:
-    void initCaching();
-
     void cacheInfoRequest(QObject* reciever, QString slot);
     MiniPromise::Promise cacheInfoRequestAsync(MiniPromise::Promise deferred = nullptr);
     MiniPromise::Promise queryCacheMetaAsync(const QUrl& url, MiniPromise::Promise deferred = nullptr);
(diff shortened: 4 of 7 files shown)


libraries/networking/src/AssetClient.cpp
```diff
@@ -57,34 +55,6 @@ AssetClient::AssetClient() {
             this, &AssetClient::handleNodeClientConnectionReset);
 }
 
-void AssetClient::initCaching() {
```
HifiExperiments, line 60: more info below, but for some reason, it seems like we were only setting up caching for this AssetClient, not for all other requests?  this code moved to NetworkAccessManager::getInstance()

libraries/networking/src/HTTPResourceRequest.cpp
```diff
@@ -58,7 +58,7 @@ void HTTPResourceRequest::doSend() {
     networkRequest.setHeader(QNetworkRequest::UserAgentHeader, NetworkingConstants::VIRCADIA_USER_AGENT);
 
     if (_cacheEnabled) {
-        networkRequest.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferCache);
+        networkRequest.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferNetwork);
```
HifiExperiments, line 61: this *should* theoretically fix the cachebusting issue, see the docs for more info: https://doc.qt.io/qt-5/qnetworkrequest.html#CacheLoadControl-enum

libraries/networking/src/NetworkAccessManager.cpp
```diff
@@ -11,18 +11,42 @@
...
+                QString cachePath = QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
+#else
+                QString cachePath = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
+#endif
+                cacheDir = !cachePath.isEmpty() ? cachePath : "interfaceCache";
+            }
+            QNetworkDiskCache* cache = new QNetworkDiskCache();
+            cache->setMaximumCacheSize(MAXIMUM_CACHE_SIZE);
+            cache->setCacheDirectory(cacheDir);
+            networkAccessManager->setCache(cache);
```
HifiExperiments, line 43: according to the docs: https://doc.qt.io/qt-5/qnetworkaccessmanager.html#setCache

"QNetworkAccessManager by default does not have a set cache."

so...I think we were forcing PreferCache above, but then...not actually setting up a cache?  hopefully this new code is more correct and more performant

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58_b8b9cdc5.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58.tar_b8b9cdc5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58.tar_b8b9cdc5.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58_b8b9cdc5.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-Quest-Alpha-PR1012-2c4fd58_b8b9cdc5.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58_b8b9cdc5.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-2c4fd58_b8b9cdc5.exe

JulianGro:
So originally when we discussed cachebusting the idea was turned down because of concerns about web servers being set up properly to support that feature.
While I was always for cachebusting, has the discussion happened again and the result changed this time?

digisomni:
This is implementing I think exactly what we were talking about. Webservers do have cache information in them, but currently we do not respect that. This gets us a bit closer to respecting that where possible.

HifiExperiments:
yeah I think this is a pretty standard feature?  this PR basically amounts to “use more default Qt values, and use them more correctly”

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882_b9b801f8.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882.tar_b9b801f8.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882_b9b801f8.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882.tar_b9b801f8.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882_b9b801f8.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-3879882_b9b801f8.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-Quest-Alpha-PR1012-3879882_b9b801f8.apk

Aitolda:
Seems to work. A refresh button would definitely be useful. I uploaded a model,then modified is slightly in blender, reuploaded to my S3, then removed the link and repasted and the model.  The updated model appeared without cache-busting.

digisomni:
One way to do this would be to have the refresh button empty the field and then paste back in the URL quickly (hacky method). The other method might be to have an actual scripting API call that allows you to just trigger a reload for a resource. Given that the system handles that pretty well...

It's possible that multiple parts of the system are looking at the URL for a change, but I'm hoping that it's simply the case that the resource being refeteched would update all layers. As for falling out of sync with others, it can send a message for others' Create app's to do the same call on that model?

HifiExperiments:
yeah for now the best way to do this would be to set the URL to "" and then set it back to the actual URL, just to make sure everything picks up on the change.  we might just need to copy how it works for the script field?  someone who understands create more than me might need to take that over

JulianGro:
Theoretically you could lose the URL if the reload button removes the URL and adds it back. On Linux this happens often enough.

ctrlaltdavid:
Note: In the Create app, the client script and server entity script fields both have a "reload" button.

HifiExperiments:
ok added a reload button, ready for testing

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d-.tar_2fed5124.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d-.tar_2fed5124.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d-_2fed5124.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d-_2fed5124.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d-_2fed5124.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-fee508d_2fed5124.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-Quest-Alpha-PR1012-fee508d_2fed5124.apk

Aitolda:
Refresh seems to work, but its worth noting the the new object will retain the dimensions or the original (reset dimensions will correct this).  This is not a bug, and is desirable in many cases, but worth considering in the future.

daleglass (approved):

ctrlaltdavid:
Compared loading The Hub content on this PR versus 2021.1.0. There doesn't seem to be any noticeable difference for me,

digisomni:
> Compared loading The Hub content on this PR versus 2021.1.0. There doesn't seem to be any noticeable difference for me,

Is there any way to make downloads work even if links redirect? I'm not sure how the system works or how things like that work in general.

However, between caching improvements and redirecting users to downloads that are on edge servers I think we can get fast loading times.

JulianGro:
> Compared loading The Hub content on this PR versus 2021.1.0. There doesn't seem to be any noticeable difference for me,

My understanding is that, load times should be same or worse. Old behaviour is to just never update the cache. So whatever you downloaded would not change even if the source file changed. My understanding is that this PR contacts the server to find out if the cache is outdated.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159-_2fed5124.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159-_2fed5124.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159-.tar_2fed5124.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159-.tar_2fed5124.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159-_2fed5124.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-d6b3159_2fed5124.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-Quest-Alpha-PR1012-d6b3159_2fed5124.apk

digisomni:
Test with webserver caching options set to no cache...

digisomni:
So, I've found that the refresh button doesn't update for file:/// or https:// cases, did not try others but basically I see the model blink as if it's reloading except the new model doesn't show up. 

I test this by... 
* renaming my model on the server after loading it once
* hitting refresh in the Create app, but the model stays the same it was even though it should be a whole new model, just with the exact same filename and path

However, the test for cachebusting in general worked successfully...
* rename the model on the server after loading it once
* go to the tutorial world under "Help" -> "Tutorial" then return back to the world
* notice that the model is now the new model

HifiExperiments:
hmmm how is that different than what @Aitolda tried originally and reported success for?

Aitolda:
So I finally got to digging into this again and now I'm simply confused.  I tried the newest version of this PR and couldn't get it working, so I started backtracking all the way to the first release on Feb 9, which was on the old protocol and it worked suddenly.  Got excited, tried again, and it didn't work.  Frustrated, I tried again and this time it worked.  Seems to be inconsistent. 

Here's what I've been doing.

I have two models that are both simply named "cube.glb" One of these is rotated off axis.
I load up one, confirm it's working.  Go to the model box in the properties dialogue and remove the link or add some gibberish.
SOMETIMES the box becomes a green outline as expected. sometimes there is no effect on screen and the model remains.

Next I replace the original file on my S3 with the alternate cube.glb.
In the properties dialogue I re-paste the link (identical).  If the model had turned into the green outline previously, I now see the new model as expected, but if it remained visible, nothing happens.

Again so far I've only got this working on the original Feb 9th version but it's worth trying the others again.  Will keep investigating.

daleglass:
Going by Aitolda's comments it seems this needs more testing still

daleglass (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-Interface-PR1012-a2125fb-_8e472fc4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-a2125fb-_8e472fc4.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-a2125fb-.tar_8e472fc4.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-a2125fb-.tar_8e472fc4.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-a2125fb-_8e472fc4.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1012/Vircadia-PR1012-a2125fb-_8e472fc4.exe

Issue #172, opened 2020-02-18
BUG: Zone entities with a “shapeType” other than “Box” cause a glitches


AleziaKurdis:
If we set the “shapeType” of a Zone entity to “cylinder” or “sphere”, when the user enters the cube of the zone but is still outside the cylinder or the sphere (let say he’s in the corner) then a glitch happens.

The Zone is momentarily applied to be immediately reverted. It behaves as if the evaluation of our position related to the sphere or cylinder was done after have been considered inside the zone.

Priority: Medium

HifiExperiments:
@AleziaKurdis is this still a problem?  I've tried to reproduce it a few times since this was posted and never could.  I'm using a zone with a skybox color and everything else set to inherit.  does it happen with some particular property configuration?  from the code, I don't see how this would be possible

AleziaKurdis:
1- Go there: http://metaverse.bashora.com/explore/serverlessWothal-UM.json
2- using create app, set radius to 10000 and find the Zone entity: "PLANET W ATMOSPHERE"
3- Rotate it to -135 deg on the X axis it to be in a corner outside the zone spheric shape

4- Now look the horizon, and click on the arrow of the X field to rotate it with a fraction of degree and observe the refresh.. you see momentarily the sky of that zone.
![image](https://user-images.githubusercontent.com/60075796/107463932-2dbd1980-6b2d-11eb-8ede-103bd235f560.png)

(Originally, there was a domain script doing that slow rotation, so the glitch was very visible. Sadly. I decommissioned this content, but you still can see the refresh issue rotating it manually.)
Sorry to can't provide you something better to reproduce it.

This is clearly not a high priority problem. But there is glitch.
There is certainly a workaround with the renderWithZones feature for this specific case (it wasn't existing when I created that)
So we can say that the priority just dropped from medium to low

PR #1033, opened 2021-02-12
Non-uniformly scaled sphere zones and models behave like ellipsoids (and fix EntityItem::contains sphere case)

HifiExperiments:
fix https://github.com/vircadia/vircadia/issues/172

which zones you are in is determined by `EntityTreeRenderer::findBestZoneAndMaybeContainingEntities`, which first calls `evalEntitiesInSphere` (which does a rough bounding box/sphere check) and then, under certain circumstances, calls `contains` on each of those entities to make sure we're actually inside them.  however, when a zone's visibility or transform is edited, we also immediately call `EntityTreeRenderer::LayeredZones::update`, which calls `contains` on the zone (not `evalEntitiesInSphere`).  `EntityItem::contains` was wrong for this case, so it would improperly return `true` when you were in the corner, but then next frame, when we call `evalEntitiesInSphere`, it would remove the zone before `contains` could be called.  this would result in a single frame pop of the spherical zone anytime you edited its transform

the fix is simple, but exposes another question: when you non-uniformly scale a "sphere" zone currently, we use its *max* dimension (both in EntityItem::contains and in entityShapeVisualizer.js for Create).  this means that if you make one dimension bigger, it actually extends outside of its bounding box in the other two dimensions, which seems wrong.  for zones specifically, it also means that this sphere clips against its bounding box (because `evalEntitiesInSphere` falls back to a bounding box check for non-uniformly scaled spheres).  is this current behavior for non-uniformly scaled sphere zones correct?  it would be easy to change these to `min`s in EntityItem::contains and entityShapeVisualizer.js

...but models actually have the same behavior when you give them sphere colliders and non-uniform dimensions.  see `ShapeInfo::setParams` and `ShapeFactory::createShapeFromInfo` should these be changed too?

UPDATE: "sphere" zones and models now behave like sphere shapes; they handle uniform dimensions and automatically switch to ellipsoids internally to handle non-uniform dimensions.

Test plan:
- Confirm that the repro steps in #172 no longer show a flash
- Confirm that "sphere" zones work properly, when they are scaled uniformly and when they are stretched
- Confirm that "sphere" collisions for models now scale with them properly, unlike in master

### DIFF
--- a/libraries/physics/src/ShapeFactory.cpp
+++ b/libraries/physics/src/ShapeFactory.cpp
@@ -278,12 +278,6 @@ const btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info)
             shape = new btBoxShape(glmToBullet(info.getHalfExtents()));
         }
         break;
-        case SHAPE_TYPE_SPHERE: {
-            glm::vec3 halfExtents = info.getHalfExtents();
-            float radius = glm::max(halfExtents.x, glm::max(halfExtents.y, halfExtents.z));
-            shape = new btSphereShape(radius);
-        }
-        break;
         case SHAPE_TYPE_MULTISPHERE: {
             std::vector<btVector3> positions;
             std::vector<float> radiuses;
@@ -297,6 +291,7 @@ const btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info)
             shape->setMargin(MULTI_SPHERE_MARGIN);
         }
         break;
+        case SHAPE_TYPE_SPHERE:
         case SHAPE_TYPE_ELLIPSOID: {
             glm::vec3 halfExtents = info.getHalfExtents();
             float radius = halfExtents.x;
--- a/libraries/shared/src/ShapeInfo.cpp
+++ b/libraries/shared/src/ShapeInfo.cpp
@@ -115,12 +115,6 @@ void ShapeInfo::setParams(ShapeType type, const glm::vec3& halfExtents, QString 
         case SHAPE_TYPE_HULL:
         case SHAPE_TYPE_MULTISPHERE:
             break;
-        case SHAPE_TYPE_SPHERE: {
-                float radius = glm::length(halfExtents) / SQUARE_ROOT_OF_3;
-                radius = glm::max(radius, MIN_HALF_EXTENT);
-                _halfExtents = glm::vec3(radius);
-            }
-            break;
         case SHAPE_TYPE_CIRCLE: {
             _halfExtents = glm::vec3(_halfExtents.x, MIN_HALF_EXTENT, _halfExtents.z);
         }
@@ -228,6 +222,7 @@ float ShapeInfo::computeVolume() const {
             volume = 8.0f * _halfExtents.x * _halfExtents.y * _halfExtents.z;
             break;
         }
+        case SHAPE_TYPE_ELLIPSOID:
         case SHAPE_TYPE_SPHERE: {
             volume = 4.0f * PI * _halfExtents.x * _halfExtents.y * _halfExtents.z / 3.0f;
             break;
--- a/scripts/system/create/entityProperties/html/js/entityProperties.js
+++ b/scripts/system/create/entityProperties/html/js/entityProperties.js
@@ -281,7 +281,7 @@ const GROUPS = [
             {
                 label: "Shape Type",
                 type: "dropdown",
-                options: { "box": "Box", "sphere": "Sphere", "ellipsoid": "Ellipsoid", 
+                options: { "box": "Box", "sphere": "Sphere",
                            "cylinder-y": "Cylinder", "compound": "Use Compound Shape URL" },
                 propertyID: "zoneShapeType",
                 propertyName: "shapeType", // actual entity property name
--- a/scripts/system/create/modules/entityShapeVisualizer.js
+++ b/scripts/system/create/modules/entityShapeVisualizer.js
@@ -13,6 +13,7 @@
 var SHAPETYPE_TO_SHAPE = {
     "box": "Cube",
     "ellipsoid": "Sphere",
+    "sphere": "Sphere",
     "cylinder-y": "Cylinder",
 };
 
@@ -34,14 +35,6 @@ function getEntityShapePropertiesForType(properties) {
                     type: "Model",
                     modelURL: properties.compoundShapeURL,
                     localDimensions: properties.localDimensions
-                };
-            } else if (properties.shapeType === "sphere") {
-                var sphereDiameter = Math.max(properties.localDimensions.x, properties.localDimensions.y,
-                    properties.localDimensions.z);
-                return {
-                    type: "Sphere",
-                    modelURL: properties.compoundShapeURL,
-                    localDimensions: {x: sphereDiameter, y: sphereDiameter, z: sphereDiameter}
                 };
             }
             break;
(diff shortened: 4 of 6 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761.tar_51bf54c7.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761_51bf54c7.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761.tar_51bf54c7.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761_51bf54c7.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-Quest-Alpha-PR1033-5653761_51bf54c7.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761_51bf54c7.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-5653761_51bf54c7.dmg

Aitolda:
I'm not seeing any clear behavior difference between RC3 and this PR in regards to sphere shape/collisions on meshes OR zones.  The scaling still seems to be [uniform.](https://i.gyazo.com/7e89a2e6d07a3f1d8a91a35b44fe2c7b.png)

HifiExperiments:
@Aitolda yeah I haven’t changed any behavior yet, for now this just fixes the flash issue that Alezia reported.  I am still working on making zones and model collisions become ellipsoids automatically when they have non-uniform dimensions

libraries/physics/src/ShapeFactory.cpp
```diff
@@ -278,12 +278,6 @@ const btCollisionShape* ShapeFactory::createShapeFromInfo(const ShapeInfo& info)
             shape = new btBoxShape(glmToBullet(info.getHalfExtents()));
         }
         break;
-        case SHAPE_TYPE_SPHERE: {
-            glm::vec3 halfExtents = info.getHalfExtents();
-            float radius = glm::max(halfExtents.x, glm::max(halfExtents.y, halfExtents.z));
-            shape = new btSphereShape(radius);
-        }
-        break;
```
HifiExperiments, line 286: this is where we actually compute the Bullet collision hull.  we want to use btSphereShape when possible for performance reasons, but the ellipsoid case below already handles this

libraries/shared/src/ShapeInfo.cpp
```diff
@@ -115,12 +115,6 @@ void ShapeInfo::setParams(ShapeType type, const glm::vec3& halfExtents, QString
         case SHAPE_TYPE_HULL:
         case SHAPE_TYPE_MULTISPHERE:
             break;
-        case SHAPE_TYPE_SPHERE: {
-                float radius = glm::length(halfExtents) / SQUARE_ROOT_OF_3;
-                radius = glm::max(radius, MIN_HALF_EXTENT);
-                _halfExtents = glm::vec3(radius);
```
HifiExperiments, line 121: the halfExtents are already set above, and we want to pass the actual halfExtents through so that the collision hull code can determine if we're actually a sphere or if we're an ellipsoid

scripts/system/create/entityProperties/html/js/entityProperties.js
```diff
@@ -281,7 +281,7 @@ const GROUPS = [
             {
                 label: "Shape Type",
                 type: "dropdown",
-                options: { "box": "Box", "sphere": "Sphere", "ellipsoid": "Ellipsoid",
```
HifiExperiments, line 284: this removes the "ellipsoid" option from Create because you can do the same exact thing with a non-uniformly scaled sphere, so we can simplify the options.  you could still set the shapeType to "ellipsoid" via another script and it would work the same way.

scripts/system/create/modules/entityShapeVisualizer.js
```diff
@@ -35,14 +36,6 @@ function getEntityShapePropertiesForType(properties) {
                     modelURL: properties.compoundShapeURL,
                     localDimensions: properties.localDimensions
                 };
-            } else if (properties.shapeType === "sphere") {
```
HifiExperiments, line 38: we can remove this because non-uniform sphere zones are no longer a special case

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8_f9b88c68.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8.tar_f9b88c68.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8_f9b88c68.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8.tar_f9b88c68.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8_f9b88c68.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-Quest-Alpha-PR1033-332dea8_f9b88c68.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1033/Vircadia-PR1033-332dea8_f9b88c68.apk

Aitolda:
Ellipsoid seems to be working.  Tested with [this mesh](https://aitolda.s3.amazonaws.com/Vircadia/Domains/Gorge/Oddtree.fbx) as well as a spherical zone.

AleziaKurdis:
What's the difference between ellipsoid and spherical ?
(I thoughtthat a sphere war an ellipsoid where the 3 dimensions were equal.)
Also, there are no changes to do for the zone visualizer?

HifiExperiments:
you are right that spheres are ellipsoids where the dimensions are all the same!  the problem was that you could still take a spherical zone/model and scale it non-uniformly, and it would confusingly still act like a sphere.  now, ellipsoid and sphere are effectively the same thing to the user, but internally, we handle them as effectively as possible.  this matches the behavior for shape entities: there is no ellipsoid type, but you can scale a sphere.

overall I think this behavior is MUCH simpler for end users, I have no idea why it was so complicated before

also, I did update the shape visualizer, you can see the changes here: https://github.com/vircadia/vircadia/pull/1033/files#diff-eb5b6680416bd90d271a085140085c454b47226c0133824224944087a52e4238

ctrlaltdavid (approved):

AleziaKurdis:
Tests:

1- Zone Shape Visualizer
Create a zone with a shape = "Sphere"
Make sure it behave as an Ellipsoid. (and no more always a sphere)
image

image

**This is working Fine**

2- Zone must be applied when we are inside the ellipsoid and no more based on an enforced sphere.
**This is also solved, no more glitch.**

3- Model with Collision Hull = "Sphere" must behave as ellipsoid and no more as a pure sphere.
image

image
image

**Work now as expected.**

4- Shape of type Sphere must behave as ellipsoid and no more as a pure sphere
image

image

**This was working before and still working now.**



**So, I think we are all good with this solution.**

PR #1040, opened 2021-02-14
Name a bunch of threads

HifiExperiments:
this will hopefully help with debugging.  in VS, when you hit a breakpoint or pause execution (or crash), in the Thread dropdown, more threads will have descriptive names now.  there's still a lot of unnamed ones...I think some are from the QThreadPool, and evidently you can't name them because there's no way to name a thread multiple times (when they get reused).  a bunch are seemingly from web processes...and there's a bunch that I'm not sure what they do...

Test plan:
- no real functional change, make sure server and client both still work

### DIFF
--- a/libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
@@ -304,7 +304,9 @@ void GLTextureTransferEngineDefault::processTransferQueues() {
 #if THREADED_TEXTURE_BUFFERING
     if (!_transferThread) {
         _transferThread = new TextureBufferThread(*this);
-        QObject::connect(_transferThread, &QThread::started, [] { setThreadName("TextureBufferThread"); });
+        QString name = "TextureBufferThread";
+        _transferThread->setObjectName(name);
+        QObject::connect(_transferThread, &QThread::started, [name] { setThreadName(name.toStdString()); });
         _transferThread->start();
     }
 #endif
--- a/libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
+++ b/libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
@@ -8,8 +8,10 @@
 
 #include "GLTexture.h"
 
+#include <QObject>
 #include <QtCore/QThread>
 #include <NumericalConstants.h>
+#include <ThreadHelpers.h>
 
 #include "GLBackend.h"
 
@@ -64,7 +66,7 @@ public:
 protected:
     class TextureBufferThread : public QThread {
     public:
-        TextureBufferThread(GLTextureTransferEngineDefault& parent) : _parent(parent) { start(); }
+        TextureBufferThread(GLTextureTransferEngineDefault& parent) : _parent(parent) {}
 
     protected:
         void run() override {
@@ -302,6 +304,8 @@ void GLTextureTransferEngineDefault::processTransferQueues() {
 #if THREADED_TEXTURE_BUFFERING
     if (!_transferThread) {
         _transferThread = new TextureBufferThread(*this);
+        QObject::connect(_transferThread, &QThread::started, [] { setThreadName("TextureBufferThread"); });
+        _transferThread->start();
     }
 #endif
 
--- a/.gitignore
+++ b/.gitignore
@@ -111,3 +111,4 @@ tools/unity-avatar-exporter
 server-console/package-lock.json
 vcpkg/
 /tools/nitpick/compiledResources
+qt/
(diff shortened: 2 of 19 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6_30c4779e.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6.tar_30c4779e.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6.tar_30c4779e.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6_30c4779e.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6_30c4779e.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-4bea4d6_30c4779e.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-Quest-Alpha-PR1040-4bea4d6_30c4779e.apk

libraries/gpu-gl-common/src/gpu/gl/GLTextureTransfer.cpp
```diff
@@ -302,6 +304,8 @@ void GLTextureTransferEngineDefault::processTransferQueues() {
 #if THREADED_TEXTURE_BUFFERING
     if (!_transferThread) {
         _transferThread = new TextureBufferThread(*this);
+        QObject::connect(_transferThread, &QThread::started, [] { setThreadName("TextureBufferThread"); });
```
daleglass, line 307: Maybe setObjectName() missing here?

daleglass (changes requested):

daleglass:
This looks nice and useful, a few notes though:

* What is the deadlock danger here? I want to make sure I'm not missing something.
* `setThreadName` is... interesting. I found [an explanation](https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2019) and still have doubts. It says "Only works if the debugger is attached at the time the exception-based method is used" -- what happens otherwise? Just gets ignored hopefully?
* Personally I'd rename it to something like `setWinThreadName` because it doesn't do anything on anything else, and that makes the code confusing to read as to why one needs naming things twice.
* Why not use [SetThreadDescription](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription?redirectedfrom=MSDN) instead?

HifiExperiments:
I added the label more for the threading aspect, just want to be wary as it touches a lot of areas

I tried to avoid messing with the implementation of setThreadName, since I agree it’s very weird.  yes, it’s ignored if the debugger isn’t attached at the time it happens.  I think we should leave it as setThreadName, and instead implement it on other systems in the future.  it accomplishes a different goal than setObjectName (which makes the thread object itself more identifiable in the debugger, but doesn’t help identify which thread is crashing).  we could certainly switch to setThreadDescription but I’m not sure it solved any problems that we really need to solve, and it isn’t supported on all systems

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b.tar_eefe8099.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b.tar_eefe8099.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b_eefe8099.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b_eefe8099.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b_eefe8099.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-PR1040-cba619b_eefe8099.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1040/Vircadia-Quest-Alpha-PR1040-cba619b_eefe8099.apk

daleglass (approved):

daleglass:
I've given it a test, the results look nice :-)

There's a bunch of threads that are named "interface" and seem to be coming from TBB. Perhaps there's still room for improvement there? But I think that can wait.

Issue #744, opened 2020-09-21
Gizmo entity not supporting Material or unlit


AleziaKurdis:
Pluging a Material entity on a "Gizmo" entity is doing nothing.

The reason of the existence of the Gizmo entity seems to be for UI.
Using as UI, entities have advantage to be able to be set as "unlit".

Can we have Material working with Gizmo to be able to set it as unlit,
or have an "unlit" property on it (as Text and Image)

The Gizmo ring can have a start and a end angle that is useful to display angles or to make gauges, this can't be achieve with an image or a model. (not without external system assistance)

AleziaKurdis:
Probably addressed by PR #1076
To be tested, and then closed if it works.

Issue #224, opened 2020-03-21
Some shape type get occulted when alpha is applied. 


AleziaKurdis:
Some specific shape types like "Sphere", "tetrahedron" and a few others
are occulted completely if they have transparency when they're behind a transparent cube. But it works for other shape types like "cube", "cylinder" and a few others.

![image](https://user-images.githubusercontent.com/60075796/77219170-7df04f00-6b09-11ea-8372-c9fa8b9c9ebd.png)

![image](https://user-images.githubusercontent.com/60075796/77219172-847ec680-6b09-11ea-8e9a-f3a3dd430037.png)

SilverfishVR:
I can confirm this is still an issue (running PR442), and is related to the kind of shape, there seem to be some weird kind of rendering order between them:

![shapes_2020-06-14 23 53 07](https://user-images.githubusercontent.com/53531160/84605333-ca46f300-ae9c-11ea-9a99-e75d4f33a907.png)
all shapes (40% alpha) in front of all shapes (98% alpha)

get it here to confirm: https://silverfish-freestuff.s3.eu-north-1.amazonaws.com/ShapeOclutionTest.json

PR #1076, opened 2021-03-04
Material improvements

HifiExperiments:
fixes the remaining issue with transparent shape entities mentioned here: https://github.com/vircadia/vircadia/issues/224#issuecomment-643830147
fixes https://github.com/vircadia/vircadia/issues/744
fixes https://github.com/vircadia/vircadia/issues/1085

support material entities on Gizmo entities, Image entities, and Text entities

caveats:
- Gizmos don't currently define UVs, which means any material textures or procedural shaders that use them won't work.  this is not hard to fix...but we have to define how we want those UVs to work.  do they act as a cutout, or do the UVs go along the 2D torus.  gizmos also don't currently respect material albedo/opacity, because they are controlled by multiple entity property parameters each
- Image entities still need a valid imageURL, even if you override their rendering completely via a material.  keepAspectRatio will still respect that original image
- Text entities only support materials on the background, not the text itself currently.  when this is later supported, it will be by targeting "mesh part" "1" instead of "0"

Test plan:
- follow the instruction in https://github.com/vircadia/vircadia/issues/224#issuecomment-643830147 to confirm that transparent shapes render correctly now
- run [this](https://gist.github.com/HifiExperiments/52a9ab66807756d3fbebf5820589c312/raw/5ef13e1110a3dbe82f71111eec87ef8f94b22637/VariousMaterialsAndEntitiesTest.js) script.  you will see 5 entities and 6 materials in front of you.
- press "u" to apply each material in turn to each entity.  they should behave as expected, notwithstanding the above caveats.  note that the 4th material will make the flat entities (text, image, gizmo) only render on one side because it defaults to a non-fallthrough cullFaceMode of CULL_BACK, so for example the gizmo will only be visible from underneath.
- verify that the create rotation circle things are now unlit (finally!)


libraries/entities-renderer/src/RenderableEntityItem.cpp
```diff
@@ -491,6 +495,176 @@ void EntityRenderer::removeMaterial(graphics::MaterialPointer material, const st
     emit requestRenderUpdate();
 }
 
+EntityRenderer::Pipeline EntityRenderer::getPipelineType(const graphics::MultiMaterial& materials) {
```
HifiExperiments, line 498: all of the changes in this file are moved from RenderableShapeEntityItem more or less without change, so that this all can be reused for other entities that support materials

libraries/entities-renderer/src/RenderableEntityItem.cpp
```diff
@@ -491,6 +495,176 @@ void EntityRenderer::removeMaterial(graphics::MaterialPointer material, const st
...
+    return Pipeline::SIMPLE;
+}
+
+bool EntityRenderer::needsRenderUpdateFromMaterials() const {
+    MaterialMap::const_iterator materials;
+    {
+        std::lock_guard<std::mutex> lock(_materialsLock);
+        materials = _materials.find("0");
+
+        if (materials == _materials.cend()) {
```
HifiExperiments, line 523: one thing to be really careful about here: if a method is const, we need to use a const_iterator, which means we need to compare against cend() NOT end().  I've learned the hard way that cend() is not always == end()

libraries/entities-renderer/src/RenderableShapeEntityItem.cpp
```diff
@@ -248,20 +132,29 @@ void ShapeEntityRenderer::doRender(RenderArgs* args) {
...
+            } else {
+                geometryCache->renderSolidShapeInstance(args, batch, geometryShape, outColor, pipeline);
+            }
         } else {
-            geometryCache->renderSolidShapeInstance(args, batch, geometryShape, outColor, pipeline);
+            if (wireframe) {
+                geometryCache->renderWireShape(batch, geometryShape, outColor);
+            } else {
+                geometryCache->renderShape(batch, geometryShape, outColor);
+            }
```
HifiExperiments, line 157: this is the fix for the transparent shape ordering issue.  before, we were using instanced rendering all the time for shapes.  but instanced rendering messes up the order that we draw things in (in exchange for faster rendering).  when shapes are transparent, we have to actually draw them one at a time.

libraries/entities-renderer/src/RenderableTextEntityItem.cpp
```diff
@@ -98,58 +68,126 @@ void TextEntityRenderer::doRenderUpdateAsynchronousTyped(const TypedEntityPointe
...
+        transparent |= procedural->isFading();
+        procedural->prepare(batch, transform.getTranslation(), transform.getScale(), transform.getRotation(), _created, ProceduralProgramKey(transparent));
+    } else if (pipelineType == Pipeline::MATERIAL) {
+        if (RenderPipelines::bindMaterials(materials, batch, args->_renderMode, args->_enableTexturing)) {
+            args->_details._materialSwitches++;
+        }
+    }
+
     auto geometryCache = DependencyManager::get<GeometryCache>();
-    // FIXME: we want to use instanced rendering here, but if textAlpha < 1 and backgroundAlpha < 1, the transparency sorting will be wrong
```
HifiExperiments, line 140: changed my mind!  we don't support a native "Quad" shape, it's just a squashed cube, which has the wrong tex coords for a quad + a bunch of unnecessary vertices.  the quad we use here is only 4 vertices so it won't make much of a difference

libraries/gpu-gl-common/src/gpu/gl/GLBackend.cpp
```diff
@@ -840,10 +840,7 @@ void GLBackend::do_glColor4f(const Batch& batch, size_t paramOffset) {
     if (_input._colorAttribute != newColor) {
         _input._colorAttribute = newColor;
         glVertexAttrib4fv(gpu::Stream::COLOR, &_input._colorAttribute.r);
-        // Color has been changed and is not white. To prevent colors from bleeding
-        // between different objects, we need to set the _hadColorAttribute flag
-        // as if a previous render call had potential colors
-        _input._hadColorAttribute = (newColor != glm::vec4(1.0f, 1.0f, 1.0f, 1.0f));
+        _input._hasColorAttribute = true;
```
HifiExperiments, line 843: `renderShape` calls the old hacky batch._glColor4f, which revealed it had a bug whereby the first object would render white.  I tried fixing it up

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb-.tar_1abb408a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb-_1abb408a.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb-.tar_1abb408a.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb-_1abb408a.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb-_1abb408a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-Quest-Alpha-PR1076-f2517fb_1abb408a.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-f2517fb_1abb408a.apk

Aitolda:
Are objects supposed to be disappearing when I get within  meter or two of them?

SilverfishVR:
🤔 they do seem to disagree on who is in-front

https://user-images.githubusercontent.com/53531160/110321050-732c1400-8011-11eb-9e56-f1a2c681547e.mp4

HifiExperiments:
@Aitolda @SilverfishVR based on those videos, here are my guesses as to what's happening:

1. in the first video, with the sphere with the shader applied, did you move the wavy material entity at all?  it could also be a bug with my script.  but the way procedural vertex shaders work is that the material entity is used for the culling (because there's no way for the entity to know where the vertex shader will move the vertices).  it looks like the material is still over by the gizmo maybe, so when that moves off the screen, the sphere stops rendering.  if that's really what's happening, that is expected.
2. in both videos, with the shape entities: I believe this is just our normal transparency sorting limitation at work.  since the backdrop red shape is also transparent, the system does its best to sort them, but it can often fail when the bounding boxes are close or overlapping.  the same is true of transparent models: since the drawing order matters for transparency, but we can't do per-pixel sorting, it's all or nothing.  I'm afraid there's not much we can do to alleviate this, it's a common limitation of game engines, and although there are advanced techniques that help, nothing is currently planned.

Aitolda:
I guess I'm not fully understand what I'm looking for here.  I see some unlit happening, but I'm a bit confused.

HifiExperiments:
haha sorry, the materials, in order, are: normal green, transparent green, unlit green, a normal material with a texture, a procedural fragment shader, and a procedural vertex + fragment shader.  pressing u will cycle through these, applying them all first to the sphere, then the model, etc.

Aitolda:
The procedural fragment shader, and a procedural vertex don't work on the gnome unless you firt go into the developer menu and enable procedural materials (but it will work on some of the other items).  Only the vertex shader seems to work properly on the "ring" which for some reason doesn't show up in the entity list.

HifiExperiments:
yeah, so, like I said, there are some caveats, in particular for gizmos (which don’t show up in the entity list, which we should post an issue for, because it was just never implemented): they don’t respect the albedo/opacity, so the first three don’t totally work (although the third one SHOULD make it unlit), and they don’t have UVs, so the fourth one won’t really work (and it’ll only render on the bottom, because it’s using CULL_BACK)

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1-_9a247172.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1-.tar_9a247172.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1-.tar_9a247172.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1-_9a247172.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1-_9a247172.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-PR1076-b25b7f1_9a247172.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1076/Vircadia-Quest-Alpha-PR1076-b25b7f1_9a247172.apk

Aitolda:
> yeah, so, like I said, there are some caveats, in particular for gizmos (which don’t show up in the entity list

Ah... so the ring is a gizmo.  I hadn't realized.  My bad.

Aitolda:
On that note, with the latest build (as of this morning) things are working as expected.

daleglass:
As best as I can tell, it looks good :)

PR #1098, opened 2021-03-17
Add avatarEntities permissions

digisomni:
### DONE

Server Settings > Permissions tables
- Changed permission name from "rez attachments" to "avatar entities".
- Fix column widths (last column was very narrow).
- Add help on the new permission (pop-up when hover the "?" at the top of each table).

Avatar Mixer
- Don't accept avatar entity data from a user if they don't have rez avatar entity permission.
- Clear user's avatar entities if their rez avatar entities permissions changes to "no".

Interface
- Warn user when connecting to a domain when they have avatar entities but don't have rez avatar entities permission, giving them the option of connecting to the domain or not.
- Don't rez own avatar entities if don't have rez avatar entities permission.
- Don't send own avatar entities to avatar mixer if don't have rez avatar entities permission.
- Stop displaying avatar entities to self and others if rez avatar entities permission changes to "no".
- Start displaying avatar entities to self and others if rez avatar entities permission changes to "yes". (5s delay.)
- Avatar app: Disable wearables lock and dialog buttons if on a domain where user doesn't have rez avatar entity permission. Button click displays a notification dialog.
- Disable adding avatar entities if on a domain where user doesn't have rez avatar entity permission. 
- Disable all `MyAvatar` methods dealing with avatar entities / attachments.
- Remove deprecated `MyAvatar.storeAvatarEntityDataPayload()` from API.

Domain Server
- Scripted avatars (`Agent` and `Avatar` APIs) have `canRezAvatarEntities` permission.
- Avatar recordings (the "Record" app) play back including any wearables.

### WON'T BE ADDRESSED

- 2021.1 Interface doesn't warn user if they're connecting to a domain where their avatar entities won't display.
- 2021.1 Interface still displays user's own avatar entities when user doesn't have rez avatar entities permission. Other users don't see these avatar entities, though.
- 2021.1 Interface user needs to relog after being granted rez avatar entities permission in order for other users to see that user's avatar entities.
- 2021.1 Interface doesn't disable Avatar app's wearables buttons.
- Avatar entity-related API calls are not disabled.

#### PR Server and PR Interface

Check the default domain server permissions: rez avatar entities permission should be enabled when connect permission is enabled.
- New domain server.
- Upgraded domain server.

When you have avatar entities and you start up in a domain where you don't have rez avatar entities permissions, you are prompted with a "continue | yes/no" dialog box before you connect to the domain. If you answer "no" you aren't connected to the domain.
Same when teleporting to the domain from another domain.
While on the domain, if the domain restarts, or your Internet connection momentarily drops out, you are not prompted again. But you are prompted again if you teleport away and back again.

Check the above in the following situations:
- Domain allows anonymous login.
- Domain requires metaverse login and you're already logged in.
- Domain requires metaverse login and you're not logged in.

If you don't have avatar entities, you are not prompted "continue | yes/no" when connecting to a domain where you don't have rez avatar entities permissions.

Using two separate PCs and separate user accounts, two avatars each with avatar entities, one user with rez avatar entities permissions and the other without...
- The user with rez avatar entities permissions sees their avatar entities but does not see the user without's avatar entities.
- The user without rez avatar entities permissions doesn't see their avatar entities but does see the user with's avatar entities.
- Changing the user with permissions to not have permissions, causes their avatar entities to no longer be visible to themself or by the other user.
- Changing the user to have permissions again, causes their avatar entities to become visible to themself and to the other user.

Check that, if you don't have rez avatar entities permissions, the two wearables buttons (lock and open dialog) are disabled; a notification message is displayed.

Check that, if you don't have rez avatar entities permissions, a new wearable added via API calls doesn't display - neither for yourself nor for observers. `Entities.addEntity()`, `MyAvatar::setAvatarEntityData()`, `MyAvatar::attach()`, `MyAvatar::setAttachmentData()`.

Check that avatar entities display on serverless domains (without any "continue | yes/no" prompt).

When using this PR's domain server on an existing domain (i.e., upgrading an existing domain), check the server settings has default Rez Avatar Entities permission values the same as Connect permissions, for each permissions entry.

Check that avatar with wearable recorded by the "Record" app plays back including the wearable, even if no-one has rez avatar entities permission the domain.

Check that the "continue without wearables" dialog redisplays if you previously said "no" but then re-enter the domain in the Explore app.

Check that the "continue without wearables" dialog does not redisplay if you previously said "yes" and are teleporting about on the domain:
- Using the Explore app.
- Using bookmarks.

#### PR Server and 2021.1 Interface

If user doesn't have canRezAvatarEntities permissions:
- User sees their own avatar entities but other users on PR Interface don't.
- User sees their own avatar entities but other users on 2021.1 Interface don't.

If user's canRezAvatarEntities permission changes from "can't" to "can" then they need to relog in order for other users to see user's avatar entities.

If user's canRezAvatarEntities permission changes from "can" to "can't" then other users see their avatar entities disappear. 

#### 2021.1 Server and PR Interface

User always sees their own avatar entities and those of others.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-f811676-_87def7a1.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-f811676-_87def7a1.exe

domain-server/resources/describe-settings.json
```diff
@@ -295,10 +295,10 @@
         },
         {
           "name": "approved_safe_urls",
-          "label": "Approved Script and QML URLs",
+          "label": "Approved Script and QML URLs (Not Enabled)",
           "help": "These URLs will be sent to the Interface as safe URLs to allow through the whitelist if the Interface has this security option enabled.",
-          "placeholder": "0",
-          "default": "1",
+          "placeholder": "",
+          "default": "",
```
HifiExperiments, line 301: this and the above seems separate?

domain-server/resources/describe-settings.json
```diff
@@ -338,7 +338,7 @@
...
...><li><strong>Rez Temporary Certified</strong><br />Sets whether a user can create new certified entities with a finite lifetime.</li><li><strong>Write Assets</strong><br />Sets whether a user can make changes to the domain&rsquo;s asset-server assets.</li><li><strong>Ignore Max Capacity</strong><br />Sets whether a user can connect even if the domain has reached or exceeded its maximum allowed agents.</li><li><strong>Rez Certified</strong><br />Sets whether a user can create new certified entities.</li><li><strong>Get and Set Private User Data</strong><br>Sets whether a user can get and set the privateUserData entity property.</li></ul><p>Note that permissions assigned to a specific user will supersede any parameter-level permissions that might otherwise apply to that user. Additionally, if more than one parameter is applicable to a given user, the permissions given to that user will be the sum of all applicable parameters. For example, let&rsquo;s say only localhost users can connect and only logged in users can lock and unlock entities. If a user is both logged in and on localhost then they will be able to both connect and lock/unlock entities.</p>'>domain-wide permissions</a>.",
```
HifiExperiments, line 341: hard to tell from the code but is this order desirable?  seems like avatar entity rez should be closer to the other rez perms

interface/src/Application.cpp
```diff
@@ -9172,6 +9175,32 @@ void Application::setShowBulletConstraintLimits(bool value) {
...
+        return;
+    }
+
+    QString continueMessage = "Your wearables will not display on this domain. Continue?";
+    _confirmConnectWithoutAvatarEntitiesDialog = OffscreenUi::asyncQuestion("Continue Without Wearables", continueMessage,
+        QMessageBox::Yes | QMessageBox::No);
+    if (_confirmConnectWithoutAvatarEntitiesDialog->getDialogItem()) {
+        QObject::connect(_confirmConnectWithoutAvatarEntitiesDialog, &ModalDialogListener::response, this, [=](QVariant answer) {
+            QObject::disconnect(_confirmConnectWithoutAvatarEntitiesDialog, &ModalDialogListener::response, this, nullptr);
+            _confirmConnectWithoutAvatarEntitiesDialog = nullptr;
```
HifiExperiments, line 9197: is this a memory leak?  do we need to `delete` the dialog?

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1533,7 +1536,19 @@ void MyAvatar::storeAvatarEntityDataPayload(const QUuid& entityID, const QByteAr
 
 void MyAvatar::clearAvatarEntity(const QUuid& entityID, bool requiresRemovalFromTree) {
     // NOTE: the requiresRemovalFromTree argument is unused
+
+    if (!DependencyManager::get<NodeList>()->getThisNodeCanRezAvatarEntities()) {
+        qCDebug(interfaceapp) << "Ignoring clearAvatarEntity() because don't have canRezAvatarEntities permission on domain";
```
HifiExperiments, line 1541: what if someone ACTUALLY tries to delete an avatar entity in this case?  or if its lifetime expires or a script tries to kill it?

libraries/networking/src/DomainHandler.cpp
```diff
@@ -364,17 +366,39 @@ void DomainHandler::setIsConnected(bool isConnected) {
             _lastDomainConnectionError = -1;
             emit connectedToDomain(_domainURL);
 
+            // FIXME: Reinstate the requestDomainSettings() call here in version 2021.2.0 instead of having it in
```
HifiExperiments, line 369: maybe an issue for this?

libraries/networking/src/DomainHandler.h
```diff
@@ -287,6 +292,8 @@ private slots:
     HifiSockAddr _iceServerSockAddr;
     NetworkPeer _icePeer;
     bool _isConnected { false };
+    bool _haveAskedConnectWithoutAvatarEntities { false };
+    bool _canConnectWithoutAvatarEntities { false };
```
HifiExperiments, line 296: do these get reset when you switch domains?  should they?

HifiExperiments:
this is cool and also amazing test plan

digisomni:
It's all David's! :D

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242-_177196c9.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242-_177196c9.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242-.tar_177196c9.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242-.tar_177196c9.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242-_177196c9.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-9047242_177196c9.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-9047242_177196c9.apk

domain-server/resources/describe-settings.json
ctrlaltdavid: This is a miscellaneous fix to the server settings that Kalila included at the start of the branch. : )

interface/src/Application.cpp
ctrlaltdavid: The `OffscreenUi::asyncQuestion()` dialog is automatically destroyed when `_confirmConnectWithoutAvatarEntitiesDialog` is set to null. (Confirmed by instrumenting code.)

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: The avatar entities aren't in the tree so can't be operated on in the tree.

libraries/networking/src/DomainHandler.cpp
ctrlaltdavid: https://github.com/vircadia/vircadia/issues/1099

libraries/networking/src/DomainHandler.h
ctrlaltdavid: Yes, they get reset in `DomainHandler::hardReset()`.

domain-server/resources/describe-settings.json
ctrlaltdavid: Good point. The order is under discussion.

ctrlaltdavid: The order is desirable: the ability to use avatar entities is more closely associated with the ability to connect. Typically you'll have both "Connect" and "Avatar Entities" ticked but perhaps none of the others.
Resolution: Rename the server settings columns from "Rez Avatar Entities" to "Avatar Entities".

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20-_26aa6354.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20-.tar_26aa6354.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20-.tar_26aa6354.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20-_26aa6354.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-4c13f20_26aa6354.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20_26aa6354.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-4c13f20-_26aa6354.exe

interface/src/avatar/MyAvatar.cpp
HifiExperiments: by “tree” do you mean entity tree?  avatar entities are in the entity tree for clients, just not on the avatar mixer.

(I think)

ctrlaltdavid: I mean the client entity tree. When avatar entities cannot be displayed, they're removed from the client entity tree in `MyAvatar::handleCanRezAvatarEntitiesChanged()`. Because they're not in the tree the cannot be deleted.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55-.tar_eac2758c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55-_eac2758c.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55-.tar_eac2758c.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55-_eac2758c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55-_eac2758c.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-56e8d55_eac2758c.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-56e8d55_eac2758c.apk

interface/src/avatar/MyAvatar.cpp
```diff
@@ -1564,6 +1583,33 @@ void MyAvatar::sanitizeAvatarEntityProperties(EntityItemProperties& properties)
...
+bool MyAvatar::hasAvatarEntities() const {
+    return _cachedAvatarEntityBlobs.count() > 0;
+}
+
+void MyAvatar::handleCanRezAvatarEntitiesChanged(bool canRezAvatarEntities) {
+    if (canRezAvatarEntities) {
+        // Start displaying avatar entities.
+        // Allow time for the avatar mixer to be updated with the user's permissions so that it doesn't discard the avatar 
+        // entities sent.
+        QTimer::singleShot(2 * DOMAIN_SERVER_CHECK_IN_MSECS, this, [this] {
```
HifiExperiments, line 1604: couldn’t it take longer?  this seems hacky

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135-.tar_70ebdf70.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135-_70ebdf70.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135-.tar_70ebdf70.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135-_70ebdf70.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135-_70ebdf70.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-1af2135_70ebdf70.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-1af2135_70ebdf70.apk

interface/src/avatar/MyAvatar.cpp
ctrlaltdavid: Here's what happens:
- Domain owner enables canRezAvatarEntities permission for a user.
- User's Interface sends a `DomainListRequest` protocol message to the domain server every `DOMAIN_SERVER_CHECK_IN_MSECS` as a "domain server check-in".
- Avatar mixer assignment client sends a `DomainListRequest` protocol message to the domain server every `DOMAIN_SERVER_CHECK_IN_MSECS` as a "domain server check-in".
- In response to each `DomainListRequest`, the domain server sends a `DomainList` protocol message. This message contains the latest permissions settings.
- User's Interface processes `DomainList` protocol message and sees that avatar entities are now enabled for user, so adds user's avatar entities to tree which causes them to be sent to the avatar mixer.
- Avatar mixer processes `DomainList` protocol message and updates its copy of the domain permissions including the updated permission for the user.
- Avatar mixer receives avatar entities from user. If the avatar entities arrive before the `DomainList` protocol message with the user's udated permissions, the avatar entities are discarded and so not displayed to other users.

In the domain server the processing `DomainListRequest` messages there could be up to `DOMAIN_SERVER_CHECK_IN_MSECS` delay between the domain server sending updated permissions to the user and sending updated permissions to the avatar mixer, assuming that incoming `DomainListRequest` messages are being received regularly. 
In practice, the domain server and avatar mixer will have a fast connection between them (same machine or local network) so they shudln't be subject to networking delays.
There may be longer delays and variability in the connection between the user's Interface and the domain server. However, the worst case scenario is that there is no delay (e.g., Interface is on the same machine as the domain server) and Interface's `DomainListRequest` request comes in at `DOMAIN_SERVER_CHECK_IN_MSECS` before the avatar mixer's `DomainListRequest` request, so Inteface potentially knows that it has canRezAvatarEntities permissions `DOMAIN_SERVER_CHECK_IN_MSECS` before the avatar mixer does.

So, we could in theory get away with having just a `1 * DOMAIN_SERVER_CHECK_IN_MSECS` delay before sending the avatar entities to the avatar mixer, but a longer delay doesn't hurt.

HifiExperiments: but none of those times are really guaranteed, right?  a slow client could be sending or processing check ins slower, and a slow mixer could be responding slowly, especially if there are lots of clients.  is there a way we can detect “we got the next check in, we’re good to go now” and move this code to the avatar update or something like that?

ctrlaltdavid: If the Interface client is slow: This doesn't matter: Interface can send its avatar entities late; they'll just display late for observers.

If the avatar mixer is slow: It it's bogged down it's still likely that incoming avatar entities from Interface will be processed after the incoming `DomainList` message from the domain server that enabled avatar entities for the user. If the avatar mixer is bogged down so much that it can't send `DomainListRequest` messages to the domain server in a timely manner nor process the `DomainList` response in a timely manner, then the user's avatar entities won't display for observers unless the user teleports away and back or relogs.

In order to address the avatar-mixer-bogged-down-bigly scenario, Interface needs to hold off sending its avatar entities until it knows that the avatar mixer has received a `DomainList` message sent after the `DomainList` that interface received granting it avatar entities permissions. This could be achieved using the timestamp already included in the `DomainList` message, and making the avatar mixer send the timestamp value to Interface. I haven't found any suitable protocol messages to add the timestamp value to (protocol change); it would probably be better to add a new message (protocol change).

However, checking the timestamps in this manner seems like overkill for this scenario? Perhaps a simpler but more conservative solution could be to base Interface's send delay on, say, the greater of `5 * DOMAIN_SERVER_CHECK_IN_MSECS` and `3 * avatarPing` (3 * the avatar mixer ping time)? 
Note: Should also cancel the send delay timer if the user's permissions revert to "can't rez avatar entities" during the delay.

HifiExperiments: ok, I'm convinced, I think maybe just a comment summarizing some of this would be sufficient and we can revisit in the future if bugs crop up, but the value probably won't be changing often anyways

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a-_308b2e08.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a-.tar_308b2e08.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a-.tar_308b2e08.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a-_308b2e08.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a-_308b2e08.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-e21cb4a_308b2e08.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-e21cb4a_308b2e08.apk

libraries/entities/src/EntityScriptingInterface.cpp
```diff
@@ -481,6 +487,15 @@ QUuid EntityScriptingInterface::addEntityInternal(const EntityItemProperties& pr
 
     _activityTracking.addedEntityCount++;
 
+    auto nodeList = DependencyManager::get<NodeList>();
+
+    if (entityHostType == entity::HostType::AVATAR && !nodeList->getThisNodeCanRezAvatarEntities()) {
```
HifiExperiments, line 492: does this codepath get called for someone without permission if someone WITH permissions tries to add an avatar entity?

HifiExperiments (approved):
looks good, one last question

libraries/entities/src/EntityScriptingInterface.cpp
ctrlaltdavid: No, this method is only called by scripting API methods: it's only used by a script to create a new entity; it isn't used to add other user's entities to the local view.

As a check, I ran an "add-avatar-entity" script:
(a) On a user with avatar entities permissions. The entity was successfully added. And it was visible to a user without avatar entities permissions.
(b) On a user withotu avatar entities permissions. The "Ignoring addEntity() because don't have canRezAvatarEntities permission on domain" message was logged and the entity wasn't created.

digisomni:
A bug: When joining a domain that does not allow wearables for myself, I can select "No" to connecting... that'll leave me in a transient state of connecting that I cannot break out of without going elsewhere and returning.

If I try to connect again without navigating elsewhere, it won't work. Attempting another domain navigation should break me out of this I'd reckon.

ctrlaltdavid:
@digisomni Selecting "no" to the continue-or-not question stops you from connecting to the domain, i.e., you're in a "not connected" state (similar to if the domain server is down) which is expected behaviour.
If you navigate away elsewhere and then come back to the domain you'll be asked the question again.
How do you try to connect again without navigating elsewhere?

digisomni:
> 
> 
> @digisomni Selecting "no" to the continue-or-not question stops you from connecting to the domain, i.e., you're in a "not connected" state (similar to if the domain server is down) which is expected behaviour.
> If you navigate away elsewhere and then come back to the domain you'll be asked the question again.
> How do you try to connect again without navigating elsewhere?

I simply press enter for the server IP (in this case it's "localhost") again in the Explore app. I figure that the "Not Connected" state is being enforced Interface side. A not connected state with retrying is normal for trying to get back into a server but I would say that trying to navigate there again (using the API for example) should trigger the dialog again, maybe.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976-.tar_bf423177.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976-.tar_bf423177.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976-_bf423177.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976-_bf423177.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976-_bf423177.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-8801976_bf423177.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-8801976_bf423177.apk

HifiExperiments (approved):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6-_bf423177.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6-_bf423177.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6-.tar_bf423177.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6-.tar_bf423177.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-Quest-Alpha-PR1098-ea7c8a6_bf423177.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6_bf423177.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1098/Vircadia-PR1098-ea7c8a6-_bf423177.exe

Issue #114, opened 2020-01-23
No ability for multiple entities to reference the same material entity


Aitolda:
At the moment only a single entity can be influenced by a material entity, requiring a separate material entity from every mesh that want's to use this material.  If we were able to allow multiple entities dramatically simplifies workflow for creators.  Right now the connection is made by turning a material entity into the child of the mesh it influences.  Perhaps there is another way to connect these allowing for flexibility.

HifiExperiments:
the trick with this is that we made material entities follow their parent so you wouldn’t end up with random material entities all over your domain, and so that we only load material entities as needed (when you see them or their parent)

perhaps a solution here is a material library?  each material entity still only affects one entity, but you have some way of defining a material once and then reusing it, and having changes to that one material affect all instances...

Aitolda:
"perhaps a solution here is a material library? each material entity still only affects one entity, but you have some way of defining a material once and then reusing it, and having changes to that one material affect all instances"

This sounds like a reasonable alternative to me, of course, but I must confess my ignorance of the level of work needed.

SilverfishVR:
Tl:DR: I advocate careful consideration over speedy implementation.

It is important to distinguish between material entities, and material resources.
Material resources are extremely useful because they let you control and fine-tune each material parameter, in-app, in-location, real-time, circumventing all of the import-export issues, the problem is, that the only way to apply a material from a resource is through a material entity, this is very impractical due to the parenting system where a material can only be applied (parented) to 1 UUID and the fact that you can’t, by selecting model entity, easily figure out which mesh part is overridden by what material entity.

The goal, I think should be a system that allows me to apply a material to as many entities as I want, and allows me to apply as many materials to an entity as I want(within reason). While maintaining the possibility to “live edit” the material data and to parent entities (for projected materials), all without making the UI an incomprehensible mess.

So far, I think my non-breaking-current-implementation idea is:

Model and shape entities gains the properties currently reserved for material https://apidocs.projectathena.dev/Entities.html#.EntityProperties-Material
With one difference. Model entities needs to support multiple materials so maybe a new type of object? I dunno, but basically:

Model and shape entities can also hold multiple instances of material-entity properties.

This way you can have it all in the entity itself, or reference an external material resource if you want to use it across multiple entities (it could even be one resource with all of your materials).

UI wise, I imagine a “material override” subpage to the _create->Properties_ that  page shows a list of “overrides” each mostly identical to what you currently see in a material entity:
[...]

PR #576, opened 2020-07-28
Improved TAA with Velcocity Buffer

HifiExperiments:
Based on Zvork's old branch: https://github.com/yozlet/interface/compare/master...Zvork:velgbuf

This PR is about half the size of the original PR because a lot of the changes already made it into master in other forms (flushed out forward rendering for all shader paths, consolidation of shader variants, removal of 3D Overlays, etc.).  The underlying logic is unchanged (improved jittering timing and use of a deferred velocity buffer), but I tried to fix up some incorrect things in shaders and elsewhere.

The overarching goal of this PR is to improve our TAA implementation so that when deferred rendering is used things are not as blurry but edges are still anti-aliased.  There should be limited ghosting, and hopefully in VR, the rendering experience is much improved.

This PR touches almost every area of rendering.  To test, we need to verify that the following are working:
- all entity types
- renderLayer front and renderLayer HUD entities
- forward and deferred modes
- mirrors
- in deferred mode, all rendering effects (particularly bloom, and obviously AA)
- rendering on different OSes
- rendering in VR

PR #1102, opened 2021-03-18
Material improvements 2: reference textures/materials by entity ID

HifiExperiments:
builds on https://github.com/vircadia/vircadia/pull/1076, diff here: https://github.com/HifiExperiments/project-athena/compare/materials...materials2

implements https://github.com/vircadia/vircadia/issues/114 (but maybe needs more discussion?)

this is either a cool feature or incredibly stupid, I can't tell which

basically, allows you to use an entity ID as the URL to a texture or material.  **textures can currently only refer to image or web entities, and materials can currently only refer to material entities (or an entity with a material entity applied)**

for example: for textures, you can make a material entity with the following JSON:
```
{
    materials: {
        "albedoMap": <ID of web or image entity>
    }
}
```
and the material entity will copy that texture.  this lets you apply web surfaces to other entities/avatars!  it also lets you have multiple web views without the cost of multiple web entities.  however, only the original web entity will accept mouse events.

and for materials, you can either just use a material entity ID as another entity's materialURL, or you can put the ID in the materials field, like:
```
{
    materials: <ID of material entity>
}
```
and you can still have multiple materials:
```
{
    materials: [
        <ID of material entity>,
        {
            name: "blah",
            albedo: [1, 0, 0]
        }
    ]
}
```
(in which case, you can still refer to each of them in the materialURL by name, and for entity IDs, the name is just the ID)

the only downside is that currently, the entity that you're referencing must also be loaded.  this means you can't make a material to another material that is halfway across the world, it needs to be pretty close so that the entity server sends it etc.  I am still thinking about how to fix this

also fixes selection by name not working when the URL is materialData

Test plan:
- Create a web entity, an image entity, and a material entity with materialURL = materialData
- Set the materialData to:
```
{
    materials: {
        "albedoMap": <ID of the web entity>
    }
}
```
- Confirm that as the web surface changes, the material updates to match it.  Apply the material to a cube or model and confirm that the web surface is applied to that entity.
- In the material entity, change the ID to the image entity.  Confirm that the material entity now uses that texture.  Change the imageURL to a different image and confirm that it updates the material.
- Confirm that procedurals can accept entity IDs as input channels.  For example, you can make a material entity with materialData:
```
{
  "materials": {
    "model": "hifi_shader_simple",
    "procedural": {
      "version": 2,
      "fragmentShaderURL": "https://gist.github.com/HifiExperiments/51edf2e107cbeab18474416b751c1e07/raw/1817d7a2f5bf73978b1a759ecac93e936fc75d1d/ProceduralTexture.fs",
      "channels": [<id of web entity>]
    }
  }
}
```
- Create a material entity with materialURL "https://hifi-content.s3.amazonaws.com/samuel/TestMaterial.json".  This should be a normal material with a texture applied.  Call this materialA;
- Create a material entity with materialData:
```
{
  "materialVersion": 1,
  "materials": [
    {
      "name": "red",
      "model": "hifi_pbr",
      "albedo": [1, 0, 0]
    },
    {
      "name": "green",
      "model": "hifi_pbr",
      "albedo": [0, 1, 0]
    },
    {
      "name": "blue",
      "model": "hifi_pbr",
      "albedo": [0, 0, 1]
    },
    "<the ID of materialA>"
  ]
}
```
- Confirm that you can select either red, blue, or green by appending "?red", "?blue", or "?green" to the materialURL (this is broken sometimes in master right now)
- Confirm that you can select the ID by name by changing the materialURL to `materialData?<the ID of materialA>`
- Confirm that you can control the texture transforms of materialA and the last material separately.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5-.tar_15ba6c86.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5-.tar_15ba6c86.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5-_15ba6c86.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5-_15ba6c86.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5_15ba6c86.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-e1b7fe5_15ba6c86.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-e1b7fe5-_15ba6c86.exe

SilverfishVR:
just tried it shortly, and web entities as texture seems to be flipped on the Y axis, so it needs an offset of 1 (Y) to show in the correct place and the image is flipped on Y so text is ƨbɿɒwʞɔɒᗺ 😮 
still need to wrap my head around how this will work but rendering web pages to any surface is a really cool idea. 👍

HifiExperiments:
@SilverfishVR huh!  you're totally right...I have no idea why that would be, will look into it!

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726-.tar_a87d6e51.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726-.tar_a87d6e51.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726-_a87d6e51.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726-_a87d6e51.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726_a87d6e51.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-078e726_a87d6e51.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-078e726-_a87d6e51.exe

HifiExperiments:
ok I think Qt's coordinate system is just flipped...so...all of our UI and web surfaces actually render upside down, and then we flip it manually by flipping the y tex coords on web entities and the normal UI.  it's not really possible to fix this, so either we would have to patch Qt to render everything upside down, or we'd have to just live with upside down web textures.  thoughts?

digisomni:
If those are our two options, then upside down web textures is the ideal path haha. We'll just have to take note that you have to flip those textures "upside down" (from the creator perspective) by default.

SilverfishVR:
hmm. upside down is not really what I'm seeing, though having the UV offset and scale as XY is a bit confusing but since you can't set the scaling -1 (in the properties of the material entity) the only ways I can make it look right is:
 * flip UV of the mesh I want to use for web projection in blender.
 * use 3d projection mode, one of the projection directions will be flipped, so look right with web texture.

I think if it can't be made to work the same as regular textures maybe the webentity-as-texture should not be included, it is sure raise confusion and questions 🤷‍♂️ 

So far the other parts seem to work but I have not fully tested multimaterial entities and procedurals yet

HifiExperiments:
with the latest commit you can set materialMappingScale negative.  if you set the y scale to -1 and the y offset to 1, you can flip a web texture and see it normally.  web textures don't repeat, so only scales -1 to 1 will really work, and they also are in a different color space, which is annoying...

I'm open to other suggestions, and can disable it if we want, but if we do, I don't expect that anything will change in the future that would make a difference here, it's an unfortunate quirk of our usage of Qt.  maybe if we switched to a different web renderer, we'd have more control over the texture?  but QML rendering will always depend on Qt and as far as I can tell, will always be upside down

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b-.tar_7ff796af.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b-_7ff796af.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b-.tar_7ff796af.gz

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b-_7ff796af.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b-_7ff796af.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-0d8b50b_7ff796af.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-0d8b50b_7ff796af.apk

SilverfishVR:
* The addition of negative mapping scale does make the web textures usable so while it does constitute a bit of a mental geometry exercise I say we keep it 👍 

* Web entity as input to a procedural shader does seem to work too but you don't get the benefit of the scaling/offset so that would probably need to to be done it the shader itself (it could maybe even fix the color space too) 👍 
I tried playing youtube 3d videos though a panosphere shader, it did not work well :-)


* Accessing multiple materials by name or ID works, it did not fail, even going several layers of material inception deep. 👍 
note: you specified using `materialData?<materialName or ID>` but the docs says to use `materialData#<materialName or ID>`
both works but is one "more" correct than the other?

* The not so great: If you do something silly like make a material entity that reference it's own ID as a material, or any circular reference I think, the interface crashes, of course it it not expected to work but mistakes happen and whenever the interface tries to load an entity like this, it crashes so you have to do a rollback of the domain to fix it, maybe this could be handled more gracefully 😄 

note: the "Developer->Render->Enable Procedural materials" seems to only affect Avatars, I thought it would also apply to models (but not shapes)?

HifiExperiments:
yaaaay!  ok so I definitely agree we need to document the gymnastics necessary.  I will try to find a good place for this

ah yeah # is more correct than ?, ? is deprecated because it could also be used in a real URL

ooooo great point about circular/self references, I will fix that

and that setting *should* apply to both models and avatars, since they render using the same code.  shapes/text/image/gizmos aren’t affected

libraries/material-networking/src/material-networking/TextureCache.cpp
```diff
@@ -1374,3 +1383,22 @@ void TextureCache::updateSpectatorCameraNetworkTexture() {
...
+    }
+    return NetworkTexturePointer();
+}
+
+std::function<gpu::TexturePointer()> Texture::getTextureForUUIDOperator(const QUuid& uuid) {
+    if (_unboundTextureForUUIDOperator) {
+        return std::bind(_unboundTextureForUUIDOperator, uuid);
+    }
+    return nullptr;
+}
```
ctrlaltdavid, line 1404: Add CR ad end of file.

libraries/procedural/src/procedural/Procedural.h
```diff
@@ -40,7 +40,7 @@ const size_t MAX_PROCEDURAL_TEXTURE_CHANNELS{ 4 };
  *     If a procedural material contains a vertex shader, the bounding box of the material entity is used to cull the object to which the material is applied.
  * @property {string} fragmentShaderURL - A link to a fragment shader.  Currently, only GLSL shaders are supported.  The shader must implement a different method depending on the version.
  *     <code>shaderUrl</code> is an alias.
- * @property {string[]} channels=[] - An array of input texture URLs.  Currently, up to 4 are supported.
+ * @property {string[]} channels=[] - An array of input texture URLs or entity IDs.  Currently, up to 4 are supported.
```
ctrlaltdavid, line 43:
```suggestion
 * @property {string[]} channels=[] - An array of input texture URLs or Material entity IDs.  Currently, up to 4 are supported.
```
I assume?

libraries/procedural/src/procedural/ProceduralMaterialCache.cpp
```diff
@@ -75,7 +76,8 @@ bool NetworkMaterialResource::parseJSONColor(const QJsonValue& array, glm::vec3&
  * A material or set of materials used by a {@link Entities.EntityType|Material entity}.
  * @typedef {object} Entities.MaterialResource
  * @property {number} materialVersion=1 - The version of the material. <em>Currently not used.</em>
- * @property {Entities.Material|Entities.Material[]} materials - The details of the material or materials.
+ * @property {Entities.Material|Entities.Material[]|string} materials - The details of the material or materials, or the ID of another
+ *     material entity.
```
ctrlaltdavid, line 80:
```suggestion
 *     Material entity.
```
I assume? Or perhaps it's an Image entity?

libraries/procedural/src/procedural/ProceduralMaterialCache.cpp
```diff
@@ -136,12 +149,12 @@ NetworkMaterialResource::ParsedMaterials NetworkMaterialResource::parseJSONMater
  *     Set to <code>"fallthrough"</code> to fall through to the material below. <code>"hifi_pbr"</code> model only.
  * @property {number|string} scattering - The scattering, range <code>0.0</code> &ndash; <code>1.0</code>. 
  *     Set to <code>"fallthrough"</code> to fall through to the material below. <code>"hifi_pbr"</code> model only.
- * @property {string} emissiveMap - The URL of the emissive texture image. 
+ * @property {string} emissiveMap - The URL of the emissive texture image, or an entity ID.
```
ctrlaltdavid, line 152:
```suggestion
 * @property {string} emissiveMap - The URL of the emissive texture image, or a Material entity ID.
```
I assume? Or perhaps it's an Image entity? Similarly, below.

libraries/procedural/src/procedural/ReferenceMaterial.h
```diff
@@ -0,0 +1,69 @@
...
+    bool isUnlit() const override;
+    glm::vec3 getAlbedo(bool SRGB = true) const override;
+    float getMetallic() const override;
+    float getRoughness() const override;
+    float getScattering() const override;
+    bool resetOpacityMap() const override;
+    const graphics::Material::TextureMaps& getTextureMaps() const override;
+    //glm::mat4 getTexCoordTransform(uint i) const override; // use my actual transform, instead of the original
+    glm::vec2 getLightmapParams() const override;
+    //glm::vec2 getMaterialParams() const override; // use my actual params, instead of the original
```
ctrlaltdavid, line 36: Extraneous comments?

libraries/procedural/src/procedural/ReferenceMaterial.h
```diff
@@ -0,0 +1,69 @@
...
+    uint64_t getFadeStartTime() const override;
+    bool hasVertexShader() const override;
+    void prepare(gpu::Batch& batch, const glm::vec3& position, const glm::vec3& size, const glm::quat& orientation,
+                 const uint64_t& created, const ProceduralProgramKey key = ProceduralProgramKey()) override;
+    void initializeProcedural() override;
+
+    bool isReference() const override { return true; }
+    std::function<graphics::MaterialPointer()> getReferenceOperator() const { return _materialForUUIDOperator; }
+
+    static void setMaterialForUUIDOperator(std::function<graphics::MaterialPointer(QUuid)> materialForUUIDOperator) { _unboundMaterialForUUIDOperator = materialForUUIDOperator; }
```
ctrlaltdavid, line 60: Line too long.

ctrlaltdavid (changes requested):

libraries/procedural/src/procedural/Procedural.h
HifiExperiments: image or web entities.  generally, anything that supported texture URLs can now reference those types by ID, and just material entities can refer to material entities.  I’m not sure how to best describe this in the docs succinctly, since they’re referenced all over the place

ctrlaltdavid: I've updated my suggested change. The API docs should say what acceptable values are / how a property can be used - even if this information is repeated multiple times at different locations because a user typically will be just looking at one particular piece of the docs, not reading the whole thing.

libraries/entities-renderer/src/RenderableEntityItem.cpp
```diff
@@ -491,6 +495,185 @@ void EntityRenderer::removeMaterial(graphics::MaterialPointer material, const st
...
+    }
+
+    return EntityRenderer::getBound(args);
+}
+
+void EntityRenderer::updateShapeKeyBuilderFromMaterials(ShapeKey::Builder& builder) {
+    MaterialMap::iterator materials;
+    {
+        std::lock_guard<std::mutex> lock(_materialsLock);
+        materials = _materials.find("0");
```
ctrlaltdavid, line 637:
```suggestion
        materials = _materials.find("0");

```

libraries/entities-renderer/src/RenderableEntityItem.cpp
```diff
@@ -491,6 +495,185 @@ void EntityRenderer::removeMaterial(graphics::MaterialPointer material, const st
...
+
+    graphics::MaterialKey drawMaterialKey = materials->second.getMaterialKey();
+    if (drawMaterialKey.isUnlit()) {
+        builder.withUnlit();
+    }
+
+    auto pipelineType = getPipelineType(materials->second);
+    if (pipelineType == Pipeline::MATERIAL) {
+        builder.withMaterial();
+
```
ctrlaltdavid, line 665:
```suggestion
```

libraries/entities-renderer/src/RenderableImageEntityItem.cpp
```diff
@@ -132,10 +166,26 @@ void ImageEntityRenderer::doRender(RenderArgs* args) {
...
+    if (pipelineType == Pipeline::PROCEDURAL) {
+        auto procedural = std::static_pointer_cast<graphics::ProceduralMaterial>(materials.top().material);
+        transparent |= procedural->isFading();
+        procedural->prepare(*batch, transform.getTranslation(), transform.getScale(), transform.getRotation(), _created, ProceduralProgramKey(transparent));
+    } else if (pipelineType == Pipeline::SIMPLE) {
+        batch->setResourceTexture(0, _texture->getGPUTexture());
+    } else {
+        if (RenderPipelines::bindMaterials(materials, *batch, args->_renderMode, args->_enableTexturing)) {
+            args->_details._materialSwitches++;
+        }
```
ctrlaltdavid, line 179:
```suggestion
    } else if (RenderPipelines::bindMaterials(materials, *batch, args->_renderMode, args->_enableTexturing)) {
        args->_details._materialSwitches++;
    }
```

libraries/material-networking/src/material-networking/TextureCache.h
```diff
@@ -39,6 +39,12 @@ class Texture {
 public:
     gpu::TexturePointer getGPUTexture() const { return _textureSource->getGPUTexture(); }
     gpu::TextureSourcePointer _textureSource;
+
+    static std::function<gpu::TexturePointer()> getTextureForUUIDOperator(const QUuid& uuid);
+    static void setUnboundTextureForUUIDOperator(std::function<gpu::TexturePointer(const QUuid&)> textureForUUIDOperator) { _unboundTextureForUUIDOperator = textureForUUIDOperator; }
```
ctrlaltdavid, line 44: Line probably too long.

ctrlaltdavid (changes requested):

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Interface-PR1102-9329adc-_632cfd03.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc-.tar_632cfd03.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc-_632cfd03.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc-.tar_632cfd03.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc-_632cfd03.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc-_632cfd03.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-9329adc_632cfd03.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-9329adc_632cfd03.apk

libraries/material-networking/src/material-networking/TextureCache.cpp
```diff
@@ -252,6 +254,10 @@ NetworkTexturePointer TextureCache::getTexture(const QUrl& url, image::TextureUs
     if (url.scheme() == RESOURCE_SCHEME) {
         return getResourceTexture(url);
     }
+    QString decodedURL = url.toDisplayString(QUrl::FullyDecoded);
```
HifiExperiments, line 257: I've integrated the requested changes and (I think) fixed the infinite loop issues, but after merging master, referencing textures by ID (this line specifically) no longer works.  the issue is that entity IDs are treated as texture urls, so they are encoded.  an entity ID like "{2cc07c81-d04f-4d80-b29a-99601ad92769}" becomes "%7B2cc07c81-d04f-4d80-b29a-99601ad92769%7D".  this line is supposed to bring back the curly brackets...but it seems like it no longer does?  I'm not sure if this is a regression due to the Qt upgrade or what.  @ctrlaltdavid do you have any ideas?  I wonder if this affects other areas of the codebase

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b-.tar_bdbb8808.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b-.tar_bdbb8808.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b-_bdbb8808.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Interface-PR1102-b4e2c7b-_bdbb8808.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-b4e2c7b_bdbb8808.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b_bdbb8808.apk

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b-_bdbb8808.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-b4e2c7b-_bdbb8808.dmg

libraries/material-networking/src/material-networking/TextureCache.cpp
ctrlaltdavid: I assume that the leading `%7B` and trailing `%7D` are actually there. ...
Perhaps `QUrl::PrettyDecoded` would work?
Or do QString replaces on `%7B` and `%7D`.
Do you have a JSON with example entity etc. for easy dev testing?

libraries/entities/src/EntityItemProperties.cpp
```diff
@@ -1379,7 +1380,8 @@ EntityPropertyFlags EntityItemProperties::getChangedProperties() const {
 /*@jsdoc
  * The <code>"Web"</code> {@link Entities.EntityType|EntityType} displays a browsable web page. Each user views their own copy
  * of the web page: if one user navigates to another page on the entity, other users do not see the change; if a video is being
- * played, users don't see it in sync. It has properties in addition to the common
+ * played, users don't see it in sync. Internally, a Web Entity is rendered as a non-repeating, upside down texture, so additional
+ * transformations may be necessary if you reference a web entity texture by UUID. It has properties in addition to the common
```
ctrlaltdavid, line 1384:
```suggestion
 * played, users don't see it in sync. Internally, a Web entity is rendered as a non-repeating, upside down texture, so additional
 * transformations may be necessary if you reference a Web entity texture by UUID. It has properties in addition to the common
```

libraries/procedural/src/procedural/ProceduralMaterialCache.cpp
```diff
@@ -136,12 +149,14 @@ NetworkMaterialResource::ParsedMaterials NetworkMaterialResource::parseJSONMater
...
  * @property {number|string} scattering - The scattering, range <code>0.0</code> &ndash; <code>1.0</code>. 
  *     Set to <code>"fallthrough"</code> to fall through to the material below. <code>"hifi_pbr"</code> model only.
- * @property {string} emissiveMap - The URL of the emissive texture image. 
- *     Set to <code>"fallthrough"</code> to fall through to the material below. <code>"hifi_pbr"</code> model only.
- * @property {string} albedoMap - The URL of the albedo texture image. 
- *     Set to <code>"fallthrough"</code> to fall through to the material below. <code>"hifi_pbr"</code> model only.
- * @property {string} opacityMap - The URL of the opacity texture image. Set the value the same as the <code>albedoMap</code> 
- *     value for transparency. 
+ * @property {string} emissiveMap - The URL of the emissive texture image, or an entity ID.  An entity ID may be that of an
+ *     Image or a Web entity.  Set to <code>"fallthrough"</code> to fall through to the material below.
```
ctrlaltdavid, line 153:
```suggestion
 *     Image or Web entity.  Set to <code>"fallthrough"</code> to fall through to the material below.
```
And similar below.

ctrlaltdavid (changes requested):

libraries/material-networking/src/material-networking/TextureCache.cpp
HifiExperiments: I worked around this with `QUrl::fromPercentEncoding` 👍 

a test is to make a web entity and an image entity, and set the image entity's image url to the web entities ID

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e-_12833b9c.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e-.tar_12833b9c.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e-.tar_12833b9c.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Interface-PR1102-bff386e-_12833b9c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e-_12833b9c.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e-_12833b9c.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-Quest-Alpha-PR1102-bff386e_12833b9c.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1102/Vircadia-PR1102-bff386e_12833b9c.apk

ctrlaltdavid (approved):

SilverfishVR:
Initial tests on PR1102-bff386e

* Image and web entities as texture works, with the caveat of the web's being flipped and colorspace is wrong, as expected 👍 
* enitity ID as Material URL works, with individual scaling 👍 
* reference by material name (```?materialName``` or ```#materialName```) works great for named materials and updates consistently 👍, but should I not be able to do ```?1``` to get the second material in a multi material or do I remember that wrong?
* Using your procedural shader example just show as Black  🤷

HifiExperiments:
@SilverfishVR thank you for testing!  my bad, there was a typo in the test instructions for the procedural test, the channels needed to be in the procedural block instead of below.   updated the description and confirmed it works.

indexing into the multi material like that has never been implemented, it would only work like that if you explicitly named them "0", "1", etc.

SilverfishVR:
> there was a typo in the test instructions for the procedural test,

oh right, I see that, it works well now 👍 As a side note, I used a shader to play around with the color correction and found that simply adding a gamma correction of 2.2 and making it emissive makes it look almost exactly like the web entity, albeit a bit blurry if you have AA enabled.
Screenshot 2021-07-11 220229
left: shader on model. right: web entity. No AA.   https://vircadia-shders.glitch.me/showWeb.fs

I wonder if a shader option to disable AA (like it is for web and image entities) could be a thing 🤔 

> indexing into the multi material like that has never been implemented

yea, I don't know why I had that idea 🤷 

TL:DR Ship it 😄

HifiExperiments:
@SilverfishVR amazing!!  as for the AA, I think we should leave it for now, #576 will theoretically help with that and otherwise yeah we can always expose a way to disable AA via material

PR #1174, opened 2021-04-11
UI for setting metaverse server on login

mirrorwolf:
I tried hosting my own private metaverse server so I wouldn't have to make a centralized account, and it worked great... there's just one problem:

The official client can't connect to it!

This fixes that, sort of. It lets you set HIFI_METAVERSE_SERVER (like the domain server already has) and also HIFI_AUTHABLE_HOSTNAMES (a comma-separated list) to set which domains get your credentials passed to them.

### DIFF
--- a/interface/resources/qml/LoginDialog/LinkAccountBody.qml
+++ b/interface/resources/qml/LoginDialog/LinkAccountBody.qml
@@ -131,8 +131,9 @@ Item {
         if (!isLoggingInToDomain) {
             var savedUsername = Settings.getValue("keepMeLoggedIn/savedUsername", "");
             emailField.text = keepMeLoggedInCheckbox.checked ? savedUsername === "Unknown user" ? "" : savedUsername : "";
-            
+
             var metaverseServer = Settings.getValue("private/selectedMetaverseURL", "");
+            console.log("Saved metaverse server:", metaverseServer);
             metaverseServerField.text = metaverseServer;
         } else {
             // ####### TODO
@@ -384,7 +385,7 @@ Item {
                 font.pixelSize: linkAccountBody.textFieldFontSize
                 styleRenderType: Text.QtRendering
                 anchors {
-                    top: loginDialogTextContainer.bottom
+                    top: passwordField.bottom
                     topMargin: 1.5 * hifi.dimensions.contentSpacing.y
                 }
                 placeholderText: "Metaverse Server (optional)"
@@ -403,6 +404,7 @@ Item {
                         case Qt.Key_Return:
                             event.accepted = true;
                             if (!isLoggingInToDomain) {
+                                console.log("Setting metaverse server to", metaverseServerField.text);
                                 Settings.setValue("private/selectedMetaverseURL", metaverseServerField.text);
                             }
                             linkAccountBody.login();
@@ -426,9 +428,9 @@ Item {
                 color: hifi.colors.white;
                 visible: !isLoggingInToDomain
                 anchors {
-                    top: passwordField.bottom;
+                    top: metaverseServerField.bottom;
                     topMargin: hifi.dimensions.contentSpacing.y;
-                    left: passwordField.left;
+                    left: metaverseServerField.left;
                 }
                 onCheckedChanged: {
                     Settings.setValue("keepMeLoggedIn", checked);
@@ -749,4 +751,4 @@ Item {
                 break;
         }
     }
-metaverseServerField
+}
--- a/interface/resources/qml/LoginDialog/LinkAccountBody.qml
+++ b/interface/resources/qml/LoginDialog/LinkAccountBody.qml
@@ -74,6 +74,19 @@ Item {
     }
 
     function login() {
+        // make sure the metaverse server is set so we don't send your password to the wrong place!
+        if (!isLoggingInToDomain) {
+            Settings.setValue("private/selectedMetaverseURL", metaverseServerField.text);
+        }
+        
+        if (keepMeLoggedInCheckbox.checked) {
+            if (!isLoggingInToDomain) {
+                Settings.setValue("keepMeLoggedIn/savedUsername", emailField.text);
+            } else {
+                // ####### TODO
+            }
+        }
+
         if (!isLoggingInToDomain) {
             loginDialog.login(emailField.text, passwordField.text);
         } else {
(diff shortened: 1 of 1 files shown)


digisomni:
https://github.com/vircadia/vircadia/blob/master/libraries/networking/src/MetaverseAPI.cpp Was this of any use? I believe we should probably document that setting in the docs, however, if you set that using `Settings.setValue` in the scripting console and restart Interface does it start to work?

digisomni:
Also, NetworkingConstants is supposed to hold mainly constants, so if there is functionality missing then it should probably be added to MetaverseAPI.cpp.

mirrorwolf:
Oh wow, there's actual API for that? Nice! I totally didn't see that _anywhere_. I was going off the old fork that swapped in support for some specific alternative server that was floating around.

digisomni:
Well, there's a class we made early on, it's not been touched since as we don't switch Metaverse servers often, so it's gone relatively untested the ability to switch on the fly! But, if you would like to play with that then hopefully we'll find what's missing (if anything). :)

mirrorwolf:
It looks like that uses the HIFI_METAVERSE_URL _already_? I just tested it and yep, yep, it works perfectly and lets me log in with my own server on the unmodified client. Whoops.

digisomni:
Ah, cool, then that's a failure of us not documenting it appropriately. I hope it didn't waste too much of your time. :) We're going to be renaming it to "VIRCADIA_METAVERSE_URL" in the future, just a heads up.

mirrorwolf:
Awesome, that's good to know. I'll add that to my domain server configuration file too then. (:

So, sounds like we just need to build UI for this, and a setting to persist it!

digisomni:
There are a lot of settings that would benefit from having quick and easy settings for them. Unfortunately, our UI is mainly QML based so it's a non-trivial matter to add in more settings for those at a moment's notice.

Mayhaps it would be in order to make multple settings tabs that are web based so we can quickly add and iterate on settings + functionality for 'em.

mirrorwolf:
Web-based settings would be neat! Although in this case, I'm wondering if just an additional field on the login screen would be better, having it automatically persist when you hit "log in". Assuming we can do that – I'm not entirely sure if that page is client-side or metaverse-server-side.

digisomni:
That page is client-side, it would be a handy addition there I reckon. However, we'd want to confirm that the change propagates and works in real-time without a restart, did it work for you without having to restart?

mirrorwolf:
Nope, in fact I can't get it to set. The console is saying something about a "restricted write".

digisomni:
Oh dang, that's because it is a restricted write. Only a QML page can set that then. All settings that have "private/" in them are restricted from scripts. We don't have a full scripting security system yet. We'll probably look into implementing the foundation for that when we start moving to our new scripting engine.

mirrorwolf:
Hah, that makes sense! In that case I might as well just start hacking together a UI and see if it just works.

digisomni:
This file https://github.com/vircadia/vircadia/blob/master/interface/resources/qml/LoginDialog.qml and .qml files related to it (QML relies on a component structure so you'll see multiple files pertaining to one "page" sometimes) might be a good place to look.

mirrorwolf:
Thanks!

I got a text box added, and it writes to the setting just fine, and reads from it just fine... but it only seems to actually take effect after a restart. Hm.

mirrorwolf:
Oops. That should have been a complete overwrite not a merge, no wonder things broke spectacularly. Fixed now, hopefully.

digisomni:
> 
> 
> Thanks!
> 
> I got a text box added, and it writes to the setting just fine, and reads from it just fine... but it only seems to actually take effect after a restart. Hm.

It's possible that there are classes in the codebase and in scripts pulling that metaverse URL to set it once at runtime instead of constantly reacquiring it.

digisomni:
The easiest thing to do would be to have a message appear when changing the metaverse on the login that "Changes require restart to take effect." or something like that.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221-.tar_f37e69f3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221-_f37e69f3.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221-.tar_f37e69f3.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221-_f37e69f3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221-_f37e69f3.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-Quest-Alpha-PR1174-71fa221_f37e69f3.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1174/Vircadia-PR1174-71fa221_f37e69f3.apk

mirrorwolf:
Hm! Maybe instead of having UI on the login screen I should add a menu item "change metaverse server" right below the "log in / sign up" one.

digisomni:
> 
> 
> Hm! Maybe instead of having UI on the login screen I should add a menu item "change metaverse server" right below the "log in / sign up" one.

That might be better! You might wanna take a look at the Menu.cpp class then.

daleglass:
Hello, what's the state of this one? From the discussion it seems there are still issues to be worked out

mirrorwolf:
Oops, I completely just disappeared for the past, uh, several months – sorry about that! Things still need to be worked out, yeah; it looks like I've got the beginnings of a branch for the menu item. My build setup still seems to work, fortunately, and the branch rebased just fine, so now I'm building the client again...

mirrorwolf:
I've got it building and a menu item showing up! Now I just need to figure out how to add a dialog to change it... I bet if we use an in-world dialog like the login one it'd be accessible from VR too, which is a plus.

Could have a "Logging into https://wherever [Change]" link on the login dialog, that'd open up the switch-metaverse-server one. In addition to the menu item. (Not sure if the menu bar is accessible from VR.)

mirrorwolf:
Hm, maybe it would be a good idea to just have this in the login dialog after all, instead of making an entirely new one. That would also be good if the rest of the code ever gets updated to be able to use the new URL immediately.

ksuprynowicz:
I tested it and it works well, but needs restart after changing metaverse server. I will run it with debugger, and find what needs to be updated.
Another problem is that Explore app seems to download places and domains from official metaverse server instead of the one that was set.

digisomni:
This can be merged after a few more revisions from Namark and a round of testing.

ksuprynowicz:
It's an old branch and requires restart every time metaverse server is changed. The new one is here: https://github.com/overte-org/overte/pull/2

digisomni:
I expect this PR can be updated with the latest commits and see what happens.

PR #1156, opened 2021-04-04
Obtain script backtrace when a log event is generated in C++ code

daleglass:
This is a very, very early prototype and doesn't work. Here's the overall idea:

When a C++ function logs something, such as:

```
[04/04 23:16:38] [WARNING] [hifi.scriptengine] Script.include() ignoring file path -- outside of standard libraries:  "" "/home/dale/git/vircadia/build-logging/interface/scripts"
```

We want to figure out where that came from. Which script resulted in this function being called? And this works to some extent:
```
Script context:
	:-1 include
	file:///home/dale/git/vircadia/build-logging/interface/scripts/defaultScripts.js:153 runDefaultsTogether
	file:///home/dale/git/vircadia/build-logging/interface/scripts/defaultScripts.js:185 
[04/04 23:16:38] [WARNING] [hifi.scriptengine] Script.include() ignoring file path -- outside of standard libraries:  "" "/home/dale/git/vircadia/build-logging/interface/scripts"
```

The problem is that a problem can be found at any arbitrary depth, not just in whatever QScriptEngine calls directly. So I figured out a way around that. Qt has the `QScriptEngineAgent`, which allows us invoke code at interesting points, such as script function calls, including calls to native code.

So the idea is:

1. On every native function call from ScriptEngine,  ScriptEngineLoggingAgent::functionEntry is called.
2. We generate a backtrace.
3. We give this to ScriptContextHelper, which remembers backtraces per-thread.
4. Execution continues, native function runs
5. Possibly more native code runs
6. Code logs an event. Execution eventually ends up in LogHandler.
7. LogHandler checks ScriptContextHelper, prints the backtrace if there's any.
8. Eventually the native code finishes running
9. ScriptEngineLoggingAgent::functionExit is called
10. ScriptContextHelper wipes the current backtrace.

Unfortunately it doesn't quite work. Problems:

* There's probably something thread-related horribly wrong somewhere
* functionEntry seems to be called multiple times per event, hence the stack
* Trying to generate a backtrace sometimes results in functionEntry calling itself through the Qt code.
* I tried constructing a backtrace by hand, hoping that avoiding examining objects and arguments and just printing names would be safe, but it doesn't seem to help.

```
#15 0x00007ffff51720a6 in ScriptEngineLoggingAgent::functionEntry(long long) (this=0x16e8a1a0, scriptId=-1) at /home/dale/git/vircadia/vircadia-master/libraries/script-engine/src/ScriptEngineLoggingAgent.cpp:37
[...]

PR #1206, opened 2021-05-09
drop remaining references to QScriptEngineDebugger

odysseus654:
This PR drops all remaining references to QScriptEngineDebugger.

It appears that this *used* to be available by holding down [Shift] while launching Vircadia but this feature was disabled due to launching deadlocks.  This appears to currently still be available by sticking #debug at the end of a scripting URL.

I have verified that this compiles and links but have not attempted to run or verify that scripts are running properly.

### DIFF
--- a/libraries/script-engine/src/ScriptEngine.cpp
+++ b/libraries/script-engine/src/ScriptEngine.cpp
@@ -36,8 +36,6 @@
 #include <QtScript/QScriptValue>
 #include <QtScript/QScriptValueIterator>
 
-#include <QtScriptTools/QScriptEngineDebugger>
-
 #include <shared/LocalFileAccessGate.h>
 #include <shared/QtHelpers.h>
 #include <shared/AbstractLoggerInterface.h>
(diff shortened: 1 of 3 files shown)


odysseus654:
My biggest reason for opening up this PR is to try to figure out what to do with this (Something that has both UI impact and is QtScript-specific).  I can either drop it from the build or try to find some way of stretching this code across the proxy interface.  I have options here, but I don't have to worry about making it work if it's been deleted from the system.

That and also because this is a potentially visible change that can be tested independently of the other code shifting I've been doing.

daleglass:
Mm. So I've thought about it. I think while we could indeed merge this, it would be ideal not to. Debugging is important, and in fact I've been trying to extend it with #1156.

I can let that one go, but still, my view is that any cool functionality a scripting engine can provide that can be helpful for either the user or the developer is a very good thing to have in a project in need of more users, and not a lot of developers.

I think it shouldn't be too hard to provide some sort of general "enable debugging" interface, then implement it in an engine-specific way.

odysseus654:
Okay, I will try to port this so it works within #1200 then.  It will be QtScript specific and likely not applicable to V8 without a fair amount of UI development though.

odysseus654:
I do need to look at #1156, that may be relevant in what I'm currently doing on #1200 as well.

digisomni:
As long as the components are there, we can add UI for it later. :)

odysseus654:
Yah, the real question I guess is whether the wires I rig up to get this working would be useful at all for V8.

But again, that question is really not part of *any* PR right now and can be adjusted later.

digisomni:
After talking to Heather, I realize what she meant.. Basically, this all goes away when we upgrade to Qt6 anyway because support will be dropped. Since we don't use the QtScript debugger right now then it's unlikely we will end up doing so within the next 6 or so months (within that timeframe Qt6 might be ready for us?)

So, then maybe having her port this over is not necessary?

daleglass:
Personally I think that a port to Qt6 in 6 months is very optimistic and unlikely. Qt6 still doesn't have QtWebEngine at this point, then it'll likely take fixing a bunch of stuff, no idea how much exactly. Before switching to Qt6 officially we'll likely need to do some fixes to the code, make new packages, test on all platforms, etc. It's probably doable in 6 months, but that's only if somebody is enthusiastic about making it happen.

Also there's no huge rush with making it happen either, as I don't see Qt5 go anywhere for quite some time still. I think it's definitely a very good thing to keep up with Qt development, but not so urgent that it absolutely needs to be done as soon as possible.

Anyway, yes, I realize this is going to go away eventually. My view here is:

1. It's desirable to preserve functionality if we can. We want to get stuff done after all, and debugging is a good feature to have.
2. This should be a doable thing. It's just 118 lines of code. 
3. Related to point 2, I think think this is a good test for the refactoring. Something like this should be possible in the new design, and it's small enough that I think it should be possible to work it in without that much trouble.

If it turns out that there's some serious technical block to making this work, then sure, I think we can let this one go. But I think it's worthwhile to at least try.

JulianGro:
This doesn't actually show up in the Sphinx documentation at all. Just for reference.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-Interface-PR1206-bf21d38-_04f8ccd2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-bf21d38-.tar_04f8ccd2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-bf21d38-.tar_04f8ccd2.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-bf21d38-_04f8ccd2.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-bf21d38-_04f8ccd2.dmg

odysseus654:
I think the biggest issue is that ScriptEngine::runDebuggable() is a modified version of ScriptEngine::run(), which will be operating completely outside of the script-specific area.  So all references to _debugger need to be pruned out of here and pushed on to the script-specific area (which I think here is ScriptEngine::evaluate).  The alternative is to create a proxy version of QScriptEngineDebugger (ScriptEngineDebuggerPointer?), but I'm 99% sure that this interface will be completely useless with a non-Qt scripting engine.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-Interface-PR1206-719737a-_04f8ccd2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-719737a-.tar_04f8ccd2.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-719737a-.tar_04f8ccd2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-719737a-_04f8ccd2.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-719737a-_04f8ccd2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1206/Vircadia-PR1206-719737a-_04f8ccd2.exe

odysseus654:
BTW after reviewing #1156 I can say that this PR is unrelated to that and would not affect that code.

daleglass (approved):

digisomni (approved):
Did minimal testing on Windows 10, appears to be working fine and scripts are running fine.

PR #1225, opened 2021-05-23
fix PR autobuild for ubuntu-18.04, android with correct cmake version

ArcadeFever:


ArcadeFever:
The github actions virtual-environment 20210504.1 was last successful Android build.  Checking the https://github.com/actions/virtual-environments/blob/ubuntu18/20210504.1/images/linux/Ubuntu1804-README.md shows in the Android section, cmake version 3.10.2. 
Then github actions virtual-environment was updated to 20210517.1 and checking https://github.com/actions/virtual-environments/blob/ubuntu18/20210517.1/images/linux/Ubuntu1804-README.md shows Android cmake now contains both 3.10.2 and 3.18.1 with the latter being the default, and thus breaking our build.
So we either need to find a way to roll back the github actions virtual environment to 20210504.1 or get the android build to switch to 3.10.2 as default.

JulianGro:
An updated CMake version *shouldn't* really cause any issues. Also it looks like it specifically has a problem with the CMake version number.
Maybe the version check fails because it is called "3.18.1-g262b901" instead of "3.18.1"?

Just for reference, here is the issue that caused them to upgrade CMake: https://github.com/actions/virtual-environments/issues/3321
Also apparently we aren't alone with this issue:
https://github.com/flutter/flutter/issues/75600

Someone on another project apparently fixed it by just nuking the Android CMake 3.18.1 install: https://github.com/cossacklabs/themis/pull/824/files
I would actually suggest doing the same, since I would consider it a mess to even use non-system CMake, and Google software hasn't exactly proven stable looking at some of our other dependencies.

JulianGro:
I just tried to find out the path and the standard path for sdkmanager on Ubuntu 18.04 seems to be
```
~/Android/Sdk/tools/bin/sdkmanager
```
Not sure if GitHub will be using that one though since that is a horrible path to save these things to.
Other options might be: (in order of how likely they seem to me)
```
~/.local/share/Android/Sdk/tools/bin/sdkmanager
~/.local/share/Android/sdk/tools/bin/sdkmanager
~/.config/Android/Sdk/tools/bin/sdkmanager
~/.config/Android/sdk/tools/bin/sdkmanager
```

The sdkmanager file is just a shell script, so you could test for them using cat (for lack of better ideas)
E.g. `cat ~/.local/share/Android/Sdk/tools/bin/sdkmanager`

ArcadeFever:
I wish I knew how to stop some of these actions from running...

HifiExperiments:
the “matrix” “include” list at the top defines all the different builds to produce, you could comment out all but the android one

ArcadeFever:
Thanks to @JulianGro and @HifiExperiments  for the help, we've now got a successful android build!  Of course since the PR isn't labeled for upload yet, it still fails at the end.  I'll clean up a few things and submit last commit.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5-_71ae846d.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5-.tar_71ae846d.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5-.tar_71ae846d.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5-_71ae846d.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-Interface-PR1225-c934df5-_71ae846d.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5-_71ae846d.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-PR1225-c934df5_71ae846d.apk
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1225/Vircadia-Quest-Alpha-PR1225-c934df5_71ae846d.apk

digisomni (approved):

ArcadeFever:
I tested the quest apk on my Quest1, and unfortunately it insta-quits.  I then tried the last successful PR build before the github virtual environments update (16 days ago in PR 1207) and that one also insta-quits.  So I'm not sure when PR builds of the quest apk stopped working, but this PR is irrespective of that and resolves the current build issue.

I'll have to look back at when the last PR build was made using the NDK 19 as I suspect NDK 21 is causing the insta-quits on Quest1.

HifiExperiments:
I think at a certain point we’ll have to just finally upgrade everything to a recent ndk and fix the issues…I think that’s the only path forward for fixing quest 2 builds too

JulianGro:
Can we squash all these commits?
Personally I would just make a new branch, copy that little change in, and force push that branch to this PRs branch.

digisomni:
I'm intending to use "squash and merge".

PR #1273, opened 2021-07-03
Optimize Android GitHub Actions

JulianGro:
The Action for Android builds seems to do an extra unneeded step that takes quite some time. (~13 minutes)


daleglass:
The build failures may be because we broke Mac builds early on, this should be rebased to apply on the newer working code. And of course there's the conflict.

JulianGro:
This is a "do not merge" I was experimenting with it to try to get it to not build the dependencies twice but didn't get very far. I guess I should close this.

PR #1321, opened 2021-08-26
Build Windows PR builds with memory debugging.

digisomni:


odysseus654:
Am I the only one that is getting hit with initialization bugchecks when memory debugging is active on windows?

PR #1354, opened 2021-09-21
Get make-deb-server working again

odysseus654:
This patch is in response to errors creating .DEB package files, due to the change in location for qt5 libraries.

Some attempt has been made to detect and handle system-qt compilations, but as this was not one of them (did someone post a list of the required packages to install somewhere?) this has not been tested.

make-rpm-server has not been tested either but likely requires similar changes to compensate for the new library location.


digisomni (approved):
Works for the out-of-the-box use case.

Issue #1355, opened 2021-09-21
High CPU usage of assignment client when idle on Linux


ksuprynowicz:
With current master build after starting completely empty domain server and assignment client with no users on it, current assignment client (Audio Mixer) uses up to 80% of a CPU core on Xeon E5-2690 v2, and Avatar Mixer thread uses about 36% of a CPU core.
It seems that both of them spend most time on `wait` and `notify` methods.
Here's the CPU usage for AudioMixer:
```
-   93.77%     0.03%  AudioMixerSlave  assignment-client        [.] AudioMixerSlaveThread::run
   - 93.75% AudioMixerSlaveThread::run
      - 73.96% AudioMixerSlaveThread::wait
         - 54.29% std::condition_variable::wait<AudioMixerSlaveThread::wait()::{lambda()#1}>
            + 54.05% std::condition_variable::wait
         - 11.08% std::unique_lock<std::mutex>::~unique_lock
            + std::unique_lock<std::mutex>::unlock 
         + 8.32% std::unique_lock<std::mutex>::unique_lock
      - 19.55% AudioMixerSlaveThread::notify
         + 9.74% std::unique_lock<std::mutex>::unique_lock
         + 9.33% std::unique_lock<std::mutex>::~unique_lock
```
And for AvatarMixer:
```
-   90.36%     0.03%  AvatarMixerSlav  assignment-client        [.] AvatarMixerSlaveThread::run
   - 90.33% AvatarMixerSlaveThread::run
      - 71.15% AvatarMixerSlaveThread::wait
         - 52.61% std::condition_variable::wait<AvatarMixerSlaveThread::wait()::{lambda()#1}>
            - 52.35% std::condition_variable::wait
               - 31.99% __pthread_cond_wait (inlined)
                  + __pthread_cond_wait_common (inlined)
               - 20.31% __pthread_mutex_cond_lock
                  + __lll_lock_wait
         + 10.54% std::unique_lock<std::mutex>::~unique_lock
         + 7.72% std::unique_lock<std::mutex>::unique_lock
      - 18.87% AvatarMixerSlaveThread::notify
         + 10.25% std::unique_lock<std::mutex>::unique_lock
         + 8.17% std::unique_lock<std::mutex>::~unique_lock
```

odysseus654:
[...]

PR #1358, opened 2021-09-23
Optimization coding pass over AudioMixerSlavePool and AvatarMixerSlavePool pt. 1.

odysseus654:
This is an optimization pass over the mutex logic in AudioMixerSlavePool and AvatarMixerSlavePool (which is effectively the same code).  Changes made are:
* The single _mutex has been replaced by _poolMutex and _slaveMutex, and
* An "if" statement now wraps the _poolCondition "wake" logic, to reduce spurious wakes

I have not tested this code beyond ensuring it compiles.
I do not know whether this would resolve issue #1355 as this may not be the ultimate cause (these two assignment clients are pretty much a tight while() loop repeatedly calling the pool whether anyone is connected or not) but it hopefully reduces the amount of lock contention going on

Testing:
- Smoketesting: make sure that audio and avatars still work on a server built with this (requires a server with multiple users)
- Race conditions: exercise a server long enough to know whether there are any lockups or crashes that occur in these two areas
- Profile: see if this improves the profiling numbers that have been described in issue #1355

### DIFF
--- a/assignment-client/src/audio/AudioMixerSlavePool.cpp
+++ b/assignment-client/src/audio/AudioMixerSlavePool.cpp
@@ -189,13 +189,14 @@ void AudioMixerSlavePool::resize(int numThreads) {
         while (_numStopped != (_numThreads - numThreads)) {
             {
                 Lock slaveLock(_slaveMutex);
-                _numStarted = _numFinished = _numStopped;
+                _numStarted = _numFinished = 0;
                 _slaveCondition.notify_all();
             }
             _poolCondition.wait(poolLock, [&] {
                 assert(_numFinished <= _numThreads);
                 return _numFinished == _numThreads;
             });
+            assert(_numStopped == (_numThreads - numThreads));
         }
 
         // ...wait for threads to finish...
--- a/assignment-client/src/audio/AudioMixerSlavePool.cpp
+++ b/assignment-client/src/audio/AudioMixerSlavePool.cpp
@@ -53,15 +53,16 @@ void AudioMixerSlaveThread::wait() {
 }
 
 void AudioMixerSlaveThread::notify(bool stopping) {
-    {
-        Lock lock(_pool._mutex);
-        assert(_pool._numFinished < _pool._numThreads);
-        ++_pool._numFinished;
-        if (stopping) {
-            ++_pool._numStopped;
-        }
+    Lock lock(_pool._mutex);
+    assert(_pool._numFinished < _pool._numThreads);
+    ++_pool._numFinished;
+    if (stopping) {
+        ++_pool._numStopped;
     }
-    _pool._poolCondition.notify_one();
+
+    if(_pool._numFinished == _pool._numThreads) {
+        _pool._poolCondition.notify_one();
+    }
 }
 
 bool AudioMixerSlaveThread::try_pop(SharedNodePointer& node) {
--- a/assignment-client/src/audio/AudioMixerSlavePool.cpp
+++ b/assignment-client/src/audio/AudioMixerSlavePool.cpp
@@ -38,8 +38,8 @@ void AudioMixerSlaveThread::run() {
 
 void AudioMixerSlaveThread::wait() {
     {
-        Lock lock(_pool._mutex);
-        _pool._slaveCondition.wait(lock, [&] {
+        Lock slaveLock(_pool._slaveMutex);
+        _pool._slaveCondition.wait(slaveLock, [&] {
             assert(_pool._numStarted <= _pool._numThreads);
             return _pool._numStarted != _pool._numThreads;
         });
@@ -53,7 +53,7 @@ void AudioMixerSlaveThread::wait() {
 }
 
 void AudioMixerSlaveThread::notify(bool stopping) {
-    Lock lock(_pool._mutex);
+    Lock poolLock(_pool._poolMutex);
     assert(_pool._numFinished < _pool._numThreads);
     ++_pool._numFinished;
     if (stopping) {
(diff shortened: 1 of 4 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-Interface-PR1358-f8f8c27-_e8398578.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-f8f8c27-_e8398578.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-f8f8c27-.tar_e8398578.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-f8f8c27-.tar_e8398578.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-f8f8c27-_e8398578.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-f8f8c27-_e8398578.exe

odysseus654:
Just made another tweak / potential bugfix to the code: when resizing the number of slave threads downward it would flag the proper number of slaves to shutdown, but only wake a random number of those slaves to check their _stop flag and while() loop until things "work out".  The code should be requesting *all* the slave threads to check their _stop flag.

assignment-client/src/audio/AudioMixerSlavePool.cpp
```diff
@@ -53,15 +53,16 @@ void AudioMixerSlaveThread::wait() {
...
-        }
+    Lock poolLock(_pool._poolMutex);
+    assert(_pool._numFinished < _pool._numThreads);
+    ++_pool._numFinished;
+    if (stopping) {
+        ++_pool._numStopped;
+    }
+
+    if(_pool._numFinished == _pool._numThreads) {
+        _pool._poolCondition.notify_one();
```
HifiExperiments, line 64: I’m not familiar with this code so I’m still parsing all of this, but are we sure this change is safe?  If we kick off a bunch of workers, when one finishes, shouldn’t it immediately notify the pool to start another one?  with this change, aren’t we waiting for them all to finish before starting new ones?

odysseus654: if you look at the _poolCondition.wait calls, they all use a precondition of "_pool._numFinished == _pool._numThreads".  The idea is the pool wakes up *all* the current slaves, they do their work, then the pool wakes back up when they're finished (and go back to sleep again).  Without this "if" statement the pool wait will have a spurious wake every time each of the slaves finishes its work (but won't leave the wait() call).

assignment-client/src/avatars/AvatarMixerSlavePool.h
```diff
@@ -95,19 +95,20 @@ class AvatarMixerSlavePool {
...
     // Set from Domain Settings:
     float _priorityReservedFraction { 0.4f };
     int _numThreads { 0 };
 
-    int _numStarted { 0 }; // guarded by _mutex
-    int _numFinished { 0 }; // guarded by _mutex
-    int _numStopped { 0 }; // guarded by _mutex
+    int _numStarted { 0 }; // guarded by _slaveMutex
+    int _numFinished { 0 }; // guarded by _poolMutex
+    int _numStopped { 0 }; // guarded by _poolMutex
```
HifiExperiments, line 111: for a separate PR: I haven’t looked at the rest of these classes too much but could we DRY up this code?  and I’ve suggested this before but there are lots of alternatives to the whole “slave”/“master” metaphor that are less gross, I’d love if we did a cleanup pass on that

assignment-client/src/audio/AudioMixerSlavePool.cpp
HifiExperiments: that seems weird…shouldn’t we only be waking up workers as needed like in a normal thread pool?  why do we do them all at once?  isn’t it possible some of them won’t have anything to do?

assignment-client/src/avatars/AvatarMixerSlavePool.h
odysseus654: Suggestions welcome I guess, we have a single "pool" object that sets up a number of properties before flagging a number of threads to wake up, do a unit of work, and fall asleep again.  The code between the two files is similar enough that we could refactor it out, albeit likely into template classes.

Once we get farther on the PR review (or actually pick which one we'd like to go with) I see no problem with renaming variables (and class names)

assignment-client/src/audio/AudioMixerSlavePool.cpp
odysseus654: The goal is to consume the contents of the queue as quickly as possible, if the queue is empty then the (worker thread?) flags as "finished" without doing anything.

I'm unsure why we aren't using a threadpool on this (other than "spawn 100 tasks" might be a bit weird?).  The only thing I can think of is that this is optimized the heck out to be as fast as possible with as little task setup/shutdown as possible.

HifiExperiments: ok thank you that makes sense!

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-Interface-PR1358-9da4d76-_5b88846b.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-9da4d76-.tar_5b88846b.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-9da4d76-.tar_5b88846b.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-9da4d76-_5b88846b.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-9da4d76-_5b88846b.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1358/Vircadia-PR1358-9da4d76-_5b88846b.exe

odysseus654:
We ended up going with "Door # 2" which resolved this same issue, so closing this

Issue #1392, opened 2021-10-07
Missing Parameters?


Penguin-Guru:
AppImage --help lists a parameter "--help-all". I tried running that but it just ran the client. I tried running with both "--help" and "--help-all", but output was the same as from "--help". I have also read people discussing other parameters that are not mentioned in the --help output and they did not seem interested in adding to it. I think it's important that all functional parameters be output here, and that all parameters behave as expected. If there are simply too many to output, that's the use of parameters like "--help-all".

ctrlaltdavid:
From issue https://github.com/vircadia/vircadia/issues/827 : "Note that some command line options work with either a space or an = between the parameter and the value, and some only work with one or the other."

See also: https://github.com/vircadia/vircadia/issues/446 , https://github.com/vircadia/vircadia/issues/835

Issue #835, opened 2020-10-22
Regularize command line option handling such that can have "=" or " " between option and value


ctrlaltdavid:

daleglass:
Related to #827

digisomni:
This hasn't been solved yet.

Issue #827, opened 2020-10-20
`--concurrent-downloads` command line option is invalid


JulianGro:
When running Interface with `--concurrent-downloads` it returns `Unknown option 'concurrent-downloads'.`

ctrlaltdavid:
What is the exact command line you're using?
Note that some command line options work with either a space or an `=` between the parameter and the value, and some only work with one or the other. https://github.com/kasenvr/project-athena/issues/446

kasenvr:
I just ran my Interface with that option and it ran... though my stats still show "0/16" downloads indicating my limit is still 16 when I set it to 24, or at least I thought I did.

There are other arguments that use the same method in the code, you can try `--avatarURL` and see if that works (or not) for you?

JulianGro:
> What is the exact command line you're using?
> Note that some command line options work with either a space or an `=` between the parameter and the value, and some only work with one or the other. #446

It doesn't work with `--concurrent-downloads=10` or `--concurrent-downloads 10`



> I just ran my Interface with that option and it ran... though my stats still show "0/16" downloads indicating my limit is still 16 when I set it to 24, or at least I thought I did.
> 
> There are other arguments that use the same method in the code, you can try `--avatarURL` and see if that works (or not) for you?

`Unknown option 'avatarURL'.`

kasenvr:
So something is wrong with the way those commands are processed on Linux, or in general potentially! There are about 9 references to it in Application.cpp alone.

JulianGro:
So apparently all the options that I have tried and didn't work, are not shown in `--help`.
```
motofckr9k@motofckr9k-mint-desktop:~/Vircadia/install_logging_improvements$ ./run_interface --help
Usage: /home/motofckr9k/Vircadia/install_logging_improvements/interface/interface [options]
Vircadia

Options:
  -v, --version                            Displays version information.
  -h, --help                               Displays this help.
  --url <value>
  --no-launcher                            Do not execute the launcher
  --no-updater                             Do not show auto-updater
  --checkMinSpec                           Check if machine meets minimum
                                           specifications
  --runServer                              Whether to run the server
  --serverContentPath <serverContentPath>  Where to find server content
  --cache <dir>                            set test cache <dir>
  --scripts <path>                         set scripts <path>
  --allowMultipleInstances                 Allow multiple instances to run
  --tokens <json>                          set response tokens <json>
  --displayName <string>                   set user display name <string>
[...]

PR #1428, opened 2021-10-25
Overhaul command-line parameter functionality.

Penguin-Guru:
I don't use any of these so I can't really tell whether or not they're working. If they aren't working, let me know how to test the parameter that's broken and I'll try to look into the handling. Parameter order and descriptions are based on the current documentation.

https://github.com/vircadia/vircadia/issues/1392

https://github.com/vircadia/vircadia/issues/827

https://github.com/vircadia/vircadia/issues/835

### DIFF
--- a/interface/src/Application.cpp
+++ b/interface/src/Application.cpp
@@ -992,9 +992,9 @@ Application::Application(
     _logger(new FileLogger(this)),
 #endif
     _previousSessionCrashed(setupEssentials(argc, argv, parser, runningMarkerExisted)),
-    _entitySimulation(new PhysicalEntitySimulation()),
-    _physicsEngine(new PhysicsEngine(Vectors::ZERO)),
-    _entityClipboard(new EntityTree()),
+    _entitySimulation(std::make_shared<PhysicalEntitySimulation>()),
+    _physicsEngine(std::make_shared<PhysicsEngine>(Vectors::ZERO)),
+    _entityClipboard(std::make_shared<EntityTree>()),
     _previousScriptLocation("LastScriptLocation", DESKTOP_LOCATION),
     _fieldOfView("fieldOfView", DEFAULT_FIELD_OF_VIEW_DEGREES),
     _hmdTabletScale("hmdTabletScale", DEFAULT_HMD_TABLET_SCALE_PERCENT),
@@ -2662,21 +2662,21 @@ QString Application::getUserAgent() {
     auto formatPluginName = [](QString name) -> QString { return name.trimmed().replace(" ", "-");  };
 
     // For each plugin, add to userAgent
-    auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
-    for (auto& dp : displayPlugins) {
+    const auto& displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+    for (const auto& dp : displayPlugins) {
         if (dp->isActive() && dp->isHmd()) {
             userAgent += " " + formatPluginName(dp->getName());
         }
     }
-    auto inputPlugins= PluginManager::getInstance()->getInputPlugins();
-    for (auto& ip : inputPlugins) {
+    const auto& inputPlugins = PluginManager::getInstance()->getInputPlugins();
+    for (const auto& ip : inputPlugins) {
         if (ip->isActive()) {
             userAgent += " " + formatPluginName(ip->getName());
         }
     }
     // for codecs, we include all of them, even if not active
-    auto codecPlugins = PluginManager::getInstance()->getCodecPlugins();
-    for (auto& cp : codecPlugins) {
+    const auto& codecPlugins = PluginManager::getInstance()->getCodecPlugins();
+    for (const auto& cp : codecPlugins) {
         userAgent += " " + formatPluginName(cp->getName());
     }
 
@@ -2743,7 +2743,7 @@ void Application::onAboutToQuit() {
         _firstRun.set(false);
     }
 
-    foreach(auto inputPlugin, PluginManager::getInstance()->getInputPlugins()) {
+    for(const auto& inputPlugin : PluginManager::getInstance()->getInputPlugins()) {
         if (inputPlugin->isActive()) {
             inputPlugin->deactivate();
         }
@@ -3125,14 +3125,14 @@ void Application::initializeGL() {
 }
 
 void Application::initializeDisplayPlugins() {
-    auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+    const auto& displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
     Setting::Handle<QString> activeDisplayPluginSetting{ ACTIVE_DISPLAY_PLUGIN_SETTING_NAME, displayPlugins.at(0)->getName() };
     auto lastActiveDisplayPluginName = activeDisplayPluginSetting.get();
 
     auto defaultDisplayPlugin = displayPlugins.at(0);
     // Once time initialization code
     DisplayPluginPointer targetDisplayPlugin;
-    foreach(auto displayPlugin, displayPlugins) {
+    for(const auto& displayPlugin : displayPlugins) {
         displayPlugin->setContext(_graphicsEngine.getGPUContext());
         if (displayPlugin->getName() == lastActiveDisplayPluginName) {
             targetDisplayPlugin = displayPlugin;
@@ -3382,7 +3382,7 @@ void Application::initializeUi() {
 
     // This will set up the input plugins UI
     _activeInputPlugins.clear();
-    foreach(auto inputPlugin, PluginManager::getInstance()->getInputPlugins()) {
+    for(const auto& inputPlugin : PluginManager::getInstance()->getInputPlugins()) {
         if (KeyboardMouseDevice::NAME == inputPlugin->getName()) {
             _keyboardMouseDevice = std::dynamic_pointer_cast<KeyboardMouseDevice>(inputPlugin);
         }
@@ -3432,7 +3432,7 @@ void Application::initializeUi() {
 #if !defined(DISABLE_QML)
     // Now that the menu is instantiated, ensure the display plugin menu is properly updated
     {
-        auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+        DisplayPluginList displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
         // first sort the plugins into groupings: standard, advanced, developer
         std::stable_sort(displayPlugins.begin(), displayPlugins.end(),
             [](const DisplayPluginPointer& a, const DisplayPluginPointer& b) -> bool { return a->getGrouping() < b->getGrouping(); });
@@ -4116,7 +4116,7 @@ std::map<QString, QString> Application::prepareServerlessDomainContents(QUrl dom
     nodeList->setPermissions(permissions);
 
     // FIXME: Lock the main tree and import directly into it.
-    EntityTreePointer tmpTree(new EntityTree());
+    EntityTreePointer tmpTree(std::make_shared<EntityTree>());
     tmpTree->setIsServerlessMode(true);
     tmpTree->createRootElement();
     auto myAvatar = getMyAvatar();
@@ -4451,7 +4451,7 @@ void Application::keyPressEvent(QKeyEvent* event) {
             case Qt::Key_7:
                 if (isControlOrCommand || isOption) {
                     unsigned int index = static_cast<unsigned int>(event->key() - Qt::Key_1);
-                    auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+                    const auto& displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
                     if (index < displayPlugins.size()) {
                         auto targetPlugin = displayPlugins.at(index);
                         QString targetName = targetPlugin->getName();
@@ -4506,7 +4506,7 @@ void Application::keyPressEvent(QKeyEvent* event) {
             case Qt::Key_7:
                 if (isControlOrCommand || isOption) {
                     unsigned int index = static_cast<unsigned int>(event->key() - Qt::Key_1);
-                    auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+                    const auto& displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
                     if (index < displayPlugins.size()) {
                         auto targetPlugin = displayPlugins.at(index);
                         QString targetName = targetPlugin->getName();
@@ -4679,8 +4679,8 @@ void Application::keyReleaseEvent(QKeyEvent* event) {
 }
 
 void Application::focusOutEvent(QFocusEvent* event) {
-    auto inputPlugins = PluginManager::getInstance()->getInputPlugins();
-    foreach(auto inputPlugin, inputPlugins) {
+    const auto& inputPlugins = PluginManager::getInstance()->getInputPlugins();
+    for(const auto& inputPlugin : inputPlugins) {
         if (inputPlugin->isActive()) {
             inputPlugin->pluginFocusOutEvent();
         }
@@ -5367,8 +5367,8 @@ void Application::idle() {
         PerformanceTimer perfTimer("pluginIdle");
         PerformanceWarning warn(showWarnings, "Application::idle()... pluginIdle()");
         getActiveDisplayPlugin()->idle();
-        auto inputPlugins = PluginManager::getInstance()->getInputPlugins();
-        foreach(auto inputPlugin, inputPlugins) {
+        const auto& inputPlugins = PluginManager::getInstance()->getInputPlugins();
+        for(const auto& inputPlugin : inputPlugins) {
             if (inputPlugin->isActive()) {
                 inputPlugin->idle();
             }
@@ -5511,8 +5511,8 @@ void Application::loadSettings(const QCommandLineParser& parser) {
 
     bool isFirstPerson = false;
     if (parser.isSet("no-launcher")) {
-        auto displayPlugins = pluginManager->getDisplayPlugins();
-        for (auto& plugin : displayPlugins) {
+        const auto& displayPlugins = pluginManager->getDisplayPlugins();
+        for (const auto& plugin : displayPlugins) {
             if (!plugin->isHmd()) {
                 if (auto action = menu->getActionForOption(plugin->getName())) {
                     action->setChecked(true);
@@ -5526,8 +5526,8 @@ void Application::loadSettings(const QCommandLineParser& parser) {
         if (_firstRun.get()) {
             // If this is our first run, and no preferred devices were set, default to
             // an HMD device if available.
-            auto displayPlugins = pluginManager->getDisplayPlugins();
-            for (auto& plugin : displayPlugins) {
+            const auto& displayPlugins = pluginManager->getDisplayPlugins();
+            for (const auto& plugin : displayPlugins) {
                 if (plugin->isHmd()) {
                     if (auto action = menu->getActionForOption(plugin->getName())) {
                         action->setChecked(true);
@@ -5564,8 +5564,8 @@ void Application::loadSettings(const QCommandLineParser& parser) {
     _myCamera.setMode((isFirstPerson) ? CAMERA_MODE_FIRST_PERSON_LOOK_AT : CAMERA_MODE_LOOK_AT);
     cameraMenuChanged();
 
-    auto inputs = pluginManager->getInputPlugins();
-    for (auto plugin : inputs) {
+    const auto& inputs = pluginManager->getInputPlugins();
+    for (const auto& plugin : inputs) {
         if (!plugin->isActive()) {
             plugin->activate();
         }
@@ -6387,7 +6387,7 @@ void Application::update(float deltaTime) {
         };
 
         InputPluginPointer keyboardMousePlugin;
-        for (auto inputPlugin : PluginManager::getInstance()->getInputPlugins()) {
+        for(const auto& inputPlugin : PluginManager::getInstance()->getInputPlugins()) {
             if (inputPlugin->getName() == KeyboardMouseDevice::NAME) {
                 keyboardMousePlugin = inputPlugin;
             } else if (inputPlugin->isActive()) {
@@ -8979,14 +8979,14 @@ void Application::updateDisplayMode() {
     }
 
     // Once time initialization code that depends on the UI being available
-    auto displayPlugins = getDisplayPlugins();
+    const auto& displayPlugins = getDisplayPlugins();
 
     // Default to the first item on the list, in case none of the menu items match
 
     DisplayPluginPointer newDisplayPlugin = displayPlugins.at(0);
     auto menu = getPrimaryMenu();
     if (menu) {
-        foreach(DisplayPluginPointer displayPlugin, PluginManager::getInstance()->getDisplayPlugins()) {
+        for (const auto& displayPlugin : PluginManager::getInstance()->getDisplayPlugins()) {
             QString name = displayPlugin->getName();
             QAction* action = menu->getActionForOption(name);
             // Menu might have been removed if the display plugin lost
@@ -9039,7 +9039,7 @@ void Application::setDisplayPlugin(DisplayPluginPointer newDisplayPlugin) {
         bool active = newDisplayPlugin->activate();
 
         if (!active) {
-            auto displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
+            const DisplayPluginList& displayPlugins = PluginManager::getInstance()->getDisplayPlugins();
 
             // If the new plugin fails to activate, fallback to last display
             qWarning() << "Failed to activate display: " << newDisplayPlugin->getName();
@@ -9380,7 +9380,7 @@ void Application::unresponsiveApplication() {
 
 void Application::setActiveDisplayPlugin(const QString& pluginName) {
     DisplayPluginPointer newDisplayPlugin;
-    for (DisplayPluginPointer displayPlugin : PluginManager::getInstance()->getDisplayPlugins()) {
+    for (const DisplayPluginPointer& displayPlugin : PluginManager::getInstance()->getDisplayPlugins()) {
         QString name = displayPlugin->getName();
         if (pluginName == name) {
             newDisplayPlugin = displayPlugin;
--- a/interface/src/main.cpp
+++ b/interface/src/main.cpp
@@ -8,8 +8,6 @@
 //  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
 //
 
-#include <thread>
-
 #include <QCommandLineParser>
 #include <QtCore/QProcess>
 #include <QDebug>
--- a/.github/workflows/master_build.yml
+++ b/.github/workflows/master_build.yml
@@ -1,6 +1,6 @@
 name: Master CI Build
 
-on: 
+on:
   push:
     branches:
       - master
@@ -20,7 +20,9 @@ env:
 
   # OSX-specific variables
   DEVELOPER_DIR: /Applications/Xcode_11.2.app/Contents/Developer
-  MACOSX_DEPLOYMENT_TARGET: '10.11'
+  # We are using macOS SDK 10.12 to work around a lot of OpenGL issues in newer SDKs.
+  # There are crashes on startup, crashes on window resize, etc. on newer macOS SDKs.
+  MACOSX_DEPLOYMENT_TARGET: '10.12'
 
   # WIN-specific variables
   PreferredToolArchitecture: X64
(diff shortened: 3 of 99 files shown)


interface/src/main.cpp
```diff
@@ -66,59 +66,215 @@ int main(int argc, const char* argv[]) {
...
+    QCommandLineOption checkMinSpecOption(
+        "checkMinSpec",
+        "Check if machine meets minimum specifications. The program will run if check passes."
+    );
+    QCommandLineOption runServerOption(
+        "runServer",
+        "Run the server."
+    );
+    QCommandLineOption listenPortOption(
+        "listenPort",
```
daleglass, line 97: This is missing a value name, so the option is described as:
```
  --listenPort                             Port to listen on.
```

Note the missing value name, eg:
```
  --url <string>                           Start at specified URL location
```

```

File: interface/src/main.cpp [hunk 2]
```diff
@@ -66,59 +66,215 @@ int main(int argc, const char* argv[]) {
...
+    QCommandLineOption disableWatchdogOption(
+        "disableWatchdog",
+        "Disable the watchdog thread. The interface will crash on deadlocks."
+    );
+    QCommandLineOption systemCursorOption(
+        "system-cursor",
+        "Probably prevents changing the cursor when application has focus."
+    );
+    QCommandLineOption concurrentDownloadsOption(
+        "concurrent-downloads",
```
daleglass, line 155: Missing value name here too

interface/src/main.cpp
```diff
@@ -66,59 +66,215 @@ int main(int argc, const char* argv[]) {
...
+        "Maximum concurrent resource downloads. Default is 16, except for Android where it is 4."
+    );
+    QCommandLineOption avatarURLOption(
+        "avatarURL",
```
daleglass, line 159: Missing value name here too

interface/src/main.cpp
```diff
@@ -66,59 +66,215 @@ int main(int argc, const char* argv[]) {
...
+        "Override the avatar U.R.L."
+    );
+    QCommandLineOption replaceAvatarURLOption(
+        "replace-avatar-url",
```
daleglass, line 163: Missing value name here too

daleglass (changes requested):
I like this a lot! Just a few minor quibbles so far.

interface/src/main.cpp
Penguin-Guru: I don't really understand why there are two avatar url parameters but I assume they both accept url strings.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-9dd76be-.tar_a1a9705d.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-9dd76be-.tar_a1a9705d.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-9dd76be-_a1a9705d.sh

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-9dd76be-_a1a9705d.dmg

interface/src/main.cpp
daleglass: I looked and have no idea why either. Probably something that made sense for HiFi? Like there's one place that may be using the avatar as an argument, and another that needs to be able to override that. Almost certainly not important for us until we develop such a need.

interface/src/Application.h
```diff
@@ -505,7 +510,7 @@ private slots:
 
     void notifyPacketVersionMismatch();
 
-    void loadSettings();
+    void loadSettings(const QCommandLineParser* parser);
```
daleglass, line 512: Why a pointer here and not a reference?

daleglass (changes requested):

interface/src/Application.h
Penguin-Guru: Personal prejudice. I read about references once but they just seemed like worse pointers to me. :p

Practically speaking, the only reason I can think of is that you might want to delete the parser to free up memory once the program is initialised. I didn't do that because I wasn't sure if you'd want to. Seems objectively better in the practical sense but maybe a stylistic choice. If you do want to free up the space, it would be good to confirm that the parser isn't being referenced after the application is fully initialised (not just constructed)-- it probably shouldn't be and if it is, that should be trivial to fix. You're welcome to change the pointer to a reference if you prefer though.

daleglass: Sorry for the late reply, got sick, bunch of stuff got out of whack.

A reference is better in this context. A reference can't be a null pointer without invoking undefined behavior, at which point all hope is lost anyway. So if the function is operating under the assumption that `parser` will be set to something useful, a reference is a better fit.

Also, I'm not sure what's the relationship to deletion there. You can't delete something passed with a reference without taking its address -- that's not considered a good style. The general idea is that if you get a reference passed in something else has ownership of the data, and you're not supposed to mess with it.

In any case, deallocation wouldn't be something it'd be appropriate for `loadSettings` to perform.

Penguin-Guru: It's ok.

I still don't really understand the value of references (not a pun) but you're welcome to use a reference if that's what you prefer. My pull requests are basically just suggestions. 👍

daleglass: The value is basically that it's a pointer with added restrictions that you want to exist a lot of the time. If you get a pointer it could be `nullptr`, which means now you have to figure out what to do about that. Do you check for it? Could null be a valid value? What if you get one, do you return a special value, do you silently ignore it, do you skip something, do you crash the program? It's potentially troublesome, you may need to reorder code so that any checks/cleanup are correctly ordered, which could get messed up later.

If you're guaranteed to have a valid object as an input it removes all that from consideration.

Penguin-Guru: So what you're saying is that passing a reference is the same as `if (parser == nullptr) return;`, just with a different format? I assume it would still have to be tested at run-time.

I don't think it's possible for parser to be nullptr in this case but I could be wrong.

daleglass: No, you can't normally create a reference to nullptr. The compiler just won't allow it. To create a reference you have to take the address of an already existing, valid object.

The one way to do it anyway is to dereference a null pointer, but that is in itself undefined behavior. And at that point you can consider the program irrevocably broken.

Penguin-Guru: Right, so the test `if (parser == nullptr)` must occur at run-time regardless of whether a pointer or reference is used. Testing it explicitly is just more flexible, since you could return from the function, run a quit routine, or whatever is appropriate. It's also more clear to read for programmers who aren't very familiar with C++. If using references eliminates the run-time test then I would prefer it. Do you know?

daleglass: There is no test. You can't create a reference to nullptr with valid code.

https://godbolt.org/z/7c6cv8M5M

Try changing it to: `iTakeAReference(nullptr);` -- it'll simply fail to compile.

Penguin-Guru: No, I mean the that the reference is probably tested implicitly when the function is called to make sure it is not a nullptr. It might be possible for compilers to discern in some cases that the value could not possibly be nullptr at compile time, I'm not sure if they do. Or maybe pointers can not be passed as references in general?

daleglass: There is no runtime test, no. The language's syntax prevents a reference to a nullptr from being created in the first place. Try playing with the linked program.

Penguin-Guru: I am. It seems like pointers can be passed by reference, so the program must check to make sure the pointers are not null. Maybe I'm misunderstanding something. 🤔
https://godbolt.org/z/qxeq8E61Y

Penguin-Guru: Oh, you probably mean that it's not possible for the value unless it is a reference to a pointer. That makes sense.

Penguin-Guru: I'm still not sure how I feel about references in general but I will try to change this for you since you went to the trouble of explaining references for me. Hopefully nothing gets broken. My ability to test these parameters is limited.

Penguin-Guru: Is this how you wanted it? https://github.com/vircadia/vircadia/pull/1428/commits/46d2ff0d0774c0ac7475bc6a40c2da5eac100e94

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-4c0bca7-_87768cc3.dmg

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4c0bca7-_87768cc3.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4c0bca7-.tar_87768cc3.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4c0bca7-_87768cc3.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4c0bca7-.tar_87768cc3.Z

interface/src/Application.h
namark: Sorry for jumping in, but I'm a total c++ fanboy.
Yes the point is to replace the pointer to the value with a reference to the value. There are no runtime checks because it's less of an absolute safety guarantee and more of a clarification for the programmers to prevent mistakes. If you have a function that accepts a pointer there is always a question of who's going to be handling null, the implementer or the caller, assuming they are two different people (or same person at different times). The implementer is like "ok do I check it and return an error or something? But what if the caller can guarantee it is not null and doesn't want the overhead of the check?" and the caller is like "ok do I check it? But what if the implementer already checked it and then I'll be checking for no reason". Sometimes it can all work out, other times you end up with either too many or not enough null checks. 
 The reference settles the matter - if the function accepts a reference to a value it will not handle null (as you are not allowed to pass it null, which is a consequence of the language rules, which the compiler will check if/when it can, but otherwise it's up to the programmers to adhere to in general), otherwise if it accepts a pointer it promises to handle null somehow so it's ok to pass it. So as a rule of thumb if your logic requires a value, you accept a reference instead of something like top level null check that just returns, and you use a pointer if the parameter is optional and checking it is part of your logic.

Penguin-Guru: That makes a lot of sense. It feels like a kind of heavy handed way to solve such a small problem but the problem does come up a lot and maybe syntax is a good way to fix it. Could they have added a pointer-specific constraint "nonull" like "const"? That feels more intuitive to me but also takes up a lot more space per line. What do you think?

namark: The reference is actually a big part of the language, and it serves a lot of other purposes as well. This is just one of the use cases, and not the sole point of the design. Some other differences from pointers that come to mind:
A reference is completely transparent alias for the value, you can't get a handle on it in any way, it has no address, (if you try to take an address you get the address of the value it refers to), so you can't have a pointer to a reference, or a reference to a reference.
There is a special magic about const references in particular - they can bind to temporaries and extend their lifetime. 
Since c++11 there are rvalue references that allow even greater control over temporaries. 

I don't know the actual history of it, but in general c++ is very big on value semantics and that's a plausible explanation. If you asked a c++ programmer what's would be a perfect world to them, they would likely say "a world where everything behaves like a value", and a lot of the language's design is about achieving that dream. From that point of view references are an attempt to bring value semantics to the world of indirection.

Penguin-Guru: Interesting. I wish the one C++ book I read had introduced them that well. :p

I would say that I'm a fan of pointers in general but that's not to say they're perfect. I won't ask any more questions here because I don't want to spam Dale and the Vircadia Discord's GitHub channel too much, but I will plan on looking into references a bit more sometime. Thanks for jumping in.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-42c252c-_46d2ff0d.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-_46d2ff0d.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-.tar_46d2ff0d.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-.tar_46d2ff0d.gz

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-_46d2ff0d.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-_46d2ff0d.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-_46d2ff0d.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-.tar_46d2ff0d.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-42c252c-.tar_46d2ff0d.Z

interface/src/Application.h
Penguin-Guru: Since Dale doesn't seem to mind, do you know any good books to recommend?

interface/src/main.cpp
```diff
@@ -66,59 +66,219 @@ int main(int argc, const char* argv[]) {
...
+    );
+    // The documented "--disable-lod" does not seem to exist.
+    // Below are undocumented.
+    QCommandLineOption noLauncherOption(
+       "no-launcher",
+       "Do not execute the launcher."
+    );
+    QCommandLineOption overrideScriptsPathOption(
+       "overrideScriptsPath",
+       "Probably specifies where to look for scripts.",
```
daleglass, line 187: Remove "probably". This is the path from which scripts are loaded by default. The default is the scripts directory that can be found in the root of the source tree.

interface/src/main.cpp
```diff
@@ -66,59 +66,219 @@ int main(int argc, const char* argv[]) {
 
     setupHifiApplication(BuildInfo::INTERFACE_NAME);
 
-    QStringList arguments;
-    for (int i = 0; i < argc; ++i) {
-        arguments << argv[i];
-    }
-
     QCommandLineParser parser;
     parser.setApplicationDescription("Vircadia");
```
daleglass, line 70: I know you didn't change it, but while we're at it, this could use being a bit more informative. As it is, we're getting:

```
Usage: interface/interface [options]
Vircadia

Options:
  -h, --help                               Displays help on commandline
                                           options.
```

daleglass (changes requested):
Thanks for the improvements :)

I've been testing, and there's a number of 'probably' that shouldn't be there, so I checked the code to see if that could be improved.

interface/src/main.cpp
Penguin-Guru: Can you clarify what "where scripts are loaded by default" means? How is this different from the other script-related parameters?

daleglass: This is the directory where the code will look for scripts.

`--defaultScriptsOverride` is a filename, it's used to replace `defaultsScripts.js`. That's the script that will be loaded automatically on start, and which is used to set up the UI and other basic functionality.

`--scripts` seems broken from the looks of it, and may not be doing anything useful

interface/src/main.cpp
```diff
@@ -66,59 +66,219 @@ int main(int argc, const char* argv[]) {
...
+        "standalone",
+        "Emulate a standalone device."
+    );
+    QCommandLineOption disableWatchdogOption(
+        "disableWatchdog",
+        "Disable the watchdog thread. The interface will crash on deadlocks."
+    );
+    QCommandLineOption systemCursorOption(
+        "system-cursor",
+        "Probably prevents changing the cursor when application has focus."
```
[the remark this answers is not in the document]
Penguin-Guru: I have attempted to fix this. Please let me know if it is resolved after I push the commit.

interface/src/main.cpp
```diff
@@ -66,59 +66,219 @@ int main(int argc, const char* argv[]) {
...
+        "serverContentPath"
+    ); // This data type will not be familiar to users.
+    QCommandLineOption overrideAppLocalDataPathOption(
+        "cache",
+        "Set test cache.",
+        "dir"
+    );
+    QCommandLineOption scriptsOption(
+        "scripts",
+        "Set path for defaultScripts. These are probably scripts that run automatically.",
```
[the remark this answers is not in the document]
Penguin-Guru: Should we try to fix it or comment out the parameters? This one is actually in the docs but I can't tell what it's supposed to do from the description.
https://docs.vircadia.dev/interfaces/native/command-line-parameters.html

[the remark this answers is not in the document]
daleglass: It should be an easy fix. The main issue is that it doesn't actually pass the path to `scriptEngines`.

There's also the `loadDefaultScripts` vs `loadScripts` difference that is suspicious.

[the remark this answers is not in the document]
Penguin-Guru: These both call loadScript(), which does not seem to support directories. Maybe it does somehow.

loadScripts reads from the settings before handling "_defaultScriptsOverride". I think this parameter does something different but I can't figure it out right now. I'm going to push the other fixes for you to test once I've made sure the program runs in general.
https://github.com/vircadia/vircadia/blob/master/libraries/script-engine/src/ScriptEngines.cpp#L343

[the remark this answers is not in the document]
Penguin-Guru: Actually, shouldn't this already be assigned here?
https://github.com/Penguin-Guru/vircadia/blob/FixingParameters/interface/src/Application.cpp#L823

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-.tar_37e1f75f.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-.tar_37e1f75f.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-_37e1f75f.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-45808df-_37e1f75f.dmg

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-_37e1f75f.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-_37e1f75f.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-_37e1f75f.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-.tar_37e1f75f.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-45808df-.tar_37e1f75f.gz

interface/src/Application.h
namark: It was all courses, talks and articles for me, along with a lot of practice, so can never remember or find something specific to recommend ._. I would say there is a general lack of good beginner materials, so it's normal to just stick to one or two things that you like and then slowly absorb the rest as you go.

Penguin-Guru: Ok. I know it's hard to keep books up to date but I am specifically in need of books to read. I don't mind if they're advanced as long as they're well organised, accurate, and complete in their topics, Let me know if you ever run into any. :)

interface/src/main.cpp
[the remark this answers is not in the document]
daleglass: No, that's a different option. That deals with `--defaultScriptsOverride`. It's `--scripts` that seems broken.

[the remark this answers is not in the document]
Penguin-Guru: Ok. I'm having a really hard time telling these features apart. The variable names overlap a lot and I'm not sure how the parameters are supposed to be different from each other.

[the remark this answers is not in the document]
Penguin-Guru: `loadScripts()` does not accept a parameter, and it contains code that handles `--defaultScriptsOverride` in a way that completely overlaps with the specified functionality of `--scripts`. The variable names for `--scripts` suggest it serves the exact same purpose. Perhaps is was only partially implemented, intended to replace `--defaultScriptsOverride` and mirror the name of the settings value? But `--scripts` is in the web documentation and `--defaultScriptsOverride` is not.

interface/src/Application.h
namark: Well http://elementsofprogramming.com/eop.pdf is on my must-read list. It's a bit dense and abstract though.

interface/src/Application.cpp
```diff
@@ -5814,10 +5816,10 @@ void Application::resumeAfterLoginDialogActionTaken() {
 
         // if the --scripts command-line argument was used.
         if (_overrideDefaultScriptsLocation && _defaultScriptsLocation.exists()) {
-            scriptEngines->loadDefaultScripts();
+            scriptEngines->loadScripts();
```
daleglass, line 5819: Wait, why did you swap this around?

My comment was that the fact that different functions get called in different branches here is weird. You'd expect that an option that changes a path just changes a path, not results in different behaviors being implemented.

That is, what I would have expected the code to look like here is something along these lines:

```
if (_overrideDefaultScriptsLocation && _defaultScriptsLocation.exists()) {
    scriptEngines->setPath(_defaultScriptsLocation);
}
scriptEngines->loadScripts();
```

Looking at the code:

`scriptEngines->loadDefaultScripts();` -- this works out to `loadScript(DEFAULT_SCRIPTS_LOCATION);`, where DEFAULT_SCRIPTS_LOCATION is a constant.

`scriptEngines->loadScripts()` loads a list of scripts from what's listed in the config file.

I think it was better the way it was before: the default path is to load whatever the user had decided.

daleglass (changes requested):

interface/src/Application.cpp
daleglass: At this point I think half this stuff is probably useless to us, and a whole bunch of it is confusingly named.

I think the important part here is not to break anything important to us, and then do a bit of further cleanup in a future PR where we figure out better names for some options and maybe outright remove some of them.

interface/src/Application.h
Penguin-Guru: Ok, thanks! I'm not very good at math but this seems like a useful perspective on programming. I'll give it a shot. :)

ksuprynowicz: It looks like _Embracing Modern C++ Safely_ might be interesting. I will order it as soon as it's available because I like Vittorio Romeo's work on open source projects.

Penguin-Guru: This one is kind of expensive but let me know what you think of it once it arrives. If you like it then I will buy a copy too.

interface/src/Application.cpp
Penguin-Guru: Ok. I agree. There are probably other parameters that would be useful and I'm sure someone will want to change some of the descriptions.

I will swap the functions back.

Penguin-Guru: The code you suggested seems much better to me but I don't want to break anything that might get confused with other changes in this pull request. Reconciling these script parameters could be another.

daleglass: No, I was just explaining why I made the original comment. I wasn't saying "swap this around", but "this looks kinda weird".

On a second thought, I think I get the reason why it was done this way now. If you override the path, then we only load the default, hardcoded scripts because that's all we can expect to be present at whatever path the user passed in.

If we're not overriding anything, then we have the normal function where the interface remembers what scripts the user wanted in its config file.

So I think the right way is just leaving it how it was before

Penguin-Guru: Ok. I think we should open an issue to reconcile the script parameters in a different PR. It would be good to reconsider them all, including their names, descriptions, and desired functionality.

daleglass (approved):

interface/src/main.cpp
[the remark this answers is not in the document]
Penguin-Guru: If you have confirmed that `--scripts` does not work, should I comment it out for now or make it a pseudonym for `--defaultScriptsOverride`? If we're commenting it out, it should probably also be removed from the web docs.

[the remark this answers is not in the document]
daleglass: I tried it, and as far as I can tell, it's not doing anything. I'll create an issue.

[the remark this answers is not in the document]
Penguin-Guru: Ok. I will add a warning to the `--help` output for now but I suggest commenting it out of the code and removing it from the web docs if no other resolution is determined before this PR is merged. The main goal of this PR is to fix such inconsistencies and I think this is the only one left (hopefully).

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-a653b66-_8dfe17ee.dmg

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-5339ae3-_22cabbc2.dmg

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-_22cabbc2.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-.tar_22cabbc2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-.tar_22cabbc2.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-_22cabbc2.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-_22cabbc2.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-_22cabbc2.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-.tar_22cabbc2.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-5339ae3-.tar_22cabbc2.gz

Penguin-Guru:
@daleglass Should be up to date with master now. I can't test this right now but I assume it will work since there were no conflicts.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-4d8b276-_3a2646e5.dmg

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-_3a2646e5.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-.tar_3a2646e5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-_3a2646e5.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-.tar_3a2646e5.Z

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-_3a2646e5.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-.tar_3a2646e5.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-.tar_3a2646e5.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-4d8b276-_3a2646e5.sh

digisomni:
Needs a merge from master to fix merge conflict.

Penguin-Guru:
Go for it.

vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-.tar_ca98b81a.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-.tar_ca98b81a.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-_ca98b81a.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-Interface-PR1428-44c5373-_ca98b81a.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-_ca98b81a.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-_ca98b81a.sh
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-.tar_ca98b81a.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1428/Vircadia-PR1428-44c5373-.tar_ca98b81a.gz

digisomni:
Tested with the Pantheon Launcher on Windows, it successfully launched however my OpenVR disabling parameters did not work correctly and it launched with SteamVR anyway.

digisomni:
@namark can you see about testing out the command lines apply the necessary fixes/revisions?

PR #1514, opened 2021-12-04
MIDI doesn't compile?

odysseus654:
I just tried to merge master into #1200 and got a compile error... because midi.cpp doesn't have access to mmeapi.h?

I'm unsure if this is related to #1200 and so am including it here for independent evaluation.

### DIFF
--- a/libraries/midi/src/Midi.cpp
+++ b/libraries/midi/src/Midi.cpp
@@ -16,10 +16,9 @@
 #include <QtCore/QLoggingCategory>
 
 #if defined Q_OS_WIN32
-#include "Windows.h"
-#endif
+#include <Windows.h>
+#include <mmeapi.h>
 
-#if defined Q_OS_WIN32
 const int MIDI_BYTE_MASK = 0x0FF;
 const int MIDI_NIBBLE_MASK = 0x00F;
 const int MIDI_PITCH_BEND_MASK = 0x3F80;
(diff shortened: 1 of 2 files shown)


odysseus654:
If this is working for everyone else then I'm tempted to just dump this into #1200 and be done with it.

digisomni:
If it doesn't happen for any other people/PRs, just go ahead and apply it with #1200 then.

digisomni:
Shall we close this for now, then?

libraries/midi/src/Midi.cpp
```diff
@@ -16,10 +16,9 @@
 #include <QtCore/QLoggingCategory>
 
 #if defined Q_OS_WIN32
-#include "Windows.h"
-#endif
```
daleglass, line 20: Why was this `#endif` removed? This seems to make the entire file a no-op on non-windows systems.

daleglass (changes requested):

libraries/midi/src/Midi.cpp
odysseus654: The next line below this was an #if defined Q_OS_WIN32, seemed kind of useless added complexity

odysseus654:
I appreciate the code review, whether this gets merged or not as-is... I'd like to know why it doesn't compile for #1200 now if possible.  Not the highest priority at the moment as I'm finding myself lobotomizing a lot more of QtScript than I had originally planned to, will likely circle back to this after that code starts working (again and/or better).

PR #1525, opened 2021-12-09
Switch to different Qt build on macOS

JulianGro:
This PR aims to make Vircadia run on Big Sur *with* the old SDK.
Since I haven't been able to create a proper package, I am testing this here.


JulianGro:
Unfortunately this doesn't run on Big Sur either.

PR #1556, opened 2021-12-28
Fix static function warning in SocketType.h

JulianGro:
This PR fixes a compiler warning. It should not change behaviour at all.

Specifically it fixes all the
```
SocketType.h:31:16: warning: ‘QString socketTypeToString(SocketType)’ defined but not used [-Wunused-function]
```

PR #1558, opened 2021-12-31
Use Qt 5.12.3 to work around macOS issues

### LINKED ISSUE #1516: 2022.1.0-Selene will not launch on macOS 11.6 "Big Sur"
The 2022.1.0-RC launches successfully on macOS 11.6. https://athena-public.s3-eu-west-3.amazonaws.com/master/393/Vircadia-393-aabcdea-.dmg

Production release of the Interface for macOS will not complete the launch process. This implies a difference in the build process. We should use the same build process for RC and Production.

JulianGro:
This PR works around the issues we are experiencing when mixing macOSXSDK10.12 with Qt 5.15.2.

Fixes #1516

Depends on PR #1556

### DIFF
--- a/assignment-client/src/audio/AvatarAudioStream.h
+++ b/assignment-client/src/audio/AvatarAudioStream.h
@@ -23,7 +23,12 @@ public:
     AvatarAudioStream(bool isStereo, int numStaticJitterFrames = -1);
 
 private:
+#if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
+    AvatarAudioStream(const AvatarAudioStream&);
+    AvatarAudioStream& operator= (const AvatarAudioStream&);
+#else
     Q_DISABLE_COPY(AvatarAudioStream)
+#endif
 
     int parseStreamProperties(PacketType type, const QByteArray& packetAfterSeqNum, int& numAudioSamples) override;
 };
--- a/domain-server/src/NodeConnectionData.cpp
+++ b/domain-server/src/NodeConnectionData.cpp
@@ -51,12 +51,12 @@ NodeConnectionData NodeConnectionData::fromDataStream(QDataStream& dataStream, c
 
     dataStream >> newHeader.lastPingTimestamp;
     
-    SocketType publicSocketType, localSocketType;
+    quint8 publicSocketType, localSocketType;
     dataStream >> newHeader.nodeType
         >> publicSocketType >> newHeader.publicSockAddr >> localSocketType >> newHeader.localSockAddr
         >> newHeader.interestList >> newHeader.placeName;
-    newHeader.publicSockAddr.setType(publicSocketType);
-    newHeader.localSockAddr.setType(localSocketType);
+    newHeader.publicSockAddr.setType((SocketType)publicSocketType);
+    newHeader.localSockAddr.setType((SocketType)localSocketType);
 
     // For WebRTC connections, the user client's signaling channel WebSocket address is used instead of the actual data 
     // channel's address.
--- a/libraries/controllers/src/controllers/UserInputMapper.cpp
+++ b/libraries/controllers/src/controllers/UserInputMapper.cpp
@@ -738,7 +738,11 @@ Mapping::Pointer UserInputMapper::newMapping(const QString& mappingName) {
 //        if (request->getResult() == ResourceRequest::Success) {
 //            result = parseMapping(QString(request->getData()));
 //        } else {
+// #if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
+//            qCWarning(controllers) << "Failed to load mapping url <" << jsonUrl << ">" << endl;
+// #else
 //            qCWarning(controllers) << "Failed to load mapping url <" << jsonUrl << ">" << Qt::endl;
+// #endif
 //        }
 //        request->deleteLater();
 //    }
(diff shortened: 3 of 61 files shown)


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-Interface-PR1558-4eff071-_28c736e0.dmg

**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-_28c736e0.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-.tar_28c736e0.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-.tar_28c736e0.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-_28c736e0.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-.tar_28c736e0.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-.tar_28c736e0.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1558/Vircadia-PR1558-4eff071-_28c736e0.sh

Steve-Pruneau:
Oh wow! I am absolutely loving PR 1558 !!! This is the best macOS interface yet!

It is stable and movement smooth (not as much lag during loading).

This should go into production.

digisomni:
Maybe before each if statement we should have a "TODO" noting that it's a hack and may not really last long term as we have to move up in Qt versions over time which includes reworking more code.

JulianGro:
That would make the code even messier than it already is, and all occurrences can be found by the same if statement.
Adding something like "TODO: Remove once there is a Metal rendering backend" to ~120 if statements just seems a bit much.
Maybe someone that actually works on Qt deprecation like @daleglass could chime in.

JulianGro:
We actually also have if statements like this in the code already.
So if we decide on adding TODOs, then we should add them to those as well.

ArcadeFever:
Sorry for the late comment on this, but is the goal to downgrade all macOS builds or split out "Big Sur" from the mac builds to resolve Steve's issue?  If this PR is doing the former, then it will break web entity support for zoom/google docs/ etc, which currently work fine with newer macOS versions and Selene.

JulianGro:
I don't see how downgrading all builds would not break web entity support while a secondary build works.

Whatever the case, this works around all the OpenGL issues. If the current client works on your mac that is fine but that is an exception. Without this PR there is crashes on startup, crashes on resizing window, and freezes on startup on most Apple machines running macOS Big Sur or newer.

ArcadeFever:
So turns out I was running the release candidate for Selene on my MacBook, thus crossed out last part above.  Testing with the full 2022.1.0 release it is also failing to launch for me.  In fact, 2021.1.3 and 2021.1.2 also fail to launch, but 2021.1.1 works (and use the "interface" client before it was renamed to Vircadia).  

The concern with downgrading QT, is it will break web entity support for certain applications like Zoom and Google Docs, so would be a step backward, but I can understand not a lot of mac users probably need this support right now.

I would think we should be investigating why the Selene RC build and 2021.1.1 (and prior) worked fine with QT 5.15.2 on mac.  So I hope this PR fix to rollback QT is only temporary.

JulianGro:
> The concern with downgrading QT, is it will break web entity support for certain applications like Zoom and Google Docs, so would be a step backward, but I can understand not a lot of mac users probably need this support right now.

Will it break support or have you actually tested the PR?

> I would think we should be investigating why the Selene RC build and 2021.1.1 (and prior) worked fine with QT 5.15.2 on mac. So I hope this PR fix to rollback QT is only temporary.

I know why some builds worked and some didn't. Try fullscreen mode on that RC build you have been using.
That is a build that was made with a different macOSXSDK.
The macOSXSDK that we need to use to work around Apples OpenGL issues doesn't work properly together with newer Qt and macOS Big Sur or higher. And with properly I mean that it doesn't work at all for me and Steve.

This PR is to gain some time. If we don't get Metal support somehow, we will eventually not be able to run on macOS anymore.

assignment-client/src/audio/AvatarAudioStream.h
```diff
@@ -23,7 +23,12 @@ class AvatarAudioStream : public PositionalAudioStream {
     AvatarAudioStream(bool isStereo, int numStaticJitterFrames = -1);
 
 private:
+#if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
+    AvatarAudioStream(const AvatarAudioStream&);
+    AvatarAudioStream& operator= (const AvatarAudioStream&);
+#else
```
daleglass, line 29: I suggest adding a new header/source file, something like QtCompatibility.h, and implementing some of the functionality there. That'll result in a smaller and cleaner patch. For instance, something like this:

    #ifndef Q_DISABLE_COPY
        #define Q_DISABLE_COPY(className) \
            className(const className &) = delete;\
            className &operator=(const className &) = delete;
    #endif

And with just that, things get smaller and nicer.

domain-server/src/NodeConnectionData.cpp
```diff
@@ -51,12 +51,12 @@ NodeConnectionData NodeConnectionData::fromDataStream(QDataStream& dataStream, c
 
     dataStream >> newHeader.lastPingTimestamp;
     
-    SocketType publicSocketType, localSocketType;
+    quint8 publicSocketType, localSocketType;
```
daleglass, line 54: Why this change? SocketType isn't a Qt concept.

libraries/controllers/src/controllers/UserInputMapper.cpp
```diff
@@ -1177,13 +1181,22 @@ Mapping::Pointer UserInputMapper::parseMapping(const QString& json) {
     if (doc.isNull()) {
         qCDebug(controllers) << "Invalid JSON...\n";
         qCDebug(controllers) << error.errorString();
+#if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
+        qCDebug(controllers) << "JSON was:\n" << json << endl;
```
daleglass, line 1185: I think we could do something like:

     #if (QT_VERSION < QT_VERSION_CHECK(5, 14, 0))
         #define VEndl endl
     #else
         #define VEndl Qt::endl
     #endif

daleglass (changes requested):

domain-server/src/NodeConnectionData.cpp
ctrlaltdavid: This and related changes in commit `058b9032` fixes Qt 5.12 not serializing/deserializing to these variables using the `>>` and `<<` operators.

digisomni:
This PR needs to be updated from Vircadia master.

Issue #1578, opened 2022-01-24
unity-avatar-exporter path separators not compatible with non-windows systems


mrkcdf:
At least one path referenced in the unity-avatar-exporter uses \ as path separators which is fine for Windows, but not Linux or Mac. 

I will push a PR later today, but opened an issue so it can be referenced.

mrkcdf:
Merged to master. Issue closed https://github.com/vircadia/vircadia/pull/1579

PR #1579, opened 2022-01-24
Update Unity Avatar Exporter with fixed cross-platform support.

mrkcdf:
The previous version of the unity avatar exporter referenced paths with backslash which is only useful for Windows.  This PR changes paths to use forward slash which is supported by Windows, Mac, Linux.  

Updated absolute path checking to be conditional on OS family.

Reference changes from High Fidelity to Vircadia (excluding copyright notices)


mrkcdf:
Addresses #1578

mrkcdf:
Note to tester: This is still only tested up to Unity Editor 2018.2.12f1.  Any versions later than this is not guaranteed to work

SilverfishVR:
Tested on windows, unity 2017.4.15f1. (LTS)

on the initial export page the default path is a mix of forward and backwards slashes:
Screenshot 2022-01-24 23 01 11
exporting still works though so I guess it's OK

If I click browse and select a folder it looks normal:
Screenshot 2022-01-24 23 25 20
again, export works still.

Another issue I saw was that once you hit the export button it opens file explorer, I assume to make it easier to find the avatar but it fails to navigate to the correct folder.

FYI. it does not work on Unity 2018.4.20f1 (LTS) because the scale slider and input box is very broken and defaults to something like a gazillion percent.

mrkcdf:
The mixture of slashes I'll address in a newer version. The aim for this one was to get it working across platforms, though it could definitely do with some TLC going forward

digisomni (approved):

PR #1589, opened 2022-02-01
Fix dedicated IPFS gateway rejecting download requests.

namark:
There is a newer, higher level officially recommended python [requests library](https://docs.python-requests.org/en/latest/) that can take care of all this mess, but I'm not sure if it's ok to introduce it here as a dependency.

### DIFF
--- a/hifi_utils.py
+++ b/hifi_utils.py
@@ -103,16 +103,20 @@ def downloadFile(urls, hash=None, hasher=hashlib.sha512(), retries=3):
     for url in urls:
         for i in range(retries):
             tempFileName = None
-            # OSX Python doesn't support SSL, so we need to bypass it.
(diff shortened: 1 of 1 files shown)


digisomni (approved):

daleglass:
This one is weird, why is SSL not present in Python on OSX? Pretty much everything uses SSL these days, and many servers are SSL exclusive.

namark:
I'm not sure myself, that logic has been there since forever it seems, I just rearranged some things, but didn't change anything on that regard, just added the user agent header. It might not be an issue anymore or maybe another quirk of this low level API, pretty much nobody recommends using it anymore as far as I can tell.

daleglass (approved):

PR #1593, opened 2022-02-10
Added hash checks for downloaded items in python build scripts.

namark:


hifi_vcpkg.py
```diff
@@ -283,9 +289,8 @@ def setupAndroidDependencies(self):
             dest = os.path.join(self.path, 'installed')
             url = self.readVar('EXTERNAL_VCPKG_ANDROID_URLS').split(';')
             # FIXME I don't know why the hash check frequently fails here.  If you examine the file later it has the right hash
```
namark, line 291: This seems to be happening everywhere now in github actions, except the self-hosted_debian-11_aarch64 build.

namark: There was a bug that causes all retries to fail once a single hash check failed (or in some cases even unconditionally it seems). After fixing that the builds are succeeding. Otherwise an occasional singular failure should be fine and should be taken care of by retries and fallbacks. Leaving this FIXME comment in since this particular case has not been tested.

vircadia-build-notifier:
**build (macOS-10.15, full)**
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-_056dfb93.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-Interface-PR1593-22918bc-_056dfb93.dmg

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-.tar_056dfb93.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-.tar_056dfb93.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-_056dfb93.sh

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-_056dfb93.exe

 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-.tar_056dfb93.gz
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-.tar_056dfb93.Z
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1593/Vircadia-PR1593-22918bc-_056dfb93.sh

Colorfingers:
Looking over the commits... I don't see anything glaringly wrong with the changes... I am wondering, where are the hash values we are checking?  Don't you have to know them in advance to check against?  I don't see any logic to verify hashes... or am I missing the purpose here?

namark:
The hash values are coming form here
https://github.com/namark/vircadia/blob/056dfb936328c486308c01e474fec310969f475e/cmake/ExternalAssetsConfig.cmake
where they are manually specified and relayed to the python scripts in a bit awkward manner of writing them to separate files and then reading them out. Same happens with the download URLs.
The hash verification logic was already implemented and the specific check is here
https://github.com/namark/vircadia/blob/056dfb936328c486308c01e474fec310969f475e/hifi_utils.py#L125-L131
but no hash values were being passed. My guess is that this was intentional due to the bug that my second commit addresses, since before that all the hash checks were failing, producing a different hash for the same file on every retry (see me talking to myself about it in the resolved review above).

daleglass (approved):

PR #1597, opened 2022-02-12
Update API documentation for Window.prompt().

SilverfishVR:
[Documentation](https://apidocs.vircadia.dev/Window.html#.prompt) for ```Window.prompt()``` states that if the user click "Cancel" it will return ```""```
It does, in fact return ```null```, I don't know if that is intentional but arguments could be made for both.

This just updates the documentation and the example .


digisomni (approved):
Just tested it and that indeed seems to be the case, thanks!

PR #1605, opened 2022-02-21
Printing hifi-scribe logs to diagnose a particular GHA issue.

namark:


namark:
Well that was stupid, the logs are available on any build as artifacts...

PR #1611, opened 2022-02-23
Bootstrapping a specific version of VCPKG,

namark:
due to problems with building nvidia texture tools on windows.


vircadia-build-notifier:
 - https://athena-public.s3.eu-west-3.amazonaws.com/builds/PR/vircadia_vircadia/PR1611/Vircadia-Interface-PR1611-e074c27-_1178f0dc.dmg

PR #1595, opened 2022-02-10
Disabled "Visit in VR" button in Domain server web panel.

namark:

Issue #1591, opened 2022-02-03
Vircadia Domain Server "Visit in VR" button uses hifi.place with an internal server error.


IceTheCoyote:
A PR may be needed to fix this issue. But when using the domain-server's web panel you see a button that should let you visit your world in VR but it uses hifi.place which might of not been carried over. I actually showed this in the today's developers meeting.

digisomni:
A possible alternative might be to have this button go to the web interface when we have v1.0 of that live. :)

vijay-prema:
This is mentioned in https://docs.vircadia.com/host/invite-users.html?highlight=link#invite-your-friends
But yeah it doesnt work.  I was hoping to have a some kind of "meeting link" that new people can use to join me

digisomni:
@namark can we just hide the button for now?

namark:
@digisomni I didn't test, but pretty sure this should do it https://github.com/vircadia/vircadia/pull/1595

PR #1627, opened 2022-03-28
Removed 'Visit in VR' button from server admin GUI

Colorfingers:
This link is an artifact of the old Hifi and is broke... removing from admin GUI to reduce confusion.

### DIFF
--- a/domain-server/resources/web/header.html
+++ b/domain-server/resources/web/header.html
@@ -53,10 +53,6 @@
             </li>
           </ul>
           <ul class="nav navbar-right navbar-nav">
-            <a id="visit-domain-link" class="btn btn-default navbar-btn" role="button" target="_blank" style="display: none;">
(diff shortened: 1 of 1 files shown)


namark:
hmmm, It should be permanently hidden in the latest release with this change https://github.com/vircadia/vircadia/pull/1595

Are you sure it's still there?

Colorfingers:
> hmmm, It should be permanently hidden in the latest release with this change #1595
> 
> Are you sure it's still there?

Why permanently hide it... what function does it provide that needs to be retained?  just curious. A user asked why it was there and Madders suggested we remove it.  I was just being proactive.  I should have done a search... we can delete this pull request if it needs to be retained.

namark:
We want to eventually re-enable it, linking to web interface once that's ready, that's why I kept it. Here is the initial discussion https://github.com/vircadia/vircadia/issues/1591

Colorfingers:
Duplicate effort... Button is being hidden later for reuse.

PR #1678, opened 2022-07-13
Enable vcpkg bootstrapping by default.

namark:
Bullet3 and Opus libraries were failing to build, so I had to update them from the latest vcpkg ports and that's what most of the additions are in this PR. Otherwise I added one small patch to bullet3 to enable profiling, and implemented workarounds for arm64 Linux as described in https://github.com/microsoft/vcpkg/issues/3647


julien-me (approved):
Builds without issues.

Looks good to me.

Issue #1421, opened 2021-10-23
Add "Domain supports" to Metaverse heartbeat & Metadata Exporter.


digisomni:
This is so users / apps / bots can determine which Domains will allow them to use things like audio codecs, connect using web, etc.

PR #1716, opened 2022-11-08
Added capabilities object to domain server -> metaverse heartbeat requests.

namark:
Including information on WebRTC configuration, exporters and audio codec preference.

### DIFF
--- a/domain-server/src/DomainServer.cpp
+++ b/domain-server/src/DomainServer.cpp
@@ -1703,12 +1703,13 @@ void DomainServer::sendHeartbeatToMetaverse(const QString& networkAddress, const
     capabilities["webrtc"] = webRTCCapabilities;
 #endif
 
-    QJsonObject prometheusCapabilities;
+    // provides server statistics in a specific format (see DomainServerExporter.h)
+    QJsonObject prometheusExporterCapabilities;
     bool isExporterEnabled = _settingsManager.valueOrDefaultValueForKeyPath(ENABLE_PROMETHEUS_EXPORTER_SETTINGS_KEY).toBool();
     int exporterPort = _settingsManager.valueOrDefaultValueForKeyPath(PROMETHEUS_EXPORTER_PORT_SETTINGS_KEY).toInt();
-    prometheusCapabilities["enabled"] = isExporterEnabled;
-    prometheusCapabilities["port"] = exporterPort;
-    capabilities["prometheus_exporter"] = prometheusCapabilities;
+    prometheusExporterCapabilities["enabled"] = isExporterEnabled;
+    prometheusExporterCapabilities["port"] = exporterPort;
+    capabilities["prometheus_exporter"] = prometheusExporterCapabilities;
 
     QJsonObject metadataCapabilities;
     bool isMetadataExporterEnabled = _settingsManager.valueOrDefaultValueForKeyPath(ENABLE_METADATA_EXPORTER_SETTINGS_KEY).toBool();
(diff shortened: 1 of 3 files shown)


domain-server/src/DomainServer.cpp
```diff
@@ -1677,6 +1686,49 @@ void DomainServer::sendHeartbeatToMetaverse(const QString& networkAddress, const
...
 
+    QJsonObject capabilities;
+
+#if defined(WEBRTC_DATA_CHANNELS)
+    QJsonObject webRTCCapabilities;
+    bool isWebRTCEnabled = _settingsManager.valueForKeyPath(WEBRTC_ENABLED_SETTINGS_KEY).toBool();
+    bool isWebRTCWSSEnabled = _settingsManager.valueForKeyPath(WEBRTC_WSS_ENABLED_SETTINGS_KEY).toBool();
+    auto acmeDomainList = _settingsManager.valueOrDefaultValueForKeyPath("acme.certificate_domains").toList();
+    QJsonArray acmeDomains;
+    for (auto&& acmeDomain : acmeDomainList) {
```
ctrlaltdavid, line 1697:
```suggestion
    for (auto&& acmeDomain: acmeDomainList) {
```

domain-server/src/DomainServer.cpp
```diff
@@ -1677,6 +1686,49 @@ void DomainServer::sendHeartbeatToMetaverse(const QString& networkAddress, const
...
+        acmeDomains.push_back(acmeDomain.toMap()["domain"].toString());
+    }
+    webRTCCapabilities["enabled"] = isWebRTCEnabled;
+    webRTCCapabilities["ssl_enabled"] = isWebRTCWSSEnabled;
+    webRTCCapabilities["ssl_domains"] = acmeDomains;
+    capabilities["webrtc"] = webRTCCapabilities;
+#endif
+
+    QJsonObject prometheusCapabilities;
```
ctrlaltdavid, line 1706: Need either a more descriptive name than "prometheus" or a comment explaining what Prometheus is.

domain-server/src/DomainServer.cpp
```diff
@@ -1677,6 +1686,49 @@ void DomainServer::sendHeartbeatToMetaverse(const QString& networkAddress, const
...
+    capabilities["metadata_exporter"] = metadataCapabilities;
+
+    QJsonObject audioCapabilites;
+    auto audioCodecPreferenceList = _settingsManager.valueOrDefaultValueForKeyPath("audio_env.codec_preference_order").toString().split(",");
+    QJsonArray audioCodecPreference;
+    for (auto&& audioCodec : audioCodecPreferenceList) {
+        audioCodecPreference.push_back(audioCodec);
+    }
+    audioCapabilites["codec_preference_order"] = audioCodecPreference;
+    // TODO: add a list of available codecs, must be retrieved from PluginManager
```
ctrlaltdavid, line 1727: Does this TODO need to be filled in now? Or perhaps there's an issue for this?

ctrlaltdavid (changes requested):

domain-server/src/DomainServer.cpp
namark: I'd say no, cause I'm not yet sure how to do it properly, and I didn't want to spend too much time on this. The audio mixer uses the PluginManager to load the codec plugins and negotiates the audio format based on that and the preference list from the settings. In the simplest case the domain server can also load an enumerate the same plugins, but it might be better to have the audio mixer set a new setting or otherwise communicate it to the domain server, for use cases where they might be separate (different servers/VMs/docker images).

It's all in context of this issue I guess https://github.com/vircadia/vircadia/issues/1421, but the part that's important for the upcoming release with the web client is the WebRTC capabilities.

ctrlaltdavid (approved):

PR #1726, opened 2022-11-28
Added metaverse server URL setting in domain server admin panel.

namark:
Not perfect in terms of communicating the setting to assignment clients https://github.com/vircadia/vircadia/issues/1727, but should be good at least for use cases that don't rely on that. Also the ICE server needs to be separately configured with the same Metaverse URL to authenticate domain servers https://github.com/vircadia/vircadia/issues/1729.


namark:
Tested account registration and login with https://ua92-metaverse.vircadia.com metaverse URL set for a locally running domain server.

ctrlaltdavid (approved):

